PR #67, opened 2014-02-21
Timestamp should always be saved in UTC.

phanle:
When saving timestamps to the database, the timezone information is truncated.  When the timestamp is loaded again later, the timezone is changed, resulting in incorrect data.

Rails solves this by always saving timestamps as UTC.  This pull request only fixes the built-in CreatedAt and UpdatedAt behavior, not any timestamp the user saves.  If you like this approach, we could try to fix it globally.

### DIFF
--- a/callback_create.go
+++ b/callback_create.go
@@ -13,8 +13,8 @@ func BeforeCreate(scope *Scope) {
 
 func UpdateTimeStampWhenCreate(scope *Scope) {
 	if !scope.HasError() {
-		scope.SetColumn("CreatedAt", time.Now())
-		scope.SetColumn("UpdatedAt", time.Now())
+		scope.SetColumn("CreatedAt", time.Now().UTC())
+		scope.SetColumn("UpdatedAt", time.Now().UTC())
 	}
 }
 
--- a/callback_delete.go
+++ b/callback_delete.go
@@ -15,7 +15,7 @@ func Delete(scope *Scope) {
 			scope.Raw(
 				fmt.Sprintf("UPDATE %v SET deleted_at=%v %v",
 					scope.TableName(),
-					scope.AddToVars(time.Now()),
+					scope.AddToVars(time.Now().UTC()),
 					scope.CombinedConditionSql(),
 				))
 		} else {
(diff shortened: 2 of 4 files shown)


davisford:
@jinzhu i know this is closed, but testing with gorm's built-in model including the `CreatedAt`, `UpdatedAt` fields is a pain.  At least for postgres, the timestamps are stored in UTC, which is good.  But to be able to compare the following sequence: 

1. Create a new struct model
2. Persist it
3. Find it
4. Compare 3 with 1

One, *must set `TZ="UTC"` before running the tests -- otherwise it will compare local time vs. UTC.  That's livable.

But more annoying is the database driver's truncation of the timestamp.  Consider this:

```go
func TestCreateReadOrg(t *testing.T) {
    first, err := orgResource.CreateOrg(&Org{Name: "Foo"})
    if err != nil {
      t.Errorf("Unexpected error: %s", err)
    }
    found, err := orgResource.Find(first.ID)
    if err != nil {
      t.Errorf("Unexpected error: %s", err)
    }
    fmt.Println("first  ", first.CreatedAt, first.UpdatedAt)
    fmt.Println("found ", found.CreatedAt, found.UpdatedAt)
    if reflect.DeepEqual(first, found) != true {
      t.Errorf("Not the same")  // always fails
    }
}
```

Prints something like:

```
first 2015-07-27 18:21:23.429832038 +0000 UTC 2015-07-27 18:21:23.429832038 +0000 UTC
found 2015-07-27 18:21:23.429832    +0000 UTC 2015-07-27 18:21:23.429832    +0000 UTC
```

Does anyone have a good workaround for this -- or do I have to write my own validation functions that test each field independently and either skip `time.Time` fields or is smarter about timezone and truncation differences?

codepushr:
@davisford I know this is an outdated discussion but did you find any solution?

davisford:
@codepushr i abandoned gorm altogether on my project.  for the truncation problem with my db driver (postgresql), i ended up writing my own assertion function that compares with some flexibility on the time difference.

codepushr:
@davisford I see... Thanks!

PR #77, opened 2014-03-16
Add DB.Tx() method to provide access to the underlying sql.Tx instance.

tstranex:
Hi

We have a lot of existing code that uses sql.DB and sql.Tx arguments. Sometimes it's necessary to for our new GORM code to interface with the old code. Calling DB.DB() is useful for this, but it doesn't work inside transactions. Inside transactions, it's more useful to access a sql.Tx instance instead. I've added a new DB.Tx() method to return the underlying sql.Tx instance. I think this will be useful for anyone who is migrating to GORM from existing database/sql code.

Regards,
Timothy

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,20 +28,15 @@ func Open(driver, source string) (DB, error) {
 	return db, err
 }
 
-// Return the underlying sql.DB instance.
-//
-// If called inside a transaction, it will panic.
-// Use Tx() instead in this case.
 func (s *DB) DB() *sql.DB {
 	return s.db.(*sql.DB)
 }
 
-// Return the underlying sql.Tx instance.
-//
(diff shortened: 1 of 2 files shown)


sdklab007:
db.CommonDB() no longer works in v1.21.

Meaning, the method is removed entirely. 

Any alternative way to get the underlying sql.Tx instance?

My use-case requires me to find out if the current db object is already a transaction.

ashwinsadeep:
@sdklab007 The following snippet has worked for me:
```golang
func isActiveTransaction(tx *gorm.DB) bool {
	_, ok := tx.Statement.ConnPool.(gorm.TxCommitter)
	return ok
}
```

PR #292, opened 2014-11-22
Create empty slice when records are not found.

jnfeinstein:
There is a gotcha in golang (go figure!) resulting from the
difference between a nil slice and an empty slice. While the length
of both are `0`, the JSON encoder will interpret the nil slice as
`null` and the empty slice as `[]`. Other libraries (i.e. ActiveRecord)
will return `[]` when no values are found. I have found
that this translates into better client-side code, since you don't
have to protect against `null` arrays.

See [this](http://danott.co/posts/json-marshalling-empty-slices-to-empty-arrays-in-go.html) for more information.

### DIFF
--- a/callback_query.go
+++ b/callback_query.go
@@ -77,6 +77,15 @@ func Query(scope *Scope) {
 		}
 
 		if !anyRecordFound {
+			if isSlice && dest.IsNil() {
+				var emptySliceType reflect.Type
+				if isPtr {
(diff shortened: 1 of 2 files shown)


Pokerkoffer:
So why was this closed? Returning nil when no association was found is not correct, it should be empty array. I think this should be handled by gorm.

PR #295, opened 2014-11-22
Allow .Open to reuse existing database connections.

jnfeinstein:
This commit allows you to pass a string or an existing database
connection as the source for gorm. The dialect is still required
because a) there is no common reference to it as far as i know, and
b) gorm allows the dialect to differ from the driver. So, for the sake
of simplicity, you still have to specity the dialect.

This is useful if you have an existing transaction, but still
want to use gorm to format your queries.

This is dependent on the defintion of DB in pkg database/sql having
the field 'dsn', which is the database source, obtained via reflect.

This is a better implementation than #277.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ db, err := gorm.Open("postgres", "user=gorm dbname=gorm sslmode=disable")
 // db, err := gorm.Open("mysql", "user:password@/dbname?charset=utf8&parseTime=True")
 // db, err := gorm.Open("sqlite3", "/tmp/gorm.db")
 
+// You can also use an existing database connection handle
+// dbSql, _ := sql.Open("postgres", "user=gorm dbname=gorm sslmode=disable")
+// db := gorm.Open("postgres", dbSql)
+
 // Get database connection handle [*sql.DB](http://golang.org/pkg/database/sql/#DB)
 db.DB()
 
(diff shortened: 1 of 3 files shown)


j4zz:
@jnfeinstein I stumbled upon this when trying to write gorm migrations with [goose](https://bitbucket.org/liamstask/goose/) after my discovery of #277.

I’m completely new to go, so I’m stuck with the fact that default goose migrations hand you an `*sql.Tx`:

```go
func Up_20130106222315(txn *sql.Tx) { }
```

If I understand correctly, a `*DB` could likely be opened with gorm since this pull request. Can I somehow grab the `*sql.Tx` to use it with gorm?

bigokro:
@j4zz Did you manage to solve this issue? I'm trying to do the exact same thing now. If you've written a wrapper, can you send it via pull request?

pjebs:
You can't get the sql.DB object from the sql.Tx object.
You won't be able to use it with gorm at the moment.

you need to modify gorm to inject the sql.Tx object directly to gorm instead of using gorm.Begin()

pjebs:
Look at the code approximately around where the Begin() function is defined in gorm. That is how you start the transaction.

Try and modify that or make something that was private public so you can access it.

j4zz:
@bigokro No, sorry, I’ve come to terms with writing sql statements for now.

bigokro:
Ok, thanks for the tip @pjebs . @j4zz, looks like I'll have to get used to that, too :)

PR #327, opened 2015-01-04
Fix error when chaining empty where conditions

### LINKED ISSUE #326: Empty struct in Where-chain causes syntax error
I get an SQL syntax error when I chain multiple `.Where()` conditions and one of them is an empty struct.

`db.Where(&Post{}).Where("author = ?", "me").First(&p)`
causes invalid SQL
`SELECT  * FROM "posts"  WHERE  AND author = 'me' ORDER BY "posts".id ASC LIMIT 1`

[Full test gist](https://gist.github.com/xconstruct/77fccbbf8b79dc0c2608)

Background: I'm using the structs as external filters and sometimes they can be empty ('no filter criteria specified'). I expect that an empty struct is optimized away.

Thanks!

xconstruct:
Fixes #326.

### DIFF
--- a/query_test.go
+++ b/query_test.go
@@ -229,6 +229,25 @@ func TestSearchWithMap(t *testing.T) {
 	}
 }
 
+func TestSearchWithEmptyChain(t *testing.T) {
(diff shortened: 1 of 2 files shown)


jinzhu:
Looks good, thank you!

PR #336, opened 2015-01-11
Fixes #335 - Added scoped table name to "deleted_at" where clause. This ...

### LINKED ISSUE #335: "Column 'deleted_at' in where clause is ambiguous" when using Join
The following causes an error:

```go
type Buffer struct {
	Bid  int64      `gorm:"primary_key:yes"`
	Cid  int64      `sql:"not null"`
	Name string     `sql:"size:255;not null"`

	DeletedAt time.Time `sql:"not null"`
}

type Connection struct {
	Cid       int64            `gorm:"primary_key:yes"`
	Name     string `sql:"size:255;not null"`
	DeletedAt time.Time `sql:"not null"`
}

db := repo.DB.Debug().Table("buffer").Select("buffer.*").Joins("JOIN connection ON buffer.cid = connection.cid").Find(&buffers)
```

The generated SQL looks like:
```SQL
SELECT  buffer.* FROM buffer JOIN connection ON buffer.cid = connection.cid WHERE (deleted_at IS NULL OR deleted_at <= '0001-01-02')
```

I believe the code that generates the "deleted_at" should prepend the table name.

bramp:
...ensures it works when the deleted_at column is ambiguous, for example, during a join.

### DIFF
--- a/scope_private.go
+++ b/scope_private.go
@@ -168,7 +168,8 @@ func (scope *Scope) whereSql() (sql string) {
 	var primaryCondiations, andConditions, orConditions []string
 
 	if !scope.Search.Unscope && scope.HasColumn("DeletedAt") {
-		primaryCondiations = append(primaryCondiations, "(deleted_at IS NULL OR deleted_at <= '0001-01-02')")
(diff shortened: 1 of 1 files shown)


jinzhu:
Thank you.

PR #338, opened 2015-01-13
check isIgnored when compiling attrs to update

tdonia:
Hey @jinzhu (and community!) - thanks for the excellent work.  I found an issue where scope.Updates() weren't ignoring fields that had been tagged `sql:"-"` and tracked it down to this.  Note that while I was digging, I also found this introduction from a while ago: https://github.com/jinzhu/gorm/commit/c3d2746912d9e16af54b75e2baefca87ae7990cc#diff-ee3fac9069b0a4978ec2ce9e09640bf7R103 that appears to be no-op?  Was there another way that ignored fields where supposed to be ignored that was never completely implemented / is that a better solution?

thanks!

### DIFF
--- a/utils_private.go
+++ b/utils_private.go
@@ -65,7 +65,7 @@ func convertInterfaceToMap(values interface{}) map[string]interface{} {
 		default:
 			scope := Scope{Value: values}
 			for _, field := range scope.Fields() {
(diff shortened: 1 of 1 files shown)


jinzhu:
Looks good, thank you!

PR #347, opened 2015-01-22
local import to make testing easier on forks

rubysolo:


rubysolo:
For other Go n00bs: https://blog.splice.com/contributing-open-source-git-repositories-go/

PR #353, opened 2015-01-30
check for IsScanner on slices

deoxxa:
I've got a list type that I want to serialise to a single value (included below). With this change, gorm will check to see if a slice can be treated as a value, just as happens with structs below. The column in the database is like `create table x (identifiers jsonb[])`.

```go
type Identifier struct {
	Use      Code          `json:"use"`
	Label    string        `json:"label"`
	System   URI           `json:"system"`
	Value    string        `json:"value"`
	Period   *Period       `json:"period"`
	Assigner *Organization `json:"assigner"`
}

type IdentifierList []Identifier

func (i IdentifierList) Value() (driver.Value, error) {
	bits := []string{}

	for _, v := range i {
		if d, err := json.Marshal(v); err != nil {
			return nil, err
		} else {
			bits = append(bits, strconv.Quote(string(d)))
		}
	}

	v := []byte("{" + strings.Join(bits, ",") + "}")

	return v, nil
}

func (i *IdentifierList) Scan(input interface{}) error {
	b, ok := input.([]byte)
	if !ok {
		return ErrInvalidInput
	}

	var bits []string
	if err := json.Unmarshal([]byte("["+string(b[1:len(b)-1])+"]"), &bits); err != nil {
		return err
	}

	for _, s := range bits {
		var v Identifier
		if err := json.Unmarshal([]byte(s), &v); err != nil {
			return err
		}

		*i = append(*i, v)
	}

	return nil
}
```

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -351,7 +351,9 @@ func (scope *Scope) fieldFromStruct(fieldStruct reflect.StructField, withRelatio
 		case reflect.Slice:
 			typ = typ.Elem()
 
-			if (typ.Kind() == reflect.Struct) && withRelation {
+			if field.IsScanner() {
+				field.IsNormal = true
+			} else if (typ.Kind() == reflect.Struct) && withRelation {
 				if foreignKey == "" {
 					foreignKey = scopeTyp.Name() + "Id"
 				}
--- a/slice_test.go
+++ b/slice_test.go
@@ -1,0 +1,70 @@
+package gorm_test
+
+import (
+	"database/sql/driver"
+	"encoding/json"
+	"testing"
+)
+
+func TestScannableSlices(t *testing.T) {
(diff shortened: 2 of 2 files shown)


jaytaylor:
@deoxxa How about adding a unit-test to ensure correct behavior? ;)

deoxxa:
I can definitely do that - I just wanted to push this when I wrote it so I wouldn't forget to push it later.

jinzhu:
Hi @deoxxa 

Thank you for that, will merge it after get your tests ;)

deoxxa:
There's a test for it now. It doesn't use any advanced features of the database, unlike my example above, but it does test the functionality of gorm.

jaytaylor:
Nice @deoxxa, thanks for coming through on this.  LGTM.

jinzhu:
Hi @deoxxa 

Thank you for your pull request and test

PR #359, opened 2015-02-07
Add Scope.addForeignKey()

sendyhalim:
Add ability for adding foreignkey and update readme

```
// 1st    fieldname
// 2nd    destination table(id)
// 3rd    ONDELETE
// 4th    ONUPDATE
db.Model(&Post{}).AddForeignKey("category_id", "categories(id)", "CASCADE", "CASCADE")
```

PR #360, opened 2015-02-08
Add Scope.addForeignKey()

sendyhalim:
Add ability for adding foreignkey and update readme

```
// 1st    fieldname
// 2nd    destination table(id)
// 3rd    ONDELETE
// 4th    ONUPDATE
db.Model(&Post{}).AddForeignKey("category_id", "categories(id)", "CASCADE", "CASCADE")
```

This is a re-submitted pull request #359 , but with different branch (#359 was on master branch)

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,13 @@ db.AutoMigrate(&User{}, &Product{}, &Order{})
 // Add index
 db.Model(&User{}).AddIndex("idx_user_name", "name")
 
+// Add foreign key
+// 1st param : foreignkey field
+// 2nd param : destination table(id)
+// 3rd param : ONDELETE
+// 4th param : ONUPDATE
+db.Model(&User{}).AddForeignKey("user_id", "destination_table(id)", "CASCADE", "RESTRICT")
+
 // Multiple column index
 db.Model(&User{}).AddIndex("idx_user_name_age", "name", "age")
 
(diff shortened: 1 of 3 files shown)


jinzhu:
Hello @sendyhalim

But seems like this won't works with sqlite3? as far as I know, it doesn't support this syntax ;(

sendyhalim:
Hi @jinzhu ! Thanks for your response. Yeah I've been reading sqlite3 syntax, and it seems that sqlite3 cannot add foreign key by altering the table.

http://stackoverflow.com/questions/1884818/how-do-i-add-a-foreign-key-to-an-existing-sqlite-3-6-21-table

any idea? or would it be good to just print error and return when using sqlite3 driver?

jinzhu:
Yes, it doesn't support it, seems it only support this syntax.

```sql
CREATE TABLE child ( 
    id           INTEGER PRIMARY KEY, 
    parent_id    INTEGER, 
    description  TEXT,
    FOREIGN KEY (parent_id) REFERENCES parent(id)
);
```

I am thinking maybe better we could auto create references when AutoMigrate?

sendyhalim:
@jinzhu I agree, but on some cases people don't want to add foreignkey constraint(just use normal index to prevent constraint). If we create references automatically in AutoMigrate, I think we need to provide some kind of option/flag to prevent automatic foreignkey creation.

jinzhu:
Let just merge this pull request for now, I will rethink about this later.

PR #419, opened 2015-03-13
Postgres driver updated for FoundationDB auto-migration compatibility.

jaytaylor:
$ ./test_all.sh
    testing postgres...
    [info] replacing callback `create` from callback_test.go:94
    [info] removing callback `create` from callback_test.go:107
    `testdb` is not officially supported, running under compatibility mode.
    PASS
    ok  	gorm	1.364s

### DIFF
--- a/postgres.go
+++ b/postgres.go
@@ -76,13 +76,13 @@ func (s *postgres) Quote(key string) string {
 
 func (s *postgres) HasTable(scope *Scope, tableName string) bool {
 	var count int
-	scope.NewDB().Raw("SELECT count(*) FROM INFORMATION_SCHEMA.tables WHERE table_name = ? AND table_type = 'BASE TABLE'", tableName).Row().Scan(&count)
+	scope.NewDB().Raw("SELECT count(*) FROM INFORMATION_SCHEMA.tables WHERE table_schema = current_schema AND table_name = ? AND (table_type = 'BASE TABLE' OR table_type = 'TABLE')", tableName).Row().Scan(&count)
 	return count > 0
 }
 
 func (s *postgres) HasColumn(scope *Scope, tableName string, columnName string) bool {
 	var count int
-	scope.NewDB().Raw("SELECT count(*) FROM INFORMATION_SCHEMA.columns WHERE table_name = ? AND column_name = ?", tableName, columnName).Row().Scan(&count)
+	scope.NewDB().Raw("SELECT count(*) FROM INFORMATION_SCHEMA.columns WHERE table_schema = current_schema AND table_name = ? AND column_name = ?", tableName, columnName).Row().Scan(&count)
 	return count > 0
 }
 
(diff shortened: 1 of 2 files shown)


jinzhu:
Hello @jaytaylor 

Do you think hack postgres dialect to support foundation db is good? or have a foundation dialect maybe better?

jaytaylor:
Hey @jinzhu,

I think this is a good option for us to consider and discuss.

Pros of bundling them together:

* Foundation is already highly compatible with all things postgres, in fact the only driver issue I've run into is this one - where pg_* tables are being referenced and queried.
* Less duplicated code to maintain.

Cons:

* Supporting both in one driver complicates the PG driver.

My proposal is to merge this compatibility PR for now, and re-evaluate if/when additional compatibility issues arise.  In my opinion the optimal path is to start with a simple approach that works and then spend the effort to refactor FDB into it's own driver when we are confident it's warranted.

Believe me, I know Gorm is great and I really like this project!  I don't want to see any of it get ugly.  At the same time I want to ensure we keep it maximally useful and avoid prematurely introducing  complexity in the form of ~95% duplicated drivers.

What are your thoughts?

jinzhu:
Ok, lets do this, but seems like this will broken pg test in my machine

selection_477

```go
+ // Determine whehter the pg_indexes talbe is
+ detect := []int{} //make([]int, 2) // If detect[0] == 1 then we have postgres, else if detect[1] == 1 then we have detectationDB.
+ scope.NewDB().Raw(`SELECT count(*) "detect" FROM INFORMATION_SCHEMA.tables WHERE table_name = 'pg_indexes' UNION SELECT count(*) "detect" FROM INFORMATION_SCHEMA.tables WHERE table_name = 'indexes'`).Pluck("detect", &detect)

Seems like we can't use detect[0] == 1, detect[1] == 1 to detect it is pg or fdb
```

jaytaylor:
Nice catch, I will fix this up.

jaytaylor:
@jinzhu This should now be cleaned up and ready, I've tested thoroughly on both postgres and  foundation.

=== RUN TestAutoMigration
--- PASS: TestAutoMigration (0.01s)

jaytaylor:
Actually hang on, I am seeing issues with this.

jaytaylor:
I've added an additional improvement in the Pluck error messaging.

jinzhu:
Hello @jaytaylor 

Did you noticed my latest change about the dialect? the postgres dialect embedded a common dialect, so I am thinking, maybe we could add a foundationdb dialect, and embed the postgres dialect, then just overwrite those methods behavioured differently?

jaytaylor:
Sure, that sounds reasonable to me!

jaytaylor:
@jinzhu I'm closing this PR in favor of a new one for the full FoundationDB driver.

Cheers,
Jay

PR #426, opened 2015-03-17
Fix creation of composite unique indexes

willglynn:
Previously, specifying a struct like:

    type Foo struct {
    	Email        string    `sql:"unique_index:idx_email_agent"`
    	UserAgent    string    `sql:"unique_index:idx_email_agent"`
    }

…would result in an `idx_email_agent` index created over `user_agent`, excluding `email`.

### DIFF
--- a/scope_private.go
+++ b/scope_private.go
@@ -570,7 +570,7 @@ func (scope *Scope) autoIndex() *Scope {
 			if name == "UNIQUE_INDEX" {
 				name = fmt.Sprintf("uix_%v_%v", scope.TableName(), field.DBName)
 			}
-			uniqueIndexes[name] = append(indexes[name], field.DBName)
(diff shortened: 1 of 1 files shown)


jinzhu:
Ah... sorry for the problem. 

Thank you.

ohohco:
```
type Account struct {
	ID       string `gorm:"primary_key;uuid"`
	Username string `sql:"-"`
	Email    string `sql:"unique_index:idx_email_mobile"`
	Mobile   string `sql:"unique_index:idx_email_mobile"`
	Password string `sql:"not null; type:varchar(50);"`
	Salt     string
	Status   int
	Code     string
}
```
still not work with `unique_index:idx_email_mobile` or `index:idx_email_mobile` in mariadb `5.5.5-10.0.20-MariaDB`

PR #438, opened 2015-03-28
Skip saving associations during UpdateColumns(...)

jaytaylor:
Yesterday I noticed that when invoking `UpdateColumns(...)` it would try to save all associated records.  I found this to be highly undesirable and it seemed like only an `UPDATE ...` statement should have been run.

This patch achieves the desired effect for at `UpdateColumns(...)`.

Relevant unit-test is included.

    === RUN TestUpdateColumnsSkipsAssociations
    --- PASS: TestUpdateColumnsSkipsAssociations (0.10s)
    PASS

All unit tests:

    ok  	github.com/jinzhu/gorm	2.576s

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,7 +360,7 @@ db.Model(&user).Updates(User{Name: "hello", Age: 18})
 
 ### Update Without Callbacks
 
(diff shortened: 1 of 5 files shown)


jinzhu:
looks good, thank you.

PR #439, opened 2015-03-29
Fix type of OwnerType in polymorphism example

gwTumm:
`OwnerType` is either `"Cat"` or `"Dog"` in this example, so the actual type should be `string`


jinzhu:
thank you!

PR #442, opened 2015-04-04
Add an example with time

etcinit:
It took me a bit to figure out how to query time since there isn't any documentation on it. I thought it would be nice to have an example on the README for anybody looking to query fields like `created_at` or `updated_at`, etc.


jinzhu:
Thank you

jamiefoster:
How is lastweek defined in the example?

PR #464, opened 2015-04-20
mutex.lock  var smap = map[string]string{} on update

leonardaustin:


jinzhu:
Hi @leonardaustin 

Thank you for your pull request, but I think this should be not really necessary, I have implemented it before, but removed it later.

Even we need it, the sync should not be there. anyway, thank you for your work ;)

PR #476, opened 2015-04-26
Quote the primary column name when doing queries

Femaref:
Postgresql requires certain column names to be quoted.
When unquoted, all upper-case characters will be converted to lower-case,
and column names like 'typeID' will result in a query on 'typeid', which will fail the query.


jaytaylor:
LGTM.

jinzhu:

Sorry for the delay, I am still on vocation

PR #479, opened 2015-05-02
Update README.md

nguyendangminh:
The line:
// db := gorm.Open("postgres", dbSql)
should be
// db, _ := gorm.Open("postgres", dbSql)


jinzhu:
looks good, thank you.

PR #488, opened 2015-05-13
Add examples for join.

tuvistavie:
I had to look in the issues to find how to use `db.Joins` to fetch records directly, so I thought a few more examples would not hurt.


jinzhu:
thank you

PR #491, opened 2015-05-14
Implementing crate dialect

aanm:
I believe this is only a small amount of work for crate to be implemented on this ORM. But for now those were the only workarounds that I've found so crate would work.

Feel free to discuss anything.
The crate that I'm talking about is https://crate.io/
Signed-off-by: André Martins <martins@noironetworks.com>

### DIFF
--- a/callback_create.go
+++ b/callback_create.go
@@ -80,8 +80,17 @@ func Create(scope *Scope) {
 				if results, err := scope.SqlDB().Exec(scope.Sql, scope.SqlVars...); err != nil {
 					scope.db.RowsAffected, _ = results.RowsAffected()
 				}
-			} else if scope.Err(scope.SqlDB().QueryRow(scope.Sql, scope.SqlVars...).Scan(primaryField.Field.Addr().Interface())) == nil {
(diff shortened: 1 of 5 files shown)


jaytaylor:
Hey @aanm,

Thanks for this, I hadn't seen crate before.  It looks interesting!

And.. how about some unit-test integration for crate (like is available for all the other supported backends)?

aanm:
@jaytaylor Are you talking about this https://github.com/jinzhu/gorm/blob/master/main_test.go

jaytaylor:
@aanm yep, that kind of stuff is what I'm referring to.

aanm:
@jaytaylor Sorry, it turns out we are not going to pursuing this since it is not exactly what we need. You can close this PR or leave it open if anyone want to continue want it is missing.

jaytaylor:
@aanm thanks for following up!  I am curious about what you guys are pursuing instead, can you share?

jinzhu:
I am going to close this one, seems like I am not a big fan of this database.

I  will open an API in the future, so you could implement dialects for other databases w/o modify gorm's source code.

PR #499, opened 2015-05-20
`Alive' column soft-deletion support.

jaytaylor:
**This makes it possible to still enforce unique constraints (e.g. "name", "alive"), but not have the unique constraint enforced for "deleted" records.**

----

**Relevant excerpt from the updated README:**

### Soft-Delete

If a struct has a `DeletedAt` field, it will get soft-delete ability automatically!
Then it won't be deleted from database permanently when `Delete()` is called.

Or alternatively..

If a struct has an `Alive` field it will get soft-delete ability automatically!
Then it won't be deleted from the database permanently when `Delete()` is called.
`Alive` will be set to `null`.  This makes it possible to still enforce unique
constraints (e.g. "name", "alive"), but not have the unique constraint enforced
for "deleted" records.

```go
type Friend struct {
  ID    int64
  Name  string `sql:"type:varchar(255)`
  Alive int    `json:"-" sql:"type:integer;DEFAULT: 1;"`
}

db.Delete(&friend)
//// UPDATE friends SET alive=null WHERE id = 42;
```

Note: Choose only one or the other.  It won't work right if both `DeletedAt` and
`Alive` columns are defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ type User struct{} // struct User's database table name is "users" by default, w
 * Use `CreatedAt` to store record's created time if field exists
 * Use `UpdatedAt` to store record's updated time if field exists
 * Use `DeletedAt` to store record's deleted time if field exists [Soft Delete](#soft-delete)
+* Use nullable `Alive` to store record's aliveness if field exists [Soft Delete](#soft-delete)
 * Gorm provide a default model struct, you could embed it in your struct
 
 ```go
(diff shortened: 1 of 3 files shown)


jaytaylor:
@jinzhu I've been using this change in production for two weeks now and it's been working great.  Can you share your thoughts on this proposed PR?

jinzhu:
Hello @jaytaylor 

Why not just create unique index with name and deleted_at? I think that will also match your need?

jaytaylor:
Hey @jinzhu,

It's true that would work; however it requires complicating the definition of what constitutes a unique record, which [IMO] makes soft-deletion trickier to reason about.

The "alive" column alternative provides the soft-deletion support without requiring any additional unique constraint indexes or complexity in that department.

I think it's useful to include support for both approaches and let the library users make their own choice about which one best suits their use case.

jinzhu:
Hi @jaytaylor 

I think that's pretty easy to do that with unique_index tag, you could do it like this:

```go
type User struct {
  ID uint `sql:"unique_index:unique_user"`
  DeletedAt *time.Time  `sql:"unique_index:unique_user"`
```

If this pull request got merged, we will have similar solutions to resolve similar issues, and this will slow down all query, create operations, I think this is we all don't want to see this.

But if you want like to have this feature in your application, we could redefine the callbacks to have it. refer https://github.com/jinzhu/gorm/blob/master/doc/development.md

jaytaylor:
Hi @jinzhu,

I'd be glad to implement this via callbacks only, however I am running into an roadblock.

There is some special behavior for the 'alive' field in `scope_private.go`.`whereSql()`.  I don't see any way to override this via callbacks.

jinzhu:
Hi @jaytaylor 

I should be easy to do that, just with this:

```go
func AppendAliveToQuery(scope *gorm.Scope) {
	if !scope.Search.Unscoped && scope.HasColumn("alive") {
		scope.Search.Where("alive IS NOT NULL")
	}
}

db.Callback().Query().Before("gorm:query").Register("append_alive", AppendAliveToQuery)
```

jaytaylor:
Awesome, that works like a charm.

Thank you so much @jinzhu !!!

PR #517, opened 2015-06-04
insert_id only support auto_increment field

CodeRushing:
When create a record, insert_id which supported can only assign to an auto_increment field


jinzhu:
Seems like this broken the test?

PR #518, opened 2015-06-04
Fix Create when dialect does not support last inserted id

bojan-embroker:
This fixes a panic when err it not nil and result is nil, but error is still not reported by Create.


bojan-embroker:
Fix for error reporting added.

jinzhu:
Looks good, thank you.

PR #520, opened 2015-06-05
Support for preload of Struct Ptr

crystalin:
This fixes the issue when preloading .Preload("Project.Repositories").Find(&[]User{})
with 
type User struct {
  Project *Project
}

type Project struct {
  Repositories []Repository
}

type Repository struct {
...
}


jinzhu:
Looks good, thank you.

PR #527, opened 2015-06-09
Update README.md with more realistic Transaction example

yanfali:
Add a more detailed slightly more realistic example for a transaction. This example contains error handling and transaction usage that is not reflected in the current README. Hopefully this will help others learn to how use gorm more quickly.


jinzhu:
Thank you

Issue #530, opened 2015-06-11
Feature: Support for associated foreign key in has_one & belongs_to relations


crystalin:
Support the possibility to have:
```go
type Person struct {
  Id int64
  Email string
}

type Invitation struct {
  Id int64
  Address string

  Person *Person `gorm:"foreignkey:Email;associatedkey:Address`
}
```

I don't think it requires a lot of changes to do so.

jinzhu:
This should been fixed. refer https://github.com/jinzhu/gorm/blob/master/customize_column_test.go#L108-L141

PR #537, opened 2015-06-15
Adds "associationforeignkey" tag in the has_one and belongs_to cases

crystalin:
To handle when the foreign key is not the primary key.
This has not been heavily tested (enough tested to work in all my cases).


crystalin:
This is related to issue #530

jinzhu:
Hi @crystalin 

Could you add some tests about this?

jinzhu:
Going to close this one, as we have rewritten those part of code, and it can't be merged already.

PR #538, opened 2015-06-15
Feature/insert id first support auto incrment field

CodeRushing:
insert_id first support auto_increment field.

### DIFF
--- a/callback_create.go
+++ b/callback_create.go
@@ -70,8 +70,8 @@ func Create(scope *Scope) {
 				id, err := result.LastInsertId()
 				if scope.Err(err) == nil {
 					scope.db.RowsAffected, _ = result.RowsAffected()
-					if primaryField != nil && primaryField.IsBlank {
-						scope.Err(scope.SetColumn(primaryField, id))
+					if autoIncrementField := scope.AutoIncrementField(); autoIncrementField != nil {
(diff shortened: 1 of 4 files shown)


jinzhu:
Hi @CodeRushing 

What's your use case for this?

CodeRushing:
Hi, @jinzhu 
I just add struct tag to User struct in struct_test.go. It's compatible to the old code, but first assign insert_id to the field which has the auto_increment tag

jinzhu:
I am thinking why you don't return primary key but return the first auto increment field? (or the question could be: why don't use auto increment field as primary key, but use it for another field?)

CodeRushing:
When a table has no auto_increment field, like in mysql, it returns 0 for insert_id. When you just use  primary key to fill the insert_id,  then unfortunately the insert_id will replace the real value of the PrimaryField() returned. The insert_id in mysql is for the auto_increment field , not for primary key field. On the other hand, if the insert_id is not for the field which named id, it cause mismatch value

CodeRushing:
@jinzhu 
Do you have any other questions?

jinzhu:
Hi @CodeRushing 

If the primary key has value, then it won't overwrite it with the last insert id.

So I think this case won't happen?

PR #539, opened 2015-06-15
Update README for revertable transactions

rohanthewiz:
My tests are showing that transactions are not being rolled back if done through the original *db* object.
The update to the README clarifies that revertible operations must be done through the returned *tx* object.


jinzhu:
Hi @rohanthewiz 

Yes, those operations need to be done with the transaction db object.

But seems like the update of the README doesn't really make this became clear, could you please try to improve that?

Thank you.

PR #541, opened 2015-06-17
Provide clear explanation of transaction database handle

rohanthewiz:
This should clarify the use of the transaction database handle in general.
The example following does give a very good specific example.


jinzhu:
Thank you!

PR #550, opened 2015-06-29
Added missing field name quoting for DDL statements

jaytaylor:
Added missing field name quoting for `ALTER TABLE' and `CREATE INDEX' statements.


jinzhu:
Looks good, thank you.

Issue #529, opened 2015-06-10
composite primary keys which are both uint adds auto_increment to both fields


yanfali:
In unit test multi_primary_keys_test.go you test for a composite primary key which is uint/string and this seems to work correctly. However if you attempt to use uint/uint it adds auto_increment to both fields which is not legal on mysql/mariadb using innodb.

type Content struct {
	DrawerID   uint `gorm:"primary_key" sql:"not null" json:"drawerId,omitempty"`
	ArtifactID uint `gorm:"primary_key" sql:"not null" json:"artifactId"`
	State      int  `sql:"not null" json:"state"`
}

Produces:

CREATE TABLE `content` (`drawer_id` int unsigned AUTO_INCREMENT NOT NULL ,`artifact_id` int unsigned AUTO_INCREMENT NOT NULL ,`state` int NOT NULL  , PRIMARY KEY (`drawer_id`,`artifact_id`))

Is there a way to tell it not to use auto_increment on a primary key? My current work around is to use raw sql after the tables have been created to add the primary keys.

li3p:
sounds like you should use unique key , not primary key.

yanfali:
Hi @li3p actually that wouldn't work. I need the combination (composite) of both keys to be unique. Having each column be unique would only prevent me using it as a many to many relationship. The issue as it stands is that since I'm also using gorm to generate the tables, these two use cases are interfering with each other.

If I didn't use Gorm to generate my schema and just used the markup, I believe it would work. Not having gorm recognize that these 2 fields are a key means that inserts are being instead of updates on save and this causes the constraints I set up on the table to fail.

yanfali:
So I came up with a horrible work around:

```
@@ -349,6 +353,11 @@ func (scope *Scope) generateSqlTag(field *StructField) string {
                if field.IsPrimaryKey {
                        autoIncrease = true
                }
+               _, noAutoIncrease := sqlSettings["NO_AUTO_INCREMENT"]
+               if noAutoIncrease {
+                       println("not auto")
+                       autoIncrease = false
+               }
```

which patches model_struct.go. What I would prefer to do, is add a field to modelStruct.IsComposite and use this to change the behavior of autoIncrease and not set it to true if this primary key field is part of a composite key. I couldn't quite figure out how to do that from within `generateSqlTag`.

What's a good way to do this from within `createTable`? It seems like the signature for generateSqlTag needs to change or the StructField needs a field. Any advice you have on the cleanest way to do this would be appreciated.

[...]

PR #556, opened 2015-07-02
No auto_inc of gorm:"primary_key" if foreignkey

yanfali:
This is in response to Issue #529 

If you are trying to create a table to store metadata about the
relationship of two entities, e.g. Product and Store and the quantity of
product available at that store. You may wish to use a composite foreign
key relationship.

If you also wish to use gorm to generate the SQL to create the schema,
by default gorm will add the auto_increment keyword to tables which are
both primary and integer. This is a undesirable for composite foreign
key relationships and generates invalid sql statements.

If you have a struct field which is a primary key and is marked as a
foreign key, disable the auto increment behavior while generating the
schema.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1116,6 +1116,21 @@ type Product struct {
 }
 ```
 
+## Composite Foreign Primary Integer Keys
+
+```go
+type ProductStore struct {
+       ProductID    int64 `gorm:"primary_key;foreignkey"`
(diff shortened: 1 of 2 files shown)


yanfali:
Hi @jinzhu not sure how you add tests for generated sql output? Any advice?

Thanks

Yan

yanfali:
Hmm, no working as I expected after a fork. Closing.

Issue #553, opened 2015-07-01
quoting breaks alter table on mariadb


yanfali:
git commit 27a442b5ecef401a3ba2325383a96a09b0a983d9 breaks mariadb alter table statements:

Broken generated SQL:
```
ALTER TABLE `drawer` ADD CONSTRAINT `drawer_analyst_id_foreign` FOREIGN KEY (`analyst_id`) REFERENCES `analyst(id)` ON DELETE RESTRICT ON UPDATE RESTRICT;
```
Working SQL:
```
ALTER TABLE `drawer` ADD CONSTRAINT drawer_analyst_id_foreign FOREIGN KEY (analyst_id) REFERENCES analyst(id) ON DELETE RESTRICT ON UPDATE RESTRICT;
```
Seems like mariadb specifically doesn't like quoting of REFERENCES `analyst(id)` and throws an sql error:
```
2015/07/01 15:29:29 Error 1005: Can't create table `analyst`.`#sql-562_210` (errno: 150 "Foreign key constraint is incorrectly formed")
```

yanfali:
Applying the following patch allows it to work again on mariadb.

```
-       scope.Raw(fmt.Sprintf(query, scope.QuotedTableName(), scope.Quote(keyName), scope.Quote(field), scope.Quote(dest), onDelete, onUpdate)).Exec()
+       scope.Raw(fmt.Sprintf(query, scope.QuotedTableName(), scope.Quote(keyName), scope.Quote(field), dest, onDelete, onUpdate)).Exec()
```

Not sure what the right answer is though for different databases.

jaytaylor:
Right, then it will break all the other databases which require the names
to be quoted!

:s

yanfali:
@jinzhu there may be an alternate format that works for everyone. `analyst` (`id`). This seems to work on mariadb. @jaytaylor does this work on your database?

yanfali:
OK, I think the simplest way to fix this is update Quote in mysql.go to be smarter. I'll try and get something working.

yanfali:
OK, if you patch mysql.go to work something like this:

```
func (mysql) Quote(key string) string {
	if strings.Contains(key, "(") {
		pos1 := strings.Index(key, "(")
		pos2 := strings.Index(key, ")")
		return fmt.Sprintf("`%s`(`%s`)", key[0:pos1], key[pos1+1:pos2])
	}
	return fmt.Sprintf("`%s`", key)
}
```
[...]

PR #557, opened 2015-07-02
Issue #553 quoting breaks mariadb/mysql

yanfali:
The new quoting behavior during table creation is generating
incompatible sql code for mariadb and likely mysql. Make the Quote
function a little bit smarter in mysql.go so it doesn't quote parens.

### DIFF
--- a/mysql.go
+++ b/mysql.go
@@ -3,6 +3,7 @@ package gorm
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


jinzhu:
HI @yanfali 

I have used mariadb for testing, but haven't got this error yet, could you let me know how to reproduce the issue?

yanfali:
I created a foreign key relationship using the api. That triggers for me.
Hope that helps.

yanfali:
Specifically

 o.DB.Model(&MyTable{}).AddForeignKey("other_table_id", "other_table(id)", "RESTRICT", "RESTRICT").

Since it accepts strings and is explicitly specified by user, it breaks due to the new quoting. Does that make sense?

yanfali:
Hi @jinzhu I noticed you and @jaytaylor checked in some changes related to the quoting. I tried using them, but they didn't seem to help. Is there a suggested resolution to this use case? Thanks

jinzhu:
Hello @yanfali 

Sorry for the delay, pushed a commit, that should fix the problem.

Thank you.

yanfali:
Thank you!

PR #558, opened 2015-07-02
Add `*DB' param to JoinTableHandlerInterface.`Setup' interface

jaytaylor:
Add `*DB' param to JoinTableHandlerInterface.`Setup' interface to preserve DB settings in intermediate setup scope.

Fixes previously non-functioning/broken use-case:

- Using the `gorm:many2many' relations tag with `db.SingularTable(true)' should always reference singular table names.

### DIFF
--- a/join_table_handler.go
+++ b/join_table_handler.go
@@ -8,7 +8,7 @@ import (
 )
 
 type JoinTableHandlerInterface interface {
-	Setup(relationship *Relationship, tableName string, source reflect.Type, destination reflect.Type)
+	Setup(db *DB, relationship *Relationship, tableName string, source reflect.Type, destination reflect.Type)
 	Table(db *DB) string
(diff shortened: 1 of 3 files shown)


jinzhu:
Hi @jaytaylor 

Seems like the db is not used even you passed it to setup, what do you want to use it for?

jaytaylor:
It is used during as part of Scope.`GetModelStruct ' in [the table sungularity-ness check/enforcement](https://github.com/jinzhu/gorm/blob/master/model_struct.go#L115).

```(go)
    115. 		if scope.db == nil || !scope.db.parent.singularTable {
```
when this region is hit, without the db state the check always returns false due to nil db value from temporary scope.

jaytaylor:
@jinzhu Let's hold off on merging this.  I spent some time this weekend devising a unit-test for the broken case, and while this patch fixes some broken cases when `SingularTable = true`, it doesn't fix all of them.

When I get time and get a completely working fix together I will submit a new PR.

PR #570, opened 2015-07-17
overwrite initiliazed slices as empty slices

ilgooz:
2d802c3 overwrites non nil slices as not initialized slices and this causes my json marshalled slices to be "null" instead of "[]". We can overwrite non nil slices as empty slices to avoid this unexpected behavior.


jinzhu:
HI @ilgooz,

Thank you for your PR. I have fixed this issue based on what you have written.

Thank you.

PR #572, opened 2015-07-19
Table suffix to create tables with InnoDB engine

eirmag:
Hi all,

I've forked jinzhu/gorm more than a year ago, and did modify the code at that time to provide another mysql engine. For instance, the goal was to create tables with InnoDB engine without changing the default storage engine. It's time to share the code if some of you find some interest.

Code is pretty simple, but I did create a new function to distinguish the regular Open and the version with table options, called "OpenWithTableSuffix". Looking at latest syntax, https://dev.mysql.com/doc/refman/5.7/en/create-table.html, it exists several table options. Engine is one of the many possibilities. 

Best regards,
Gabriel

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,6 @@ import (
 db, err := gorm.Open("postgres", "user=gorm dbname=gorm sslmode=disable")
 // db, err := gorm.Open("foundation", "dbname=gorm") // FoundationDB.
 // db, err := gorm.Open("mysql", "user:password@/dbname?charset=utf8&parseTime=True&loc=Local")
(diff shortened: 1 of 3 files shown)


jinzhu:
HI @eirmag 

Thank you for your PR.

What about we only support this? `db.Set("gorm:table_suffix", "...").Create(&User{})` and don't need to introduce another API

eirmag:
Hi @jinzhu,

I've modified the PR to remove the `OpenWithTableSuffix` function, and updated the README accordingly. I've also renamed the gorm parameter to `gorm:table_options` to better reflect its purpose.

Best regards,
Gabriel

jinzhu:
Thank you.

PR #576, opened 2015-07-22
Add basic support for multiple HAVING clauses.

kiwih:
All HAVING clauses are AND-ed together. Code modelled off whereSql(). Build passes with Sqlite3.

Resolves jinzhu/gorm#575


jinzhu:
Thank you for your PR.

PR #583, opened 2015-07-28
fix panic in function databaseName()

liudanking:
Fix panic in function databaseName when there are special characters in password, such as '?', '/'.


jinzhu:
thank you.

PR #585, opened 2015-07-28
Fixed issue https://github.com/jinzhu/gorm/issues/151 where postgresql connection problems fail silently

pariz:
For reference: https://github.com/jinzhu/gorm/issues/151
Errors connection to a postgresql fail silently, resulting in errors. This fix sends a ping right after the connection has been made. If the ping fails, we assume the connection is broken.

Accepting better solutions. This quick fix did it for me :)

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,10 +60,8 @@ func Open(dialect string, args ...interface{}) (DB, error) {
 			}
 			dbSql, err = sql.Open(driver, source)
 
-			// For some reason, postgres does not throw an connection error.
-			// Sending a ping request after the connection is made is a quick workaround for this
-			if driver == "postgres" && err == nil {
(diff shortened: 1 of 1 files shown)


jinzhu:
Should we always `Ping()` the database after connection?

pariz:
It's not the most pretty solution. But it an extra safetycheck to verify that the connection is in fact alive.

pariz:
Should i rewrite the pr and make the `Ping()` generic?

jinzhu:
Hi @pariz 

Yes, I think it is a better solution than just check the postgres.

pariz:
I agree. Code has been updated. Epic oneliner! :)

jinzhu:
picked your code to the master, thank you.

PR #592, opened 2015-08-02
documented foreignkey

msmarino:
Hi Jinzhu!

A great library!!

This is just a small documentation update...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1111,6 +1111,16 @@ type Animal struct {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @msmarino 

In your example, I think this is a standard foreign key?  gorm could identify it without the tag.

msmarino:
When I wrote the bug, the example failed exactly as stated, and was
corrected with that exact code. The foreign key was not detected without
the tag.

jinzhu:
So it might a bug then. does the problem fixed?

PR #600, opened 2015-08-06
`CurrentDatabase' implementation.

jaytaylor:
I've found this functionality to be very useful in my own projects to avoid hardcoding database names.

I took the time to add support for all dialects and add a corresponding unit-test.

### DIFF
--- a/common_dialect.go
+++ b/common_dialect.go
@@ -3,7 +3,6 @@ package gorm
 import (
 	"fmt"
 	"reflect"
-	"strings"
 	"time"
 )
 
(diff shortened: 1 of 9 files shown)


jaytaylor:
I just noticed that there is a pre-existing databaseName() function that attempts to infer the name from the connection string.  I plan to see clean that up in this PR.

jaytaylor:
I believe this is now ready to be reviewed!

jinzhu:
Hi @jaytaylor

Thank you for your PR as always.

But why not just return the database name as string? It sounds easier for users to use it?

jaytaylor:
I started out with that, but when using it, it felt unnatural to not return *DB like the rest of the gorm API.  So I changed it to be more like gorm-style: `gorm.First(&model)`.

jinzhu:
I think that's fine to return the database name as string. it would be easier for others to use it?

jaytaylor:
I agree that either implementation will work, happy to change it to the way you like.

jaytaylor:
@jinzhu Requested changes implemented and merged into PR #606 as well!

PR #603, opened 2015-08-07
fixes #593 - Dont include quotes on dest table

### LINKED ISSUE #593: AddForeignKey quotes reference table
Hello,

I believe the AddForeignKey table has a small bug in it:

My golang code:

```go
 db.Model(&A{}).AddForeignKey("b_id", "b(id)", "RESTRICT", "RESTRICT")
```
Generated Query:

```sql
 ALTER TABLE "a" ADD CONSTRAINT "a_b_id_foreign" FOREIGN KEY ("b_id") REFERENCES "b(id)" ON DELETE RESTRICT ON UPDATE RESTRICT;
```

Expected query (runs in postgres):

 ```sql
ALTER TABLE "a" ADD CONSTRAINT "a_b_id_foreign" FOREIGN KEY ("b_id") REFERENCES b(id) ON DELETE RESTRICT ON UPDATE RESTRICT;
```

I *think* the solution is to not QuoteIfPossible on: https://github.com/jinzhu/gorm/blob/0def184b0ca6d0b4d4b772a18b909d35b6cb86a1/scope_private.go#L574 for dest

Hope this is helpful for others who use foreign keys :)  I was able to convert all but two of my foreign keys to struct tags (`sql:"type uuid REFERENCES c(id)"`), but I encountered this when two tables referred to one another (so I can't change creation ordering, heh)

leonmaia:


jaytaylor:
@leonmaia Why?  Please provide examples of what is broken and what this fixes.

jaytaylor:
Related: https://github.com/jinzhu/gorm/issues/593

I'd love to see a solution that is able to do proper quoting, rather than reverting back to completely unquoted.

leonmaia:
Sure! I'll work on this tonight.

PR #604, opened 2015-08-07
Preserve modelStructs state

jaytaylor:
Preserve modelStructs so that multiple invocations of `db.SingularTable(true|false)` don't clobber the state.

Includes corresponding unit-test.


jaytaylor:
Doesn't actually fix the problem for me.  I was a little to eager :)

PR #605, opened 2015-08-08
Recalculate fields in `Scope.Fields' when empty (nil) or number of fields is LTE 1.

jaytaylor:
Recalculate fields in `Scope.Fields` when empty (nil) or number of fields is LTE 1.

Protect the `.fields' variable state from partial-initialization timing
issues.

As gorm warms state and makes `.GetStructFields` calls, they lead to a long
deferred function in model_struct.go.  Then the deferred function calls
`Scope.Fields`, leading to another `.GetStructFields` call.. cyclically.
Once information is cached the cycling ends.

This extra rule evicts incorrect/suspiciously short fields information.

Symptom of incorrect internal state ocurring can manifest as invalid
insert SQL statement errors generated during `DB.Save', e.g.:

    INSERT INTO "your_table" DEFAULT VALUES RETURNING "your_table"."id"

This is why we refresh if nil or only a single field (usually is the ID field).

Included is a unit-test which triggers the scenario and prove that it is
fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1210,6 +1210,7 @@ db.Where("email = ?", "x@example.org").Attrs(User{RegisteredIp: "111.111.111.111
 * Github Pages
 * AlertColumn, DropColumn
 * R/W Splitting, Validation
+* AddForeignKey compatibility with all dialects
 
 # Author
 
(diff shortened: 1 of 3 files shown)


jaytaylor:
@jinzhu Do you have any thoughts about this PR?

jinzhu:
Hi @jaytaylor 

This won't fix problem "Protect the `.fields' variable state from partial-initialization timing
issues."

Just pushed a commit to fix it, thank you for your PR

PR #606, opened 2015-08-08
Ensure DDL dialect queries propagate errors back to descendent scopes.

jaytaylor:
Ensure DDL dialect queries propagate error states to descendent scopes.

Includes relevant unit-test.

Branched from jay/current_database (please merge that branch first!).

### DIFF
--- a/main.go
+++ b/main.go
@@ -365,10 +365,7 @@ func (s *DB) Rollback() *DB {
 }
 
 func (s *DB) NewRecord(value interface{}) bool {
-	scope := s.clone().NewScope(value)
-	result := scope.PrimaryKeyZero()
-	s.err(scope.db.Error)
-	return result
+	return s.clone().NewScope(value).PrimaryKeyZero()
 }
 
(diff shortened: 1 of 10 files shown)


jaytaylor:
This is now ready for review.

main.go
```diff
@@ -365,7 +365,10 @@ func (s *DB) Rollback() *DB {
 }
 
 func (s *DB) NewRecord(value interface{}) bool {
-	return s.clone().NewScope(value).PrimaryKeyZero()
+	scope := s.clone().NewScope(value)
+	result := scope.PrimaryKeyZero()
+	s.err(scope.db.Error)
```
jinzhu: This `s.err()`  is not necessary, there will be SQL be invoked here.

jaytaylor: Without it I don't think the unit-test will pass.  I will double check and provide more details when I get home.

jaytaylor:
@jinzhu You were spot on - it works fine w/o those lines so I've reverted `NewRecord` back to the original logic.

Thanks!

PR #645, opened 2015-09-02
Change logger to use an interface.

hobeone:
This allows reuse of the nice formatting done by gorm while still sending the output to a custom logger.

For my use case it allowed easy integration with the glog library with a simple wrapper:

    type LogWrapper struct{}

    func (l *LogWrapper) Println(v ...interface{}) {
        glog.Info(v)
    }

and then something like:

`d.SetLogger(gorm.Logger{&LogWrapper{}})`

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -14,8 +14,12 @@ type logger interface {
 	Print(v ...interface{})
 }
 
+type LogWriter interface {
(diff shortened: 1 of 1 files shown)


jinzhu:
Thank you for your PR.

PR #646, opened 2015-09-06
Add query example code

deltam:
I add query example code because this behavior is unexpectedly to me.


jinzhu:
Hi @deltam 

Thank you for your PR.

Although it is possible to write code like this, but this is not a suggested way ;(

Thank you anyway.

PR #650, opened 2015-09-07
Added support to schema name in returning clause

seriallink:
The ReturningStr func was quoting schema and table name together:
```
INSERT INTO "xyz"."person" ... RETURNING "xyz.person"."idperson"
```
Causing the following error:
```
pq: missing FROM-clause entry for table \"xyz.person\"
```
I added support to schema name:
```
INSERT INTO "xyz"."person" .... RETURNING "xyz"."person"."idperson"
```

### DIFF
--- a/postgres.go
+++ b/postgres.go
@@ -5,6 +5,7 @@ import (
 	"database/sql/driver"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @seriallink 

Thank you for your PR.

I just did some different things that fixed the issue.

seriallink:
Thanks, @jinzhu

PR #652, opened 2015-09-08
Refinements to cleanup tests for MS SQL dialect support

biju-kalissery:
I have committed the following changes in my version.

Added method 'EnableIdentityInsert' to the Dialect interface and implemented it for mssql. This helps us to temporarily enable IDENTITY column inserts for specified table for MS SQL.

Changed the index in auto-migration test table (BigEmail) from unique index to normal index. MS SQL does not allow multiple NULL values in a column with unique index.

Skipped TestCreateWithNoGORMPrimayKey in create_test.go for mssql dilect because MS SQL generates error if we check for identity value for tables that does not have Id column.

Cleaned up error handling/reporting in various _test.go files to easily see the reason behind test failures.

Details and pending issues are listed at
https://github.com/jinzhu/gorm/issues/647
Thanks,
Biju

### DIFF
--- a/common_dialect.go
+++ b/common_dialect.go
@@ -115,3 +115,8 @@ func (commonDialect) CurrentDatabase(scope *Scope) (name string) {
 	scope.Err(scope.NewDB().Raw("SELECT DATABASE()").Row().Scan(&name))
 	return
 }
+
+func (commonDialect) EnableIdentityInsert(db *DB, tableName string) *DB {
+	return db
+}
+
--- a/create_test.go
+++ b/create_test.go
@@ -4,6 +4,7 @@ import (
 	"reflect"
 	"testing"
 	"time"
+	"os"
 )
 
 func TestCreate(t *testing.T) {
@@ -57,6 +58,11 @@ func TestCreate(t *testing.T) {
 }
 
 func TestCreateWithNoGORMPrimayKey(t *testing.T) {
+	
+	if dialect := os.Getenv("GORM_DIALECT"); dialect == "mssql" {
+		t.Skip("Skipping this because MSSQL will return identity only if the table has an Id column")
+	}
+	
 	jt := JoinTable{From: 1, To: 2}
 	err := DB.Create(&jt).Error
 	if err != nil {
--- a/customize_column_test.go
+++ b/customize_column_test.go
@@ -32,12 +32,23 @@ func TestCustomizeColumn(t *testing.T) {
 	if scope.PrimaryKey() != col {
 		t.Errorf("CustomizeColumn should have primary key %s, but got %q", col, scope.PrimaryKey())
 	}
+	
+	tableName := "customize_columns"
+	idInsRes := DB.EnableIdentityInsert(&DB, tableName)
+	if idInsRes.Error != nil {
+		t.Errorf("Error while setting IDENTITY_INSERT ON for table:%v :%v", tableName, idInsRes.Error)
+	}
 
 	expected := "foo"
 	cc := CustomizeColumn{ID: 666, Name: expected, Date: time.Now()}
 
-	if count := DB.Create(&cc).RowsAffected; count != 1 {
-		t.Error("There should be one record be affected when create record")
+    res := DB.Create(&cc)
+	if res.Error != nil {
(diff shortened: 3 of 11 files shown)


jinzhu:
Hello @biju-kalissery 

Thank you for your PR.

> Added method 'EnableIdentityInsert' to the Dialect interface and implemented it for mssql. This helps us to temporarily enable IDENTITY column inserts for specified table for MS SQL.

I think we might should not add it, might be better to call it when you initialized a mssql database.

> Changed the index in auto-migration test table (BigEmail) from unique index to normal index. MS SQL does not allow multiple NULL values in a column with unique index.

Thank you for let me know the information, but we should fix the place generate NULL value for it. So we still tested unique index.

And thank you for your other changes.

biju-kalissery:
@jinzhu , 

Thank you for the feedback.
Regarding 'EnableIdentityInsert' : SQL Server allows this to be done only at a table level, not for the whole database. I am open to any other suggestions you may have.  Since it is rare that we insert values into an auto-generated IDENTITY column, we could also consider skipping this test for SQL server.   But, that means we will not be testing the IDENTITY INSERT for SQL Server.  Is that what you prefer?

Regarding NULL values in auto-migration test table (BigEmail):   The NULL values are caused because current test code is adding new columns while migrating Email table to BigEmail table.  We have two options for fixing this.
Option 1:  Make the index on the new column non-unique (this is the fix I did)
Option 2: Delete all rows from the table before calling AutoMigrate. This will work, but we are not really testing AutoMigrate if we do this.

Thanks!

Biju

jinzhu:
Hi @biju-kalissery 

We could add `IDENTITY INSERT` here: https://github.com/jinzhu/gorm/blob/master/main_test.go#L60 , and also could have some tests for this.


> Regarding NULL values in auto-migration test table (BigEmail): The NULL values are caused because current test code is adding new columns while migrating Email table to BigEmail table. We have two options for fixing this.
> Option 1: Make the index on the new column non-unique (this is the fix I did)
> Option 2: Delete all rows from the table before calling AutoMigrate. This will work, but we are not really testing AutoMigrate if we do this.

Option 1 sounds better to me.

Thank you.

biju-kalissery:
@jinzhu ,

Changing main_test.go may not be ideal because we will have to create the test table to set IDENTITY_INSERT.  Here is what that code will look like.

```
case "mssql":
	fmt.Println("testing mssql...")
	db, err = gorm.Open("mssql", "server=SERVER_HERE;database=rogue;user id=USER_HERE;password=PW_HERE;port=1433")
        db.DropTable(&CustomizeColumn{})
	db.AutoMigrate(&CustomizeColumn{})
        tableName := "customize_columns"
	idSql := "SET IDENTITY_INSERT " + tableName + " ON"
	res := DB.Exec(idSql)
	if res.Error != nil {
              t.Errorf("Error while setting IDENTITY_INSERT ON for table:%v :%v", tableName, res.Error)
	}
default:
       ....
```

As you can see, this does not look good.

It looks like you are reluctant to add the new EnableIdentityInsert method to the Dialect interface.  There is another way to do this without adding the new method. We can check the GORM_DIALECT before trying to insert into the identity column.  The change will be limited to the customize_column_test.go file, TestCustomizeColumn method. 
 https://github.com/jinzhu/gorm/blob/master/customize_column_test.go#L35

Please let me know if you prefer this one instead.

```
if "mssql" == os.Getenv("GORM_DIALECT") {
	tableName := "customize_columns"
	t.Logf("Setting IDENTITY_INSERT ON for table %v", tableName)
	idSql := "SET IDENTITY_INSERT " + tableName + " ON"
	res := DB.Exec(idSql)
	if res.Error != nil {
            t.Errorf("Error while setting IDENTITY_INSERT ON for table:%v :%v", tableName, res.Error)
		}
	}
```

Thanks,

Biju

jinzhu:
Hi @biju-kalissery 

Really sorry for the late response, I think I prefer the second one. limit the change to `customize_column_test.go`, ideas?

jinzhu:
I have tried to port some changes made in this pr into master, as I don't have a mssql for me for testing, if there are still something not covered, please reopen another pr.

Thank you.

PR #666, opened 2015-09-22
Support the AUTOINCREMENT keyword on SQLite.

vibhav:
Omit the "PRIMARY KEY" clause at the end of the create statement if
any column's SQL tags already contains the PRIMARY KEY constraint.

SQLite requires the AUTOINCREMENT keyword to be specified on columns
with an INTEGER type and a PRIMARY KEY constraint: 
https://www.sqlite.org/autoinc.html.


vibhav:
@jinzhu: I'm open to any suggestions on feedback that you have on this pull request.

jinzhu:
HI @vibhav 

Thank you for your PR.

PR #669, opened 2015-09-24
Fix panic: reflect.Value.Addr of unaddressable value

nkovacs:


jinzhu:
Thank you for your PR.

PR #672, opened 2015-09-27
do not execute buildSelectQuery twice

ivpusic:
if buildSelectQuery() is executed twice, then we get more values in SqlVars than expected


jinzhu:
Oh, thank you for that.

PR #691, opened 2015-10-12
Fix FirstOrXXX access invalid memory address

athurg:
*DB is cloned, we should ALWAYS use the cloned pointer instead of original one.


jinzhu:
Thank you.

PR #695, opened 2015-10-16
Fix where clause for string primary key when query value is numeric and starts with zero

henriquemenezes:
Bug: Where condition when primary key is String and the query value start with zeros. For example: 
```go
type AddressByZipCode struct {
  ZipCode    string    `gorm:"primary_key:yes"`
  Address    string
}
```
When I search for a Address with `ZipCode = "00501"`:
```go
DB.First(&address, "00501")
```
The built SQL query with WHERE condition is:

```sql
  SELECT  * FROM "address_by_zip_codes"  WHERE ("zip_code" = '501') ORDER BY "address_by_zip_codes"."zip_code" ASC LIMIT 1
```
But the correct query is:
  ```sql
  SELECT  * FROM "address_by_zip_codes"  WHERE ("zip_code" = '00501') ORDER BY "address_by_zip_codes"."zip_code" ASC LIMIT 1
```

### DIFF
--- a/query_test.go
+++ b/query_test.go
@@ -64,6 +64,22 @@ func TestUIntPrimaryKey(t *testing.T) {
 	}
 }
 
+func TestStringPrimaryKeyForNumericValueStartingWithZero(t *testing.T) {
+	type AddressByZipCode struct {
+		ZipCode   string `gorm:"primary_key"`
+		Address   string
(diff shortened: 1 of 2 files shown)


jinzhu:
Thank you for your PR.

Issue #459, opened 2015-04-16
AutoMigrate  fail to check the existing field


davyzhang:
I don't know why but I get a table in sqlite_master table described like this:

```
SELECT * 
  FROM sqlite_master
 WHERE tbl_name = "ads"
```
result

```
table	ads	ads	169	CREATE TABLE ads (id integer, created_at datetime, updated_at datetime, deleted_at datetime, facebook_ad_id varchar (255), tag varchar (255), name varchar (255), game_id integer, status varchar (255), ex_rate integer, PRIMARY KEY (id))
```

note there's no double quote of any field name

so it failed on gorm's check:
```
SELECT count(*) FROM sqlite_master WHERE tbl_name = "ads" AND (sql LIKE '%("id" %' OR sql LIKE '%,"id" %' OR sql LIKE '%( id %' OR sql LIKE '%, id %');
```

You can check the db file here:
https://dl.dropboxusercontent.com/u/84845057/ads

PR #704, opened 2015-10-24
sqlite3: fix HasColumn - problem with detecting columns

KarolBedkowski:
Handle table definitions like ", \"col"" and ",col".

Fix: #459


jinzhu:
Thank you for your PR.

PR #706, opened 2015-10-28
add binary UUID support to postgres

sulnedinfind:
Check if the value is a byte array of 16 bytes and its type is named "uuid" or "guid", and if so, use postgres `uuid` type.


jinzhu:
Thank you for your PR.

brunoksato:
this PR broken 

    EncryptedPassword []byte         `json:"-" sql:"encrypted_password;not null"`

PR #708, opened 2015-10-31
Fix SqlTag() in case of ByteSlice

haruyama:
Before https://github.com/jinzhu/gorm/commit/3cb95264dff572434ed72977d78f16ab4fe1dc83,
SqlTag() returns "bytea" for []byte FIeld with no tag.

For example,
```
type Model struct {
   Field [] byte
}
```

After https://github.com/jinzhu/gorm/commit/3cb95264dff572434ed72977d78f16ab4fe1dc83,
SqlTag() returns error if not specifying explicit tag.

### DIFF
--- a/postgres.go
+++ b/postgres.go
@@ -53,7 +53,7 @@ func (postgres) SqlTag(value reflect.Value, size int, autoIncrease bool) string 
 			return "hstore"
 		}
 	default:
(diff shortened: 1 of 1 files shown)


jinzhu:
Thank you for your PR

PR #709, opened 2015-10-31
Fix panic() in isUUID()

haruyama:
value with value.Kind() == reflect.Slice does not have value.Type().Len()

#708 is incomplate fix. I'm sorry.


jinzhu:
Thank you for your PR.

PR #717, opened 2015-11-06
Break [] byte, I was breaking another case

brunoksato:
this PR #706 broken another case
Example: 
EncryptedPassword []byte  `json:"-" sql:"encrypted_password;not null"`


jinzhu:
Thank you for your PR.

PR #720, opened 2015-11-10
Modify count detection regex to match parenthesis instead of starting a capture.

### LINKED ISSUE #679: Bug with Order() for some queries
In function orderSql() in scope_private.go you have code:
if len(scope.Search.orders) == 0 || scope.Search.countingQuery {
    return ""
}

I cannot understand why .Order() for my query didn't work.
I've found these code in selectSql():
scope.Search.countingQuery = hasCountRegexp.MatchString(sql)

So, regexp
var hasCountRegexp = regexp.MustCompile(`(?i)count(.+)`)
is wrong for queries like

SELECT account FROM table;

Please, change it little bit to match only count(*). For example:

-var hasCountRegexp = regexp.MustCompile(`(?i)count(.+)`)
+var hasCountRegexp = regexp.MustCompile(`(?i)count\((.+)`)

The bug: I cannot apply Order() method to query like SELECT account FROM table.

jseriff:
Fixes #679

### DIFF
--- a/scope_private.go
+++ b/scope_private.go
@@ -208,7 +208,7 @@ func (scope *Scope) whereSql() (sql string) {
 	return
 }
 
-var hasCountRegexp = regexp.MustCompile(`(?i)count(.+)`)
+var hasCountRegexp = regexp.MustCompile(`(?i)count\(.+\)`)
 
 func (scope *Scope) selectSql() string {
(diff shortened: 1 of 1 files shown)


jinzhu:
Thank you for your PR.

PR #732, opened 2015-11-21
Enable to decline antuincrement on ID field

otiai10:
@jinzhu 

Thank you always for this useful library. It's very helpful for daily work and also my private work.

# request

These days, I faced to data model with `ID` field (it is `ID` actually, semantically) but it's not-zero-origin  and not-sequential.

I want to name this field `ID` but not want to be auto-incremental.

# solution

I found your code is already extensible enough! So I modified `generateSqlTag` func a bit. In this pull-request, I use `NO` for adhoc spell, but it's just an idea.

Please tell me your idea, or I'm very glad if you merge this request. 

Thank you

----

- [x] go test ./...
- [ ] review

### DIFF
--- a/model_struct.go
+++ b/model_struct.go
@@ -435,9 +435,12 @@ func (scope *Scope) generateSqlTag(field *StructField) string {
 			size, _ = strconv.Atoi(value)
 		}
 
-		_, autoIncrease := sqlSettings["AUTO_INCREMENT"]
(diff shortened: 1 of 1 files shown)


jinzhu:
HI @otiai10 

Thank you for your pull request, btw, I changed the `no` to `false`.

otiai10:
Thank you!

PR #735, opened 2015-11-21
Fix update of fields with a default value

galeone:
Thix PR fixes https://github.com/jinzhu/gorm/issues/731 and https://github.com/jinzhu/gorm/issues/730

As I specified in this comment: https://github.com/jinzhu/gorm/commit/2a46856d5238837a5f65f4a05181a6685335a4af#commitcomment-14500486

When an update is done on a field with an associated default value, and the update set this field to blank, gorm do the right thing (as it was before, when I create the default value support).

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -197,4 +197,10 @@ func (c *callback) sort() {
 	c.rowQueries = sortProcessors(rowQueries)
 }
 
+func ForceReload(scope *Scope) {
+	if _, ok := scope.InstanceGet("gorm:force_reload"); ok {
+		scope.DB().New().First(scope.Value)
+	}
+}
+
 var DefaultCallback = &callback{processors: []*callbackProcessor{}}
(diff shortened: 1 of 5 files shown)


jinzhu:
Hi @galeone 

Thank you for your PR.

So the question is if a field has default value, should we allow update it to a blank value? I think it should be yes, we should.

Just imagine a field's type is bool, and its default value is true, of course, we should have the ability to set it to false, yes?

galeone:
Yes, we should. But we either should handle the case of the update of a field (not having a default value), based on a struct containing the id.

Here's an example:

What I want `UPDATE comments SET message = 'edited' WHERE id = 9`
What I get: `UPDATE comments SET message = 'edited', "time" = '0001-01-01T00:00:00Z' WHERE id = 9`

```go
comment.Message = "edited"
comment.ID = 9 // an user gives the ID
db.Save(&comment)
```

Writing this post I think I've found a solution, that can handle both the cases.

We can generate this sql,
`UPDATE comments SET message = 'edited', "time" = DEFAULT WHERE id = 9`

In that way, if the `field.IsBlank || field.HasDefaultValue` we can generate the right query.
Do you think is a good solution?

galeone:
Allright, I fixed this. I'm opening a new pull request.

The modification works well with every dbms in the test file.

PR #737, opened 2015-11-22
Fix update of fields with a default value

galeone:
Now we can deal with the update of field having a default value even if this value is the "blank" value.

For example.
```gp
 type Animal struct {
     // ... see the changes in the tests
     Cool       bool `sql:"default:false"`
 }
```

This struct have a default value for the boolean attribute, that's equals to the Zero value of the boolean type.

Now we can set without any problem the `Cool` attribute to true or to false (when is false is replaced with the default value of the tag field).

Referring to my comment in #735:

What I want `UPDATE comments SET message = 'edited' WHERE id = 9`

Now what I get: `UPDATE comments SET message = 'edited', "time" = (now() at time zone 'utc') WHERE id = 9`

that's better of what I previously wanted, since I can see the default value being inserted in the query and is more clear to read and understand.

I get this query because the declaration of the Time field is `Time time.Time `sql:"default:(now() at time zone 'utc')".

P.S: I tough I could use the sql syntax `set field=DEFAULT` but that's not supported in DBMS like `sqlite3`, thus I fallback to this solution that works well with every DBMS.

Regards

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -2,6 +2,9 @@ package gorm
 
 import (
 	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
 )
 
 type callback struct {
@@ -201,6 +204,56 @@ func ForceReload(scope *Scope) {
 	if _, ok := scope.InstanceGet("gorm:force_reload"); ok {
 		scope.DB().New().First(scope.Value)
 	}
+}
+
+func escapeIfNeeded(scope *Scope, value string) string {
+	trimmed := strings.TrimSpace(value)
+	// default:'string value' OR
+	if (strings.HasPrefix(trimmed, "'") && strings.HasSuffix(trimmed, "'")) ||
+		strings.HasSuffix(trimmed, ")") { //sql expression, like: default:"(now() at timezone 'utc') or now() or user_defined_function(parameters.. )
+		return trimmed
+	}
+
+	lowered := strings.ToLower(trimmed)
+	if lowered == "null" || strings.HasPrefix(lowered, "current_") { // null and other sql reserved keyworks (used a default values) can't be placed between apices
+		return lowered
+	}
+	return scope.AddToVars(trimmed) // default:'something'  like:default:'false' should be between quotes (what AddToVars do)
+}
+
+func handleDefaultValue(scope *Scope, field *Field) string {
+	if field.IsBlank {
+		defaultValue := strings.TrimSpace(parseTagSetting(field.Tag.Get("sql"))["DEFAULT"])
+		switch field.Field.Kind() {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+			if numericValue, err := strconv.ParseInt(defaultValue, 10, 64); err == nil {
+				if numericValue != field.Field.Int() {
+					return escapeIfNeeded(scope, fmt.Sprintf("%d", field.Field.Int()))
+				} else {
+					return escapeIfNeeded(scope, defaultValue)
+				}
+			}
+		case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+			if numericValue, err := strconv.ParseUint(defaultValue, 10, 64); err == nil {
+				if numericValue != field.Field.Uint() {
+					return escapeIfNeeded(scope, escapeIfNeeded(scope, fmt.Sprintf("%d", field.Field.Int())))
+				} else {
+					return escapeIfNeeded(scope, defaultValue)
+				}
+			}
+		case reflect.Bool:
+			if boolValue, err := strconv.ParseBool(defaultValue); err == nil {
+				if boolValue != field.Field.Bool() {
+					return escapeIfNeeded(scope, fmt.Sprintf("%t", field.Field.Bool()))
+				} else {
+					return escapeIfNeeded(scope, defaultValue)
+				}
+			}
+		default:
+			return escapeIfNeeded(scope, defaultValue)
+		}
+	}
+	return scope.AddToVars(field.Field.Interface())
 }
 
 var DefaultCallback = &callback{processors: []*callbackProcessor{}}
--- a/callback_create.go
+++ b/callback_create.go
@@ -29,12 +29,15 @@ func Create(scope *Scope) {
 			if scope.changeableField(field) {
 				if field.IsNormal {
 					if !field.IsPrimaryKey || (field.IsPrimaryKey && !field.IsBlank) {
-						if !field.IsBlank || !field.HasDefaultValue {
+						if !field.HasDefaultValue {
 							columns = append(columns, scope.Quote(field.DBName))
 							sqls = append(sqls, scope.AddToVars(field.Field.Interface()))
-						} else if field.HasDefaultValue {
+						} else {
+							columns = append(columns, scope.Quote(field.DBName))
+							sqls = append(sqls, handleDefaultValue(scope, field))
 							scope.InstanceSet("gorm:force_reload", true)
 						}
+
 					}
 				} else if relationship := field.Relationship; relationship != nil && relationship.Kind == "belongs_to" {
 					for _, dbName := range relationship.ForeignDBNames {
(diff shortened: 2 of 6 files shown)


galeone:
I'm closing this PR since I noticed a bug. I'll fix it and I'll create a new PR.

galeone:
OK I've reopened the PR. The expressions like (now() at timezone 'utc') as default value are handled correctly, as are any other type. Tests works well.

In addiction, I fixes a test in create_test.go, changing `if user.CreatedAt != newUser.CreatedAt` to `if user.CreatedAt.Format(time.RFC3339Nano) != newUser.CreatedAt.Format(time.RFC3339Nano)` since when testing with sqlite, due to some precision issue, the test failed.
Now, using a proper conversion to a common type, everything works as aspected.

jinzhu:
Hi @galeone 

So now it works like this:

```go
 type Animal struct {
     // ... see the changes in the tests
     Name    string
     Cool       bool `sql:"default:false"`
 }

db.Model(&animal).Update(Animal{Name: "hello"})
// will update both name and cool
```

If so, I don't think this is the right behaviour.

galeone:
That's what is just happening.

Before the PR, the generated query is
`UPDATE "animals" SET "name" = 'hello', [..] "cool" = 'false' WHERE ...`
Where false is the value of the `Cool` field (since is not initialized, see my comment about the `time` fields that's initialized with the blank value of the Time type, that's wrong).

After the PR, the generated query is the same, *but*
`"cool" = 'false'` is generated because the `Cool` field is blank and the default value is false, thus nothing changed from this point of view.

If the default value of Cool was `default:true`, the query will generate `cool = false` even if the default value is true, since this case is handled (see tests).

So, when you set a field to its blank value, gorm will replace its content to the default value found in the sql tag of the field.

Other example:

Before the PR, if I have
```go
comment.Message = "edit"
comment.ID = 9
db.Save(&comment)
```
The generated query have  the `time` field equals to `0001-01-01T00:00:00Z`:
`update comments set message = 'edit' time = '0001-01-01T00:00:00Z' where id = 9`
This query is absolutely wrong, don't you think?

After the PR the same go code generates this query:
`update comments set message = 'edit' time = (now() at time zone 'utc') where id = 9`
That's the expected (right) behavior.

pedromorgan:
Does this mean the database is created with default value in coloumn spec. .which would be handy..

galeone:
In this latest commit I fixed https://github.com/jinzhu/gorm/issues/689

And I merged the logic between update and insert.

pedromorgan:
Actually bumped into problem, where I want some field to have defaults, 
eg
JobStatus = 10
New = true
and want this in the DB, rather than gorm.. but db is being created from gorm..

jinzhu:
Hi @galeone 

I still think this is not the right behaviour, so when I update a record, for those empty fields, it will use the default value? 

That's means I am not possible to change a field's value to the empty value, for example, if a bool field have default value `true`, then I could never change the value to false?

galeone:
@jinzhu look at the test I added. I kept a test from https://github.com/jinzhu/gorm/issues/689, that's the same situation you describe. As you can see, when you update a record with a value, and this value is different from the default one, the desired value is added. Otherwise the default one is used.

Have a look at the function `handleDefaultValue` to understand: https://github.com/jinzhu/gorm/pull/737/files#diff-6a4cdb31ba172474c25dc7609d662f39R224

jinzhu:
I just checked the tests, so behaviours will be changed to:

```
	type Animal struct {
		Counter    uint64    `gorm:"primary_key:yes"`
		Name       string    `sql:"DEFAULT:'galeone'"`
	}

	var animal Animal
	DB.Save(&animal)
	animal.Name // => "galeone"

	animal.Name = "amazing horse"
	DB.Save(&animal)
	animal.Name // => "amazing horse"

	animal.Name = ""
	DB.Save(&animal)
	animal.Name // => "galeone"
        // current behaviour will change the name to ""
``` 

Yes?

galeone:
@jinzhu at the moment no. When you insert/update a filed with an empty string, the default value is used. The reason is that a default value can be a sql function call, thus this value must be inserted.

Otherwise, how can you handle this case?

For other fields like boolean and numeric, if you insert a `false` and the default value for that field is `true` than true is inserted.

But, do you want a different behavior in the handling of default values in insertion and update?

I could keep this logic for the insertion (during a INSERT if a field is empty, the default value should be inserted. Otherwise what's the meaning of the default value support? I created it for this reason primarily.) and your logic for the update.

But, believe me, users do not find this behavior logic (and the prove is the amount of issues opened after your changes on the default value handling).

raitucarp:
You should reconsider this @jinzhu, for example I have struct such this:
```go
type User struct {
    ID        int       `json:"id" gorm:"primary_key"`
    CreatedAt time.Time  `json:"created_at"`
    UpdatedAt time.Time  `json:"updated_at"`
    DeletedAt *time.Time `json:"deleted_at"`
    Username  string       `json:"username" gorm:"primary_key" `
   Bio              string      `sql:"type:text;default:null;"`
    Facebook  string      ` sql:"default:null;unique"`
    Twitter   string          ` sql:"default:null;unique"`
}
```
The Facebook and Twitter field is typically null. User can update it later time.

If I do partial update for example I update bio:
```go
// this is from /post with binding
db.Where("username ?", "jinzhu").First(&user)
user.Bio = "This is me"
db.Save(&user)
```

The dblog return such this one:
```sql
UPDATE "users" SET "created_at" = '2016-02-15T12:54:24+07:00', "deleted_at" = '<nil>', "bio" = 'This is me', "twitter" = '', "facebook" = '', "updated_at" = '2016-02-15T13:32:22+07:00', WHERE ("users".deleted_at IS NULL OR "users".deleted_at <= '0001-01-02') AND ("id" = '5') AND ("username" = 'jinzhu')
```
Which is causes more conflict. Because facebook or twitter column is unique, and they are not primary key, because it's optional. Every user has unique facebook or twitter username. I did partial update, but I got facebook with empty string '' values updated.

When another user register for first time or updating facebook or twitter after jinzhu, it's disaster.
``` pq: duplicate key value violates unique constraint "users_facebook_key" ```

What do you think?

**Comment by jinzhu** [comment]
You should not use `string` as type then. you could use `*string` or `NullString` for `Facebook`, `Twitter`

**Comment by galeone** [comment]
And what about if I have a default value that's not null for the empty fields?

Eg, if the @raitucarp structure looks like:
```go
type User struct {
    ID        int       `json:"id" gorm:"primary_key"`
    CreatedAt time.Time  `json:"created_at"`
    UpdatedAt time.Time  `json:"updated_at"`
    DeletedAt *time.Time `json:"deleted_at"`
    Username  string       `json:"username" gorm:"primary_key" `
   Bio              string      `sql:"type:text;default:null;"`
    Facebook  string      ` sql:"default:null;unique"`
    Twitter   string          ` sql:"default:"sql_function()";unique"`
}
```

What the user expects is the execution of `sql_function()` not the insert of an empty value.

In that case, it doesn't matter what's the type of Twitter, if its  a `string` or a `*sql.NullString` or whatever.
The generated sql will be wrong.

jinzhu:
the `sql_function` will be call when creating, but I think it should never be run when updating.

galeone:
Why? There must be a way to call that function when updating a field, otherwise the default value is useless IMHO
Furthermore, the different logic behind the handling of default values in update and create generates confusion

jinzhu:
Added more documents about this in https://github.com/jinzhu/gorm/tree/v1.0_dev#update

Feedback welcome in #794

raitucarp:
What's the cost of updating two fields? Suppose that I have 10 fields? for example:
```go
db.Model(&user).Update("name", "hello")
//// UPDATE users SET name='hello', updated_at='2013-11-17 21:34:10' WHERE id=111;
db.Model(&user).Update("facebook", "bababa")
//// UPDATE users SET facebook='bababa', updated_at='2013-11-17 21:34:10' WHERE id=111;
```
It's send two update isn't? Should I use tx, or just normal db?

jinzhu:
@raitucarp 

You should use `db.Model(&user).Updates(map[string]interface{}{"name": "hello", "facebook": "bababa"})`

PR #742, opened 2015-11-25
Change the order between SaveAfterAssociations AND ForceReloadAfterCreate

EthanLiou:
If there're some columns having default value, we will reload it after saving it. So the AfterFind will be triggered at that moment.
Sometimes, we do some operations on associations columns in AfterFind callback. Currently, we reload it first, so we cannot get related data at that moment. Here's the example. I get the emails related to the user in AfterFind callback.
Because the flow is 
* **Create**
* **ForceReloadAfterCreate** After triggering AfterFind, we will get related emails from email table, but unfortunately we haven't inserted related emails now. So the Emails becomes empty
* **SaveAfterAssociations** Because Emails becomes empty, so we do not do any actions. 
So I think we should reload it after handling associations.
```go
type Email struct {
	ID     int
	UserID int
	Email  string
}

type User struct {
	ID     int
	Name   string
	Emails []Email
	Age    int `sql:"default:18"`
}

func (u *User) AfterFind(db *gorm.DB) (err error) {
	em := make([]Email, 0)

	if err := db.Model(&u).Related(&em).Error; err != nil {
		return err
	} else {
		u.Emails = em
	}
	return nil
}

func main() {
	user := User{
		Name:   "Ethan",
		Emails: []Email{{Email: "xxx@gmail.com"}},
	}
	db.Create(&user)
}
```

### DIFF
--- a/callback_create.go
+++ b/callback_create.go
@@ -114,8 +114,8 @@ func init() {
 	DefaultCallback.Create().Register("gorm:save_before_associations", SaveBeforeAssociations)
 	DefaultCallback.Create().Register("gorm:update_time_stamp_when_create", UpdateTimeStampWhenCreate)
 	DefaultCallback.Create().Register("gorm:create", Create)
-	DefaultCallback.Create().Register("gorm:force_reload_after_create", ForceReloadAfterCreate)
 	DefaultCallback.Create().Register("gorm:save_after_associations", SaveAfterAssociations)
+	DefaultCallback.Create().Register("gorm:force_reload_after_create", ForceReloadAfterCreate)
 	DefaultCallback.Create().Register("gorm:after_create", AfterCreate)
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @EthanLiou 

I think the problem is: we should not call find callbacks in the `ForceReloadAfterCreate`, but not just change the order?

EthanLiou:
Hi, I had the same idea with you before. But I think it's necessary. Because if we need to do some operations for the columns with default value on AfterFind. without calling callback, the value might be wrong. So I think we can change the order first.

jinzhu:
Hi @EthanLiou 

I did some updates for the `ForceReloadAfterCreate`, will only reload those columns having default value in  `ForceReloadAfterCreate`, I guess this should fixed your issue without this pull request.

Could you please check it out? Thank you for your PR anyway.

PR #764, opened 2015-12-15
Fixed a bug with invalid zero value in preload

RichardKnop:
When using pointers in model structs, there is an edge case.

When using preload to eagerly preload a many to many relationship and
the main row is not found by a primary key, use of reflect.Indirect in
preload.go results in a zero value and the subsequent call to
FieldByName panics the program.

Consider the following example:

```go
package somepackage

import (
    "database/sql"
    "github.com/jinzhu/gorm"
)

// Permission ...
type Permission struct {
    gorm.Model
    Name string
}

// Role ...
type Role struct {
    gorm.Model
    Name        string        `sql:"type:varchar(20);unique;not null"`
    Permissions []*Permission `gorm:"many2many:role_permissions"`
}

// User ...
type User struct {
    gorm.Model
    RoleID        sql.NullInt64 `sql:"index;not null"`
    Role          *Role
    Username     sql.NullString `sql:"type:varchar(100)"`
    Password      sql.NullString `sql:"type:varchar(100)"`
}
```

This will panic the program:

```go
id := 0 // invalid primary key!
user := new(User)
userNotFound := s.db.Preload("Role.Permissions").First(user, id).RecordNotFound()
```

### DIFF
--- a/preload.go
+++ b/preload.go
@@ -10,8 +10,15 @@ import (
 
 func getRealValue(value reflect.Value, columns []string) (results []interface{}) {
 	for _, column := range columns {
-		if reflect.Indirect(value).FieldByName(column).IsValid() {
-			result := reflect.Indirect(value).FieldByName(column).Interface()
+		pointedValue := reflect.Indirect(value)
+		// If v is a nil pointer, Indirect returns a zero Value!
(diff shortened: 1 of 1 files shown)


jinzhu:
Thank you for your pull request

PR #766, opened 2015-12-16
There was still an issue with preloading and pointers.

RichardKnop:
This was actually broken with the refactoring. I added a test to catch the other place the error was occurring. See the new `TestManyToManyPreloadForNestedPointer` test in preload_test.go.

See the issue also happens when there is one to one relationship on top of many to many relationship. For example:

```go
	type (
		Level1 struct {
			ID    uint
			Value string
		}
		Level2 struct {
			ID      uint
			Value   string
			Level1s []*Level1 `gorm:"many2many:levels;"`
		}
		Level3 struct {
			ID       uint
			Value    string
			Level2ID sql.NullInt64
			Level2   *Level2
		}
	)
```

And then doing:

```go
Preload("Level2.Level1s")
```

### DIFF
--- a/preload_test.go
+++ b/preload_test.go
@@ -771,6 +771,9 @@ func TestManyToManyPreloadForPointer(t *testing.T) {
 		panic(err)
 	}
 
+	var got5 Level2
+	DB.Preload("Level1s").First(&got5, "value = ?", "bogus")
+
 	var ruLevel1 Level1
 	var zhLevel1 Level1
 	DB.First(&ruLevel1, "value = ?", "ru")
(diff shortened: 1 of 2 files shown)


RichardKnop:
Wait, I'll fix the broken build.

RichardKnop:
The new test covers the nested pointer case now.

preload_test.go
```diff
@@ -783,6 +786,109 @@ func TestManyToManyPreloadForPointer(t *testing.T) {
...
+		t.Errorf("got %s; want %s", toJSONString(got3), toJSONString([]Level3{got, got2}))
+	}
+
+	var got4 []Level3
+	if err := DB.Preload("Level2.Level1s", "value IN (?)", []string{"zh", "ru"}).Find(&got4, "value IN (?)", []string{"Bob", "Tom"}).Error; err != nil {
+		panic(err)
+	}
+
+	var got5 Level3
+	DB.Preload("Level2.Level1s").Find(&got5, "value = ?", "bogus")
```
RichardKnop: Especially this line is important as it tests the specific test case I described before. It panics without my fix.

jinzhu:
Hi @RichardKnop 

Thank you for your PR.

PR #767, opened 2015-12-18
clean up wording in README

levinalex:
made the paragraph about transactions in callbacks more straightforward


jinzhu:
Hi @levinalex 

Thank you for your PR.

PR #779, opened 2016-01-04
scope: Catch error for db.Commit()

shenli:
db.Commit() will return error but it is ignored.
We find this problem in this issue https://github.com/pingcap/tidb/issues/822.


jinzhu:
Thank you for your PR.

PR #788, opened 2016-01-06
Fix polymorphic test

jcscottiii:
One assertion comment says look for association with dog but the code
still uses cat. This aligns the code to now look at the association of
the dog.


jinzhu:
Thank you for the PR.

PR #794, opened 2016-01-10
The first public release

jinzhu:
This is the pull request for the first public release of GORM, which is scheduled for March of 2016.

It has been more than two years since the first commit of GORM, and we keep backwards compatibility seriously in the last two years, but in this release, I am going to break few APIs, although most applications should not be affected, I will provide a migration checklist for everyone.

I don't really have a todo list, but will focus on changes to make developers feels more friendly and some performance improvements, If you have any requests, please post a comment or issue.


douglarek:
The latest commit d92c5db  does not work when scope.SetColumn

PR #803, opened 2016-01-16
Fixes querying with inline map when a value is nil

jkongie:
Previously, querying with an inline map created 'where conditions' which always used the equality operator for comparison. This is problematic when trying to compare find records which are `NULL`.

This changes the inline map query build condition to use `IS NULL` instead of the equality operator when the provided value is `nil`. The inverse `NOT` has also been update to produce an `IS NOT NULL` condition.


jinzhu:
Thank you for your PR.

jkongie:
没问题

Issue #729, opened 2015-11-19
Make the DB an interface for easier testing?


danqing:
It will be great if the DB type can be turned into an interface, so users can mock the interface in tests as appropriate. Otherwise we will need to do wrappers etc. to achieve this, which isn't optimal.

Were there any reasons why it started as a struct?

Thanks!

jinzhu:
refer https://github.com/erikstmartin/go-testdb

alvarowolfx:
Will be so much better to test integration with ```gorm```if we have a interface to mock ( I use [testify](https://github.com/stretchr/testify) to do this). @jinzhu I think that using [go-testdb](https://github.com/erikstmartin/go-testdb) don't resolve the problem because by using it I'm testing ```gorm```again every time, seeing if it return the correct query or result and this is not the best way to test.

Mocking a interface to ```gorm``` we can expected the right methods on ```gorm```to be called  with the right params, trusting that the ```gorm``` library is already well tested. 

I tried to do this but because ```gorm``` have a chainable API, I could not mock the return of the methods, because they return the concrete implementation ```gorm.DB```, breaking a test that requires more then one method to be called in a chain. 

I forked the project to create a common interface, but unfortunatly will be introduce a breaking change to the library. Do we have a roadmap for ```gorm```?

cc @danqing

jinzhu:
Hi @alvarowolfx 

I could accept some break changes in the v1.0 release if it is valuable, refer: https://github.com/jinzhu/gorm/pull/794

I am thinking you could go-testdb to mock some queries for testing.  but really suggest you use the real database that you are using in production for testing.

jirfag:
@smikulcik @mykytanikitenko @imiskolee @sunfmin @levinalex
I use `sqlmock` to mock-test GORM queries. Example is [here](https://github.com/jirfag/go-queryset/blob/master/queryset/queryset_test.go), search e.g. `testUserSelectAll`.
First I tried `go-testdb` and custom `gorm.DB` wrapper to interface, but current solution with `sqlmock` is the simplest.

nmische:
There are other use case beyond testing where providing a DB interface would be useful. For example, I'm currently working on a project that provides a generic storage interface. It would be useful for that interface to have a `StorageEngine() interface{}` method, however I can't do that currently due to the lack of a DB interface. 

@jinzhu: Could providing a DB interface be reconsidered given alternate use cases beyond testing?

wahyd4:
[...]

PR #805, opened 2016-01-17
Database interface for easy mocking gorm

alvarowolfx:
Still WIP but I opened the PR just to show what I tried to explain on #729. You can see in the example folder a use case of the interface based testing. 

I made a mistake of developing on master branch, later I will change and merge with the branch v1.0_dev for better integration. 

Please @jinzhu let me know your opinion on this.

### DIFF
--- a/association.go
+++ b/association.go
@@ -23,7 +23,7 @@ func (association *Association) setErr(err error) *Association {
 
 func (association *Association) Find(value interface{}) *Association {
 	association.Scope.related(value, association.Column)
-	return association.setErr(association.Scope.db.Error)
+	return association.setErr(association.Scope.db.GetError())
 }
 
 func (association *Association) saveAssociations(values ...interface{}) *Association {
@@ -42,7 +42,7 @@ func (association *Association) saveAssociations(values ...interface{}) *Associa
 		// value has to been saved for many2many
 		if relationship.Kind == "many_to_many" {
 			if scope.New(reflectValue.Interface()).PrimaryKeyZero() {
-				association.setErr(scope.NewDB().Save(reflectValue.Interface()).Error)
+				association.setErr(scope.NewDB().Save(reflectValue.Interface()).GetError())
 			}
 		}
 
@@ -68,7 +68,7 @@ func (association *Association) saveAssociations(values ...interface{}) *Associa
 		if relationship.Kind == "many_to_many" {
 			association.setErr(relationship.JoinTableHandler.Add(relationship.JoinTableHandler, scope.NewDB(), scope.Value, reflectValue.Interface()))
 		} else {
-			association.setErr(scope.NewDB().Select(field.Name).Save(scope.Value).Error)
+			association.setErr(scope.NewDB().Select(field.Name).Save(scope.Value).GetError())
 
 			if setFieldBackToValue {
 				reflectValue.Elem().Set(field.Field)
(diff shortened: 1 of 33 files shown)


jinzhu:
Hi @alvarowolfx 

But from those changes, I think I am not a fan of mock testing, don't would like to increase GORM's complexity for that.

It might always good to use real database for testing, or let me know if you have any problems of that, to see I could help on it.

alvarowolfx:
What complexity are you talking about @jinzhu ?! Just changed some method name and the mocks package was autogenerated. As I sais still WIP and can be enhanced.

If you can give me a example of testing the same method that I wrote on the example folder, without mocks, I will like and thanks.

jinzhu:
@alvarowolfx what the problem of testing it with real data in real database?

alvarowolfx:
@jinzhu I don't know a way to test the example that I wrote without re-testing gorm itself, that why I asked for a example on how would do it. 

Testing with the database also makes the tests to run more slowly. And this article talks more about mocks : http://martinfowler.com/articles/mocksArentStubs.html 

I'm not trying to complicate things, just trying to apply technics already used in other languages to facilitate testing. Your work on gorm is awesome, i'm not trying to break this.

jinzhu:
I understand, and I have tried mock testing when using Ruby, but finally dropped it, think it doesn't really provide us any value, more time costing, and even get more bugs.

My understanding, unit tests is testing get this results with those inputs, I don't really care how it works, so I don't think you are re-testing GORM, when you writes those tests, you don't even need to know it is written with GORM or not. You are just testing it works like you want or not.

And seems like Martin Fowler also not a fan of this, also refer: https://www.thoughtworks.com/insights/blog/mockists-are-dead-long-live-classicists

ryanwalls:
@alvarowolfx Thanks for your effort on this.  I would have loved to see this merged.  Using interfaces to define complex operations is idiomatic go in my opinion.  Would have made unit testing our code much much easier.

RichardKnop:
I prefer functional tests. Just create a test database and run tests against it.

You would be surprised how many small edge case issues get caught when using real database in tests over the years.

I am a big fan of unit tests but when testing something like REST API I prefer functional tests ala Django/RoR.

ryanwalls:
@RichardKnop Agree on functional tests being important, but testing error cases would be easier with mocking.  I am encapsulating gorm interactions with a service class like you suggested.  The missing piece is a real unit test that verifies that I'm doing error handling correctly.  

e.g. 
Wrapping function:
```
func (s *service) CreateOrganization(organization *models.Organization) error {
	return s.db.Create(organization).Error
}
```  
No easy way to unit test that I am returning the error from this function.  But I can live with it because it's not a very complex method... but being a TDD aficionado, I get just a little sad.  ;)

littledot:
@alvarowolfx @jinzhu Putting testing ideologies aside, could we achieve the same results with a delegate wrapper? Seems like less code and also guarantees backwards-compatibility.

For example:
```go
package gormw
...

type DB interface {
        Create(value interface{}) DB
}

// Wrap wraps gorm.DB in an interface
func Wrap(db *gorm.DB) DB {
	return &gormw{db}
}

type gormw struct {
	w *gorm.DB
}

func (it *gormw) Create(value interface{}) DB {
	return Wrap(it.w.Create(value))
}

```

RichardKnop:
@ryanwalls Actually you can unit test with the service class you are using, just define an interface:

```go
interface ServiceInterface {
    CreateOrganization(organization *models.Organization) error
}
```

Then use [mockery](https://github.com/vektra/mockery) to create a mock struct defined by that interface. And use that in unit tests, you can mock what the service returns (errors for example) using the mock struct and something like [testify](https://github.com/stretchr/testify).

ryanwalls:
@RichardKnop Yep, that's what I do.  But you still aren't verifying that `CreateOrganization` actually returns the error from the gorm `Create` method.

PR #822, opened 2016-01-25
Remove regex guess on counting query

jseriff:
The hasCountRegexp and check in the selectSql was causing issues on queries with embedded count(*) in them.  For example, a query in the following format will fail to act as expected -

db.Table("table1").Select("*, (select count(*) from table2 where table1.id = table2.id)").Order("table1.name")

The sql generated will not include the order by clause, but will also not generate any errors.

The new behavior is that the query will attempt to run exactly as the developer entered it - if it is invalid, the developer will receive an error.  If the developer uses the built in gorm Count() method, any order by clauses will be ignored, and the correct count will be returned.

### DIFF
--- a/scope_private.go
+++ b/scope_private.go
@@ -220,8 +220,6 @@ func (scope *Scope) whereSql() (sql string) {
 	return
 }
 
-var hasCountRegexp = regexp.MustCompile(`(?i)count\(.+\)`)
-
 func (scope *Scope) selectSql() string {
 	if len(scope.Search.selects) == 0 {
 		if scope.Search.joins != "" {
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @jseriff 

This is helpful for most cases, but in some edge cases (which happened in an application I written), scaning the count into a struct, but not using the `Count` method, which will be broken.

I will think about this in the v1 release, to think if there is any better solution to resolve your/my problem

jseriff:
Thanks for taking a look.  

I think it would be good for the query modifying behavior that you need to be opt-in somehow, as without this PR that are queries that cannot be run at all with gorm.  So, for example, perhaps a WithoutOrder() method which removes the order from the query prior to a Find() or similar call.

PR #823, opened 2016-01-26
Fix README typo on Related usage.

taiyangc:
Fix a minor typo when showing Related usage with field name omit, so it doesn't confuse new users.


jinzhu:
Thank you for your PR.

Issue #829, opened 2016-02-03
Foreign key names too long


speijnik:
Some DBMS, like MySQL, support foreign key names only up to a certain length. The current approach
generates a name that depends on multiple factors, possibly generating a FK name that is rather long.

The idea here is to use a hash function to generate parts of the name, possibly making it shorter and not exceed limits on MySQL.

jinzhu:
Hi @speijnik 

I think change the default length might be a better solution, Refer https://bugs.mysql.com/bug.php?id=13942

PR #830, opened 2016-02-03
Use hash function when generating foreign key names

speijnik:
This ensures foreign key names are shorter in some cases. This is a suggested fix for #829

Signed-off-by: Stephan Peijnik <speijnik@anexia-it.com>

### DIFF
--- a/scope_private.go
+++ b/scope_private.go
@@ -1,6 +1,7 @@
 package gorm
 
 import (
+	"crypto/md5"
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @speijnik 

As I mentioned in that issue, I think change the default length might be a better solution.

I think it is always good to keep the foreign name be meaningful & readable.

Issue #831, opened 2016-02-03
Support conditional adding of foreign keys


speijnik:
As of now AddForeignKey tries to add a foreign key without checking if it exists before doing so, causing
an error to be returned.

It would be nice if AddForeignKey first checked if the FK exists and only add it when it does not exist yet.

jinzhu:
#832

PR #832, opened 2016-02-03
Implement conditional foreign key generation

speijnik:
For now, support is restricted to MySQL and PostgreSQL, while all other
DBMS default to the old behavior.

Ref #831 

Signed-off-by: Stephan Peijnik <speijnik@anexia-it.com>


jinzhu:
Lets think about this in the v1.0 release.

If we are going to support this, it has to been supported by all dialects, or will be thought as a bug by people.

PR #834, opened 2016-02-04
Fix documentation for Count() function example

Xiol:
Hi,

One of your Count function examples causes a panic for me on Go 1.5.3. For example:

```go
a.db.Find(Sms{}).Count(&count)
```

Changing this to a pointer to the struct fixes the problem:

```go
a.db.Find(&Sms{}).Count(&count)
```

This PR updates the documentation to show a pointer to the struct instead.


jinzhu:
Thank you for your PR

PR #835, opened 2016-02-04
Removed guessing numeric type from a string.

everclear:
This type of guess work is dangerous. It could lead to the situation where input is accepted from a query string for an primary key query, but is injected into the SQL string directly.

/api/user/10
db.Find(&user, "10")  // SELECT * FROM users WHERE id = 10;

/api/user/1=1
db.Find(&user, "1=1")  // SELECT * FROM users WHERE "1=1";

which is equivalent to 
// SELECT * FROM users

It shouldn't behave differently based on the content of the string passed to it. Especially when that has a security implication. If the user want's to pass a int, they should pass that type.

### DIFF
--- a/scope_private.go
+++ b/scope_private.go
@@ -17,10 +17,7 @@ func (scope *Scope) primaryCondition(value interface{}) string {
 func (scope *Scope) buildWhereCondition(clause map[string]interface{}) (str string) {
 	switch value := clause["query"].(type) {
 	case string:
-		// if string is number
-		if regexp.MustCompile("^\\s*\\d+\\s*$").MatchString(value) {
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @everclear 

Thank you for your pull request. I understand your concern.
I think this is what you should care in your application to avoid security problems.

Like gorm provide a `Exec` method allow you to run any SQL, this is not the reason for you use user's data without check as SQL directly.

Thank you any way.

everclear:
I don't see the benefit of guessing the type based on the content of a variable, when you are using a typed language. A developer using the code won't know that it will behave differently based on the contents of the string.

PR #846, opened 2016-02-10
Fixed a nested preload panic bug.

RichardKnop:
Fixed another nested preload bug I encountered.

There might be more bugs similar to this one, basically every time you try to do:

```
object.FieldByName
```

You should first check if `object` is a pointer:

```
if object.Kind() == reflect.Ptr {
	object = object.Elem()
}
```

I have also refactored a test from my previous PR a bit.


jinzhu:

PR #851, opened 2016-02-16
Update main.go

ElvizLai:
Ex:
```
type User struct {
	ID        int64
	Name      string `sql:"not null;unique"`
	Hobby     []string `sql:"-"`
	HobbyStr  string `sql:"type:text"`
}
u := &User{Name:"ElvizLai"}
fmt.Println(db.Last(u, "5x").RecordNotFound())//should be true but it return false
```
It has error: "Unknown column '5x' in 'where clause'", This maybe also belongs to "RecordNotFound"

### DIFF
--- a/main.go
+++ b/main.go
@@ -369,7 +369,7 @@ func (s *DB) NewRecord(value interface{}) bool {
 }
 
 func (s *DB) RecordNotFound() bool {
-	return s.Error == RecordNotFound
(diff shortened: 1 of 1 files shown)


jinzhu:
I think it should not belongs to record not found, record not found should only including record not found error

If you want to make sure no error happened, should check `db.Last(u, "5x").Error == nil`

PR #867, opened 2016-02-25
scope.IndirectValue() and nil pointer

derlaft:
This is (patrial) solution for bug qor/qor#161. I can still reproduce it (in creation form), but this may be a qor issue.


derlaft:
OK. Seems that qor calls ``db.Model(&link).Related(nil, "BadLink")``, so this is not a gorm issue.

PR #882, opened 2016-03-03
Fix "foreign key too long" bug

cristian-sima:
I used **SHA1**  to generate a unique hash using the identifier.
I ignore the underscores.
Then I use the hash rather than the identifier

This feature solves these bugs:

- [this](https://github.com/jinzhu/gorm/issues/881)
- [also this](https://github.com/jinzhu/gorm/issues/762)
- [and this](https://github.com/jinzhu/gorm/issues/829)

### DIFF
--- a/scope_private.go
+++ b/scope_private.go
@@ -1,6 +1,7 @@
 package gorm
 
 import (
+	"crypto/sha1"
 	"database/sql"
 	"database/sql/driver"
 	"fmt"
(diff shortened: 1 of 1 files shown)


danhardman:
Having a hash for a foreign key seems confusing. I would suggest sticking to what @jinzhu suggested in one of the issues you linked to saying:

> Seems like you are able to increase the length of foreign key, by default it is 64.

Just change your mysql configuration to allow longer foreign keys.

cristian-sima:
@danhardman Hi mate and thanks for reply.

[As the official documentation states](http://dev.mysql.com/doc/refman/5.5/en/identifiers.html), it **is impossible to modify the current length of 64 characters**.

Now, about the confusing part. It is not confusing as long as you already see which elements are involved in the relation. The identifier is redundant information composed of the table, start and destination keys. In the official documentation, the identifier it is also optional. [This discussion from StackOverflow ](http://stackoverflow.com/a/28620447) supports also the fact it is optional.

Therefore, @jinzhu I suggest using the SHA1 hash is the best solution (I have tried on my server and it works).

danhardman:
@cristian-sima I'm all for it then. +1

jinzhu:
I don't want to lose the readability by using `sha1`, lets think about if we could have a solution to keep the readability and make it works.

cristian-sima:
@jinzhu Can we try to find an answer for this problem, because I am using my version of this and therefore I can not be up-to-date with the last changes. Also, I need a solution when the identifier has more than 64 chars

I am thinking the best answer could be in this way:

- if the identifier is less then 64 chars it is not hashed 
- otherwise it is hashed using sha1

What do you think?

nkovacs:
Why not allow the developer to specify the name? E.g. `db.Model(a).AddForeignKeyWithName("name", "table", "refs(id)", "CASCADE", "CASCADE")`

cristian-sima:
@jinzhu What do you think about AddForeignKeyWithName ?

PR #896, opened 2016-03-09
Fix gorm.Open dialect

deliseev:
Fix gorm.Open dialect from 'sqlite' to 'sqlite3'


jinzhu:
Thank you for your PR

PR #947, opened 2016-04-04
Added AddLongForeignKey

cristian-sima:
AddLongForeignKey can be used when the identifier of foreign key is
exceeds 64 characters
It uses sha1 to shorten it

### DIFF
--- a/main.go
+++ b/main.go
@@ -567,6 +567,15 @@ func (s *DB) AddForeignKey(field string, dest string, onDelete string, onUpdate 
 	return scope.db
 }
 
+// AddLongForeignKey can be used when the identifier of foreign key is exceeds 64 characters
+// It uses sha1 to shorten it
+//     db.Model(&User{}).AddLongForeignKey("city_id", "cities(id)", "RESTRICT", "RESTRICT")
+func (s *DB) AddLongForeignKey(field string, dest string, onDelete string, onUpdate string) *DB {
+	scope := s.clone().NewScope(s.Value)
+	scope.addLongForeignKey(field, dest, onDelete, onUpdate)
+	return scope.db
+}
+
 // Association start `Association Mode` to handler relations things easir in that mode, refer: https://jinzhu.github.io/gorm/associations.html#association-mode
 func (s *DB) Association(column string) *Association {
 	var err error
--- a/scope.go
+++ b/scope.go
@@ -1,6 +1,7 @@
 package gorm
 
 import (
+	"crypto/sha1"
 	"database/sql"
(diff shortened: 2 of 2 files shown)


cristian-sima:
Referring to https://github.com/jinzhu/gorm/pull/882

I think the best way to do it is to have 2 methods as @nkovacs said [here](https://github.com/jinzhu/gorm/pull/882#issuecomment-199801291):

1. We can still use AddForeignKey
2. I created a new func *AddLongForeignKey* which uses sha1 to shorten the long identifier

In this way, the current code does not need to be changed and it allows some developers to use this new version for long identifiers.

@jinzhu What do you think?

cristian-sima:
@jinzhu Can you give any feedback here?

benguild:
@jinzhu Could you please merge/approve this? I'm running into this issue as well. It's completely legitimate.

nkovacs:
I disagree. I think my suggestion is better, because you get more readable foreign keys instead of sha1.

benguild:
@nkovacs I'm in favor of giving the option to the user, but honestly I've never really had an issue with needing to look at the names for my foreign keys. I can just look at the columns.

Can you explain why that's helpful?

Either way, this is subtly broken in in the current release and should definitely be addressed.... Keys longer than 64 characters aren't allowed!

nkovacs:
My bigger issue is that sha1 should be automatic for mysql. A developer working with postgres won't use this thing, because it's a mysql specific hack/workaround. My suggestion wouldn't be, it would allow you to customize the key names even if you don't have the 64 character problem. But both can be implemented, I just don't think exposing the sha1 workaround as a separate public method is a good idea.

benguild:
What Propel ORM does is add a few hash characters to the end of the auto-generated string:
`relation_names_and_stuff_a1b2cd3f`

Another issue is joining the same table on two separate keys/columns. I don't know if this is an issue, but in order to address conflicts... a user defined name might be necessary unless the hash factors in the key name, etc.

nkovacs:
the hash factor would be the full key name, which would be unique.  prefixing it with a shortened user readable string that contains as much important info as possible (the most relevant are foreign table and key) is a good idea.

benguild:
OK, so I assume this PR is going to be rejected in favor of something better? Are you interested in implementing something like this? I can try to help, but I'm in transit this weekend and trying to finish up like a million other things....... >_<
Just squashed this though for the time being: https://github.com/jinzhu/gorm/issues/924

nkovacs:
Here it is: #1029

PR #959, opened 2016-04-11
Fix postgres' HasForeignKey

vibhavp:


jinzhu:
Thank you for your PR.

PR #989, opened 2016-05-03
added error types for easier error handeling with cases

cbelsole:
After reading [this article](http://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) I wanted to handle gorm errors by type assertion instead of checking the message. 

Instead of:
```go
err := something()
switch err.Error() {
case "record not found":
  // do something
case "invalid SQL":
  // do something else
default:
  // do something default
}
```

I want to do:
```go
err := something()
switch err.(type) {
case ErrRecordNotFound
  // do something
case ErrInvalidSQL
  // do something else
default:
  // do something default
}
```

This has a few distinct advantages:
* We are not relying on the internals of the object to handle the errors.
* Types are immutable and we don't have to worry about them getting overridden.

### DIFF
--- a/callback_query.go
+++ b/callback_query.go
@@ -79,7 +79,7 @@ func queryCallback(scope *Scope) {
 			}
 
 			if scope.db.RowsAffected == 0 && !isSlice {
-				scope.Err(ErrRecordNotFound)
+				scope.Err(NewErrRecordNotFound())
 			}
 		}
 	}
--- a/errors.go
+++ b/errors.go
@@ -5,19 +5,6 @@ import (
 	"strings"
 )
 
-var (
-	// ErrRecordNotFound record not found error, happens when haven't find any matched data when looking up with a struct
-	ErrRecordNotFound = errors.New("record not found")
-	// ErrInvalidSQL invalid SQL error, happens when you passed invalid SQL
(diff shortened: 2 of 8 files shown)


jinzhu:
GORM's Error possible contains more than one error, refer: 

https://github.com/jinzhu/gorm/blob/021d7b33143de37b743d1cf660974e9c8d3f80ea/main.go#L634
https://github.com/jinzhu/gorm/blob/021d7b33143de37b743d1cf660974e9c8d3f80ea/main.go#L458

So this is not going to work.

SenZhangAI:
So， What‘s ’the better way to handler Errors with record not found and other Errors？ Is there any example？ :)

SenZhangAI:
I want to return "Record not found" when err == gorm.RecordNotFound, and return "Internal Error" in other err case.

but if I just use `RecordNotFound` like below，it will ignore other Errors when RecordNotFound and other Errors happened in the same time?  So, I wan't to find a better way to handler RecordNotFound and other Errors cases.

``` go
if db.Where("user_id = ?", id).First(&user).RecordNotFound() {
	constant.ErrMsg(c, constant.UserNotFound)
	return
}
```

PR #1000, opened 2016-05-06
Fixed SQL injections with ORDER BY and GROUP BY.

RichardKnop:
This PR is a quick way to solve two found SQL injections.

More tests should be written to test for other possible injections ASAP.

This was discovered by @eazynow. Thank you for spotting this!

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -723,15 +723,22 @@ func (scope *Scope) selectSQL() string {
 	return scope.buildSelectQuery(scope.Search.selects)
 }
 
+// only match string like `name`, `users.name`, `name ASC`, `users.name desc`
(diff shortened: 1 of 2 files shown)


RichardKnop:
@jinzhu

RichardKnop:
Alright this doesn't work. Looking at other ways to fix this.

RichardKnop:
@jinzhu Fixed tests. Now passing.

RichardKnop:
Discovered by @eazynow

jinzhu:
Hi @RichardKnop 

Thank you for your PR. but `Order`, `Group` can't do it like this, as it is supposed to be possible using with sql expression.

I think for your application, it is better to doesn't accept user's input as argument for those two methods, when use those inputs, you have to verify it.

PR #1001, opened 2016-05-06
association: dont execute Append/Replace/Delete if Error already present

stroborobo:
Prevents panics if associations are used as described in the docs but an error happened beforehand. (Like missing ID or something since the model wasn't saved yet.)

Not in `(a *Association) Count() int` since, well, int. Ideas? Why does this work differently then the other `Count()`s anyway? (Returning an int instead of writing to a pointer)


jinzhu:
Thank you for your PR.

PR #1003, opened 2016-05-06
Only set autoincrement if the tag is used. Fixes #1002

### LINKED ISSUE #1002: Do not Force Auto Increment
If a field is a primary key and a int, gorm forces the field to be a auto_increment field. This breaks down when you have multiple primary keys, or simply the int field is not suppose to be auto_incrementing.

Since gorm has support for the AUTO_INCREMENT tag, it is not needed to be forced.

Shelnutt2:
This change removes the forcing of the auto increment field for sqllite, postgres and mysql when using an int and a primary key. Instead the AUTO_INCREMENT tag can be used. I've also updated the default Model in order to maintain compatibility with expected and existing behavior.

### DIFF
--- a/customize_column_test.go
+++ b/customize_column_test.go
@@ -8,7 +8,7 @@ import (
 )
 
 type CustomizeColumn struct {
-	ID   int64     `gorm:"column:mapped_id; primary_key:yes"`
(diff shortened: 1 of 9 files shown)


brysearl:
👍

PR #1012, opened 2016-05-11
Add Oracle support

andreynering:
Observations:

- [ ] I think we can't get it working right with sequences. I think we also can't return the inserted ID in the same query
- [ ] This warning is shown on creating an index: `no RowsAffected available after DDL statement`
- [x] ~~There's no `LIMIT 1` suffix for Oracle, we have to use `WHERE ROWNUM <= 1`, so the current dialect interface won't work for this.~~ I fixed this but only `LIMIT` work (via `ROWNUM`), `OFFSET` is ignored for compatibility with Oracle 11c

### DIFF
--- a/main.go
+++ b/main.go
@@ -561,7 +561,15 @@ func (s *DB) RemoveIndex(indexName string) *DB {
 
 // AddForeignKey Add foreign key to the given scope, e.g:
 //     db.Model(&User{}).AddForeignKey("city_id", "cities(id)", "RESTRICT", "RESTRICT")
-func (s *DB) AddForeignKey(field string, dest string, onDelete string, onUpdate string) *DB {
+func (s *DB) AddForeignKey(field string, dest string, options ...string) *DB {
+	var onDelete string
+	var onUpdate string
+	if len(options) >= 1 {
+		onDelete = options[0]
+	}
+	if len(options) >= 2 {
+		onUpdate = options[1]
+	}
 	scope := s.clone().NewScope(s.Value)
 	scope.addForeignKey(field, dest, onDelete, onUpdate)
 	return scope.db
--- a/scope.go
+++ b/scope.go
@@ -329,8 +329,12 @@ func (scope *Scope) QuotedTableName() (name string) {
 
 // CombinedConditionSql return combined condition sql
 func (scope *Scope) CombinedConditionSql() string {
+	limitWhereSQL, limitSuffixSQL := scope.limitAndOffsetSQL()
+	if len(limitWhereSQL) > 0 {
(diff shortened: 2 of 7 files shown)


jinzhu:
Hi @andreynering 

Could it works like this? https://github.com/jinzhu/gorm/blob/master/dialects/mssql/mssql.go#L129

andreynering:
@jinzhu Hi and thanks for helping.

I searched and that syntax should work on version 12c, but not on version 11c (the one a have here, which is free with limitations).

In order to make it compatible with both, we would need to modify the `WHERE` statement, and use `ROWNUM`.

main.go
```diff
@@ -561,7 +561,15 @@ func (s *DB) RemoveIndex(indexName string) *DB {
 
 // AddForeignKey Add foreign key to the given scope, e.g:
 //     db.Model(&User{}).AddForeignKey("city_id", "cities(id)", "RESTRICT", "RESTRICT")
-func (s *DB) AddForeignKey(field string, dest string, onDelete string, onUpdate string) *DB {
+func (s *DB) AddForeignKey(field string, dest string, options ...string) *DB {
```
andreynering: This change is to make these options optional.

scope.go
```diff
@@ -469,7 +473,7 @@ func (scope *Scope) scan(rows *sql.Rows, columns []string, fields []*Field) {
 		}
 
 		for fieldIndex, field := range selectFields {
-			if field.DBName == column {
+			if strings.EqualFold(field.DBName, column) {
```
andreynering: Case insensitive comparison: This is necessary because Oracle returns the column names in UPPER CASE.

andreynering:
@jinzhu Fixed that.

It is working with some limitations (no sequences, no offset, maybe something not tested). Maybe you can merge it and add a note about Oracle support being experimental?

(You might want to skip `go get` of `ora` to fix CI).

pedromorgan:
I Regejct hsi patch 100000%..

We dont want any Poropreiery interface taking the whole projectl..

SO this had to be a "submodule" 2000% and a take down notice from Oracles lawyers will not demolish whole project..


SORRY FOLKS>>>

pedromorgan:
NO NO NO..  lawyers are on ...

- Can we sue them ??
- Does it run on Android ??
- Is it GPL .. a threat to society and communist!!
- Need I go on..

We dont want anything mentioning Oracle in thsi project, seperate please..

pedromorgan:
examples..
mysql.. openoffice.. huge lawsuit against google for billons
For ref
http://www.bonkersworld.net/images/2011.06.27_organizational_charts.png

And where is groklaw when u need a ```pamalegal``` ;-(((

PR #1026, opened 2016-05-21
Fixing BOOL type to UNSIGNED in MySQL instances.

benguild:


jinzhu:
Hi @benguild 

Thank you for your PR, but I would like to keep it like this, Thank you anyway.

benguild:
For reference there is a whole discussion about this here:
http://stackoverflow.com/a/289759/963901

It might be better to use BIT(1), but I've never done that because apparently you need to cast it to being a BOOL which is more work.

PR #1029, opened 2016-05-21
Mysql foreign key length fix

nkovacs:
This is based on #947, but it doesn't add a new method, it automatically hashes the key name if it would be too long on mysql.
I had to add a new method to Dialect to make the foreign key name different for mysql.

### DIFF
--- a/association_test.go
+++ b/association_test.go
@@ -2,6 +2,7 @@ package gorm_test
 
 import (
 	"fmt"
+	"os"
 	"reflect"
 	"sort"
 	"testing"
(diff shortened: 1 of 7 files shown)


jinzhu:
That looks much better, thank you for the PR.

benguild:
Thanks!!

nkovacs:
There's a panic if dest is shorter than 24 bytes, I'll fix it soon (as well as the string length calculation, because it should count utf8 characters, not bytes).

cristian-sima:
Happy #947 is done :100: Tx @nkovacs for supporting this

PR #1038, opened 2016-05-23
Fixed an occurrence of a panic while preloading non existent fields.

RichardKnop:
I have encountered a possible panic while preloading non existent fields (e.g. a foreign key can be NULL).

Check the unit tests. Uncomment my bugfix in `callback_query_preload.go` and run the tests to verify the bug. You will get a panic because `currentScope.getColumnAsScope` can possibly return nil pointer and you don not check for that case currently.

### DIFF
--- a/callback_query_preload.go
+++ b/callback_query_preload.go
@@ -28,6 +28,10 @@ func preloadCallback(scope *Scope) {
 		for idx, preloadField := range preloadFields {
 			var currentPreloadConditions []interface{}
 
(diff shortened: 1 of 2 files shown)


RichardKnop:
@jinzhu Can you look into the wercker failure? It fails to get the code form github for some reason.

I ran all tests with `./test_all.sh` and they all pass so this should be fine.

jinzhu:
Thank you for your PR

PR #1052, opened 2016-06-01
Add support for polymorphic preload

### LINKED ISSUE #1051: Polymorphic Preload doesn't respect OwnerType
Hey jinzhu :)

Somehow Preload() doesn't respect the `*Type` field when the association is polymorphic, here's a gist to show what I mean: [gist](https://gist.github.com/74650f6fa9ffabf1f4c71f1d6364ee83) (You need to uncomment the create()-call on first run.)

Preload results in:

```sql
SELECT * FROM "countdowns"  WHERE ("owner_id" IN ('1')) ORDER BY "countdowns"."id" ASC
```

Adding `.Model()` doesn't change that unfortunately.

stroborobo:
Fixes #1051


jinzhu:
Thank you for your PR.

PR #1075, opened 2016-06-20
Fix autoincrement

cdevienne:
In the current implementation, AUTO_INCREMENT on simple attributes does not work on postgres.
This PR fixes that.


cdevienne:
Just had a look to sqlite doc, it seems auto increment is supported only on primary key columns, so gorm already supports it as mush as possible.

PR #1104, opened 2016-07-05
concurrent preloads

jlburkhead:
turning https://github.com/jinzhu/gorm/issues/1083 into a pr

### DIFF
--- a/callback_query_preload.go
+++ b/callback_query_preload.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync"
 )
 
 // preloadCallback used to preload associations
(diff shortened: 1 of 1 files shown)


jinzhu:
Seems CI failed.

jlburkhead:
@jinzhu If you're open to adding this I can spend time to figure out why the tests are failing. Seems like it will take some time to figure out since each test passes individually, but running all of them fails certain checks

jinzhu:
Hi @jlburkhead 

Yes, good to have that if it does improved the performance.

Fyb3roptik:
Is this something that would fix #1436 ????

zardak:
@Fyb3roptik Its about running multiple preload queries in parallel rathen than sequentially. This has nothing to do with tweaking the preload strategy.

canhlinh:
I will check this ticket.

canhlinh:
Hmm. The code can not work in concurrent.
```
In each goroutine, the result update result to scope. It causes the error. 
Need to refactor the code to implement concurrent.
```

lunemec:
Please also be very careful when running inside transaction. That will mean all of those goroutines will share 1 TCP connection to the DB (something to consider and test for).

jinzhu:
No feedback for a long while, going to close it for now.

PR #1108, opened 2016-07-09
Fixed bug when preload duplicates has many related objects.

RichardKnop:
This PR fixes a bug with preloads of `has many` related objects would be duplicated many times when using `Find` method. Consider this example:

```
	type (
		Postcode struct {
			ID     uint
			AreaID uint
		}
		Area struct {
			ID         uint
			Postcodes []*Postcode
		}
		User struct {
			ID       uint
		}
		UserArea struct {
			ID     uint
			UserID uint
			User   *User
			AreaID uint
			Area   *Area
		}
	)
```

In the structure above, we have multiple areas in our database such as `London`, `Paris`. Each area has a list of postcodes assigned to it, e.g. `EC1`, `EC2`.

We can then create a user area objects related to specific users. For example, a db structure like above could be used to store data about users wanting to buy / sell properties in different areas.

Now the bug is, if we create an area for `London` and assign two postcodes to it - `EC1`, `EC2` and since each user is looking for a house in the same area, we would create a user area object for both users.

When we fetch the users from the db and use `Preload("Area.Postcodes")`, each object would contain duplicate entries for postcodes, e.g. both users would have `["EC1", "EC2", "EC1", "EC2"]` even though that's clearly incorrect.

See the unit test I wrote to prove this bug.

You can run tests without the change in `callback_query_preload.go` to see it fail.

The change I made in `callback_query_preload.go` fixes this bug.

Please review and merge if happy :)

Also if you have a more elegant way to fix this bug, do it other way. I couldn't figure out a cleaner way to correct this behaviour.

@jinzhu please have a look

### DIFF
--- a/callback_query_preload.go
+++ b/callback_query_preload.go
@@ -186,6 +186,13 @@ func (scope *Scope) handleHasManyPreload(field *Field, conditions []interface{})
 		for j := 0; j < indirectScopeValue.Len(); j++ {
 			object := indirect(indirectScopeValue.Index(j))
 			objectRealValue := getValueFromFields(object, relation.AssociationForeignFieldNames)
+			if j > 0 {
+				prevObject := indirect(indirectScopeValue.Index(j - 1))
+				prevObjectRealValue := getValueFromFields(prevObject, relation.AssociationForeignFieldNames)
+				if toString(prevObjectRealValue) == toString(objectRealValue) {
+					continue
+				}
+			}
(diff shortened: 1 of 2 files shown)


RichardKnop:
@jinzhu All tests pass locally with `./test_all.sh` so not sure what's wrong with the CI.

jinzhu:
Hi @RichardKnop 

Thank you for your PR, just fixed this in another way.

Thank you.

PR #1113, opened 2016-07-14
update URLs to new location

baijum:


jinzhu:
I prefer keep it like now, and rely github handle redirects

Thank you for your PR anyway.

PR #1120, opened 2016-07-18
[WIP] Support count distinct

huydx:
Currently count distinct is not supported. Of course we can use `Raw` method, but I think Count distinct is common enough to support as a separate function (with `Count`)

This current PR is to demonstrate an idea first, so I've not yet added test cases. After we get to how to support it properly, i will add test later.

Cheer.

### DIFF
--- a/main.go
+++ b/main.go
@@ -298,6 +298,11 @@ func (s *DB) Count(value interface{}) *DB {
 	return s.NewScope(s.Value).count(value).db
 }
 
+// CountDistinct get how many distinct records for a model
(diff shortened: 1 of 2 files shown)


huytd:
I like the idea to put `distinct` to gorm. But it would be better to have a separate `Distinct()` function, so we have something like:

```
DB.Model(&user).Distinct().Count()
```

I believe we're not just using distinct for counting only.

jinzhu:
Hi All,

You could write it like below now:

`db.Model(&Comment{}).Select("count(distinct id)").Count(&count)`

Thank you for your PR.

PR #1123, opened 2016-07-19
Add prefix support

GhostRussia:


danhardman:
Table names shouldn't have prefixes, it's bad practice.

Issue #1054, opened 2016-06-02
"too many SQL variables" error when using Preload


jchavannes:
When using `Preload`, e.g.:

```
db.Preload("Teams").Find(&users)
```

I'm getting the following error:

```
too many SQL variables
```

The main issue is that there's no way (or at least I'm not aware of a way) to split into multiple queries based on the limit of variables per query and sqlite3 doesn't have any way to increase the maximum.

Additionally, when outputting queries, I saw that it was not de-duping the preload items.  For example, using just a couple users/teams w/ preload it'll use this query:

```
SELECT * FROM "teams"  WHERE ("id" IN ('1','1','1','1','1','1','1','2','2','2'))
```

You can see it's trying to find the same team multiple times.  I assume this is contributing to the issue.

ernestas-poskus:
> sqlite3 doesn't have any way to increase the maximum

maybe use MySQL / PostgreSQL ?

jchavannes:
> maybe use MySQL / PostgreSQL ?

Yeah, that's a potential workaround.  The workaround I'm doing is to just load manually instead of using `Preload`.

Figured it was worth filing an issue though since it would be easier/convenient to use `Preload` and it looked like there was some room for optimization :)

jinzhu:
We have did some work to remove some duplicated values in the where, hope could avoid this issue.

jonrzhang:
Hi, @jinzhu Thanks for your great work.

I didn't find any solution if I want to use Preload to distinct the where assumption. It still had duplicated values in the where.
I would ask if there was any way to avoid the problem during using Preload()?
I still met the __"too many SQL variables"__ issue.

paolobarbolini:
Hi, I just encountered this problem myself.

From what I'm reading from previous comments this doesn't seem to happen with MySQL and PostgreSQL, so being this an ORM library I think the issue should be addressed in the library, while manually preloading as @jchavannes suggested should only have been a temporary workaround while a proper fix gets implemented.

PR #1132, opened 2016-07-29
Omit duplicates and zero-value ids in preload queries. Resolves #854 and #1054.

### LINKED ISSUE #854: escape "<Nil>" when preload
Hi, when i preload a query such as 

``` golang
DB.Where("user_id = ?", userID).preload("Avatar").First(&user)
```

If user has no Avatar, gorm will still generate a query like follow:

``` SQL
select * from photo where (id = "<nil>");
```

is there any way to prevent such redundant action? Thanks

zardak:
Seems like this can be applied to every `getColumnAsArray` usage, still was unsure whether to rewrite it, instead of creating new method.

Looking forward to your comments.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -1204,10 +1204,7 @@ func (scope *Scope) autoIndex() *Scope {
 
 func (scope *Scope) getColumnAsArray(columns []string, values ...interface{}) (results [][]interface{}) {
 	for _, value := range values {
-		indirectValue := reflect.ValueOf(value)
-		for indirectValue.Kind() == reflect.Ptr {
-			indirectValue = indirectValue.Elem()
-		}
+		indirectValue := indirect(reflect.ValueOf(value))
 
 		switch indirectValue.Kind() {
 		case reflect.Slice:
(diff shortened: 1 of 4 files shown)


mcastilho:
It would be nice to see this merged in... not having the dedupe of ids causes lots of problems for us too..

scope.go
```diff
@@ -1230,6 +1227,40 @@ func (scope *Scope) getColumnAsArray(columns []string, values ...interface{}) (r
 	return
 }
 
+func (scope *Scope) getColumnAsArrayUnique(columns []string, values ...interface{}) (results [][]interface{}) {
+	unfilteredResults := scope.getColumnAsArray(columns, values...)
+
+	rootMap := map[interface{}]interface{}{}
```
danhardman: How come this is declared here instead of simply declaring `currentMap := map[interface{}]interface{}{}`?

zardak: If I understand you correctly, the main point is in `O(log(N))` lookups in case of multiple primary keys. So its basically a map of map of empty structs (which occupy zero memory on contrary to bools)  when primary key consists of two columns. Keys of first map are unique values for first column and keys of second-level maps are all unique values for second column which exist together with first column value, and so on.

jinzhu:
Merged that, thank you for your PR. https://github.com/jinzhu/gorm/commit/066abcef408f2522f0a89952e44ab17eac176ed0

Issue #1138, opened 2016-08-06
Create may set incorrect primary key [postgres]


jw-00000:
When multiple INSERTs of new records (without primary key) are executed simultaneously on a table, a race condition can occur, causing the record's primary key to be set to the incorrect value.

A test case and its output are in [this gist](https://gist.github.com/jw-00000/5396cae6ebf838d04639e6f95ebac91b). It consists of two files, `main.go` contains the test case and `output.txt` one output I got. (You may need to execute the script multiple times to trigger the error.)

What happens with one thread is the following:
1. We create a Test record, without ID.
2. We call db.Create on this record.
3. This executes `INSERT ... RETURNING "tests"."id"` ([callback_create.go line 120](https://github.com/jinzhu/gorm/blob/master/callback_create.go#L120)). (I think here the record's ID is set, correctly, using Scan.)
4. Next, [callback_create.go line 129](https://github.com/jinzhu/gorm/blob/master/callback_create.go#L129) executes `SELECT "id" FROM "tests"`. (This is not the intended query!) This sets the record's ID, possibly to an incorrect value.

The test case triggers the error using multiple threads. If two threads A and B execute at the same time, it may happen that A executes step 3, followed by B executing step 3, and then A executing step 4. In this case, thread A will get the ID of the record inserted by B, and uses that value to set the primary key of its record, incorrectly.

In the attached test case's output, you can see how the INSERT and SELECT statements from steps 3 and 4 are interleaved. On [line 141](https://gist.github.com/jw-00000/5396cae6ebf838d04639e6f95ebac91b#file-output-txt-L141) you can see that the ID of record 2 was incorrectly set to 4.

If I understand everything correctly, the error is in the function [forceReloadAfterCreateCallback of callback_create.go](https://github.com/jinzhu/gorm/blob/master/callback_create.go#L128). The for loop in this function is supposed to add a WHERE clause to its query, but does not do so in this test case. This happens because it is the primary key that is blank (the if never matches). This results in `SELECT "id" FROM "tests"`, which always returns the last inserted record, even though this may be a record of another thread.

(I suspect this may only happen on PostgreSQL, since this is the only dialect that uses the `RETURNING` mechanism, but I'm not sure about this.)

jw-00000:
I think this can be fixed by changing line 132 from
    if field.IsPrimaryKey && !field.IsBlank {
to
    if field.IsPrimaryKey {
but I'm not sure.

[...]

PR #1139, opened 2016-08-06
Fix bug #1138

jw-00000:
I think this may fix bug #1138.


jinzhu:
Hi @jw-00000 

Thank you for your report, I have fixed the issue in another way, Thank you.

PR #1146, opened 2016-08-09
Fixed a bug when joining multiple tables with the same fields

GhostRussia:
Fixed a bug when joining multiple tables with the same fields and where on the same field.


jinzhu:
Thank you for your PR.

PR #1151, opened 2016-08-11
Find: order by primary key by default

lscieux:
Hi there !

There is currently no default order for preloaded tables when Find() is used.
I have added gorm:order_by_primary_key to scope, the same way it's added in First().

### DIFF
--- a/main.go
+++ b/main.go
@@ -253,7 +253,8 @@ func (s *DB) Last(out interface{}, where ...interface{}) *DB {
 
 // Find find records that match given conditions
 func (s *DB) Find(out interface{}, where ...interface{}) *DB {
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @lscieux 

Thank you for your PR, but I think this is not necessary, if you want to add the condition by default, you could register a callback for that. 

Refer: http://jinzhu.me/gorm/development.html#write-plugins

lscieux:
Thanks for your response. Yes callback could be a good alternative.
I just thought it was a bit strange that by default First and Find do not use the same order.

jinzhu:
Not sure, but I think its better to use database's default order when call with `Find`

PR #1154, opened 2016-08-12
Add primary key check before delete

danhardman:
Added primary key validation before deleting a record to ensure all records aren't deleted if a primary key is not set when calling `db.Delete(value)`. That behaviour should be reserved for `db.DropTable()` or have it's own function.

Unfortunately I haven't been able to add a test as running `go test` returned error:

```
./query_test.go:311: cannot use gorm.Expr("name = ? DESC", "OrderPluckUser2") (type *gorm.expr) as type string in argument to scopedb.Order
FAIL	github.com/danhardman/gorm [build failed]
```

Shouldn't have to be fiddling around with other code to get this working!

### DIFF
--- a/callback_delete.go
+++ b/callback_delete.go
@@ -20,6 +20,10 @@ func beforeDeleteCallback(scope *Scope) {
 
 // deleteCallback used to delete data from database or set deleted_at to current time (when using with soft delete)
 func deleteCallback(scope *Scope) {
+	if scope.PrimaryKeyZero() {
+		scope.db.AddError(ErrNonSpecificDelete)
+	}
+
 	if !scope.HasError() {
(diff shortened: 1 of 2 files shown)


jinzhu:
Hi @danhardman 

Sorry, but I don't think this is necessary. I think before delete, you should check the data is valid or not.

Also `Delete` could also works with structs w/o primary key but other fields as conditions, like: `db.Delete(&User{Role: "normal"})`

jinzhu:
Thank you for your PR anyway.

danhardman:
Hadn't even thought about that, however it would be possible to check if every field on a given value has a "zero-value". Although it would be temperamental I guess. Shame!

PR #1181, opened 2016-09-12
Add DialectName to database handle

tboerger:


jinzhu:
Hi @tboerger 

Thank you for your PR

I am thinking it might be more useful to expose the dialect to public, so did it in this commit https://github.com/jinzhu/gorm/commit/6732a50dfbf1777a5915b7e04af006badacea3be , thank you for your idea.

PR #1186, opened 2016-09-14
skip preloading nil fields

jlburkhead:
Related to https://github.com/jinzhu/gorm/issues/854

I noticed the same issue happening when there were nested preloads and the first level didn't exist ie `db.Preload("User.Photo").Find ...` when the related User isn't found a query is still made for the photo (something like `SELECT * FROM photos WHERE id IN ('<nil>')`

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -1217,19 +1217,48 @@ func (scope *Scope) getColumnAsArray(columns []string, values ...interface{}) (r
 				var result []interface{}
 				var object = indirect(indirectValue.Index(i))
 				for _, column := range columns {
-					result = append(result, object.FieldByName(column).Interface())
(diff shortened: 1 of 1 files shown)


jlburkhead:
None of the tests fail, but I don't know enough about gorm to know if this might cause any issues. Also it will only work if a field is a pointer or one of the `database/sql.Null...` types.

jlburkhead:
somehow i missed it. but https://github.com/jinzhu/gorm/pull/1132 looks like a better version of this

jinzhu:
Fixed that. Thank you for your PR https://github.com/jinzhu/gorm/commit/066abcef408f2522f0a89952e44ab17eac176ed0

Issue #758, opened 2015-12-13
Is it possible to make an empty slice for an association that has no data instead of nil?


Gacnt:
I'm currently working on a project and right now I have some basic javascript

`blah.map(function(lBlah) {});`

`blah` is generated from an ajax request to a database which returns an array of objects (or slice if you prefer)

The only problem is I get a JS error: `cannot read property 'map' of null` this is because if my `one-to-many` association has no associated values at the time, it just makes that field `nil` instead of an empty slice so when I `json.Marshal` it it sends a null value instead of an empty `[]` which would fix all my errors, currently having to range over each `struct` that is returned from the database and check for nil values then `make([]blah, 0)` is annoying and looks messy. Is there an easier way to accomplish this? Is it possible to set a default `json:"default:[]"` or something? Any help would be great thanks.

jugglinmike:
gh-1199 addresses this.

PR #1199, opened 2016-09-27
Return empty slice for associations with 0 results

jugglinmike:
When using `Preload` to include the results of a "has many"
relationship, Gorm previously returned an uninitialized slice for any
such relations that bore zero records. This distinction was most
apparent when the results were marshalled to a JSON representation--a
record with zero related records would be represented with `null`.

For example, consider the following schema:

    id | name
    ---|------
    1  | Lorin
    2  | Sue

    id | p_id | value
    ---|------|-------------------
    1  | 1    | lorin@example.com
    2  | 1    | lorin2@example.com

Querying with:

    db.Preload("Email").Find(&people)

And marshalling the resulting value of `people` to JSON would yield the
following string:

    [
      {
        "name": "Lorin",
        "email": [
          "lorin@example.com",
          "lorin2@example.com"
        ]
      },
      {
        "name": "Sue",
        "email": null
      }
    ]

Beyond being inconsistent, the value `null` in this response differs
semantically from the actual state of the database. The database
actually has zero related records for the second user, so a JSON value
of `[]` is appropriate.

Update the callback that processes "has many" relationships to
communicate empty query results with an empty slice.

### DIFF
--- a/callback_query_preload.go
+++ b/callback_query_preload.go
@@ -186,9 +186,11 @@ func (scope *Scope) handleHasManyPreload(field *Field, conditions []interface{})
 		for j := 0; j < indirectScopeValue.Len(); j++ {
 			object := indirect(indirectScopeValue.Index(j))
 			objectRealValue := getValueFromFields(object, relation.AssociationForeignFieldNames)
+			f := object.FieldByName(field.Name)
 			if results, ok := preloadMap[toString(objectRealValue)]; ok {
-				f := object.FieldByName(field.Name)
 				f.Set(reflect.Append(f, results...))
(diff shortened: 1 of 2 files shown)


jugglinmike:
This behavior was previously reported in gh-758, although that issue was closed by the reporter without further comment.

Ganitzsh:
Would it be possible to do the same for a "many to many" relationship?

henrylusjen:
This feature is not working on v2.

jinzhu:
@henrylusjen  this should been fixed in latest master

PR #1202, opened 2016-09-28
Add named value for PolymorphicType (to optionally replace scope.TableName)

slockij:
Use case 1:
Hamster has 2 type of toys, preferred and the other.
Use case 2:
School has hall photo and set of individual photos for classrooms.

To save both types of sub-object into one table you need to set relation on something else than table name. 
Solution - add "value" into struct tag:

```
type Hamster struct {
	Id           int
	Name         string
	PreferredToy Toy `gorm:"polymorphic:Owner;value:hamster_preferred"`
	OtherToy     Toy `gorm:"polymorphic:Owner;value:hamster_other"`
}
```

### DIFF
--- a/association.go
+++ b/association.go
@@ -63,7 +63,11 @@ func (association *Association) Replace(values ...interface{}) *Association {
 	} else {
 		// Polymorphic Relations
 		if relationship.PolymorphicDBName != "" {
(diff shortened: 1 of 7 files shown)


jinzhu:
Thank you for your PR, did some changes, need to rename the tag name to `polymorphic_value`

PR #1203, opened 2016-09-30
Include 'LastInsertId' for exec queries.

sepehrmm:
Same way that `RowsAffected` is incorporated into the result include `LastInsertId` too.

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,6 +14,7 @@ type DB struct {
 	Value             interface{}
 	Error             error
 	RowsAffected      int64
+	LastInsertId      int64
 	callbacks         *Callback
 	db                sqlCommon
 	parent            *DB
--- a/scope.go
+++ b/scope.go
@@ -348,6 +348,9 @@ func (scope *Scope) Exec() *Scope {
 			if count, err := result.RowsAffected(); scope.Err(err) == nil {
 				scope.db.RowsAffected = count
 			}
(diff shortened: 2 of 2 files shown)


jinzhu:
You can get the ID from the saved struct?

sepehrmm:
Hi jinzhu,
With this changes I can get the last insert id like this:
`dbResult := transaction.Exec(`INSERT INTO test (name) VALUES (?)`, name)
fmt.Println("dbResult", dbResult.LastInsertId)`

EagleChen:
will this pr be merged? @jinzhu

EagleChen:
@jinzhu, sorry to bother you again. It seems reasonable to include 'LastInsertId' for 'exec' (insert) queries

jinzhu:
If you want to get the LastInsertID with Raw SQL, I guess better to do

```go
result, err := db.DB().Exec("insert into test (name) values (?)", name)
```

apremalal:
@jinzhu your previous answer doesn't return LastInsertID

apremalal:
@jinzhu how can we get this when it's a transaction? I tried 
` db.DB().Exec("insert into test (name) values (?)", name)` but it threw an error,
`panic: interface conversion: gorm.sqlCommon is *sql.Tx, not *sql.DB`

PR #1210, opened 2016-10-07
If no arguments, return error at once

baijum:
This is the scenario:

1. Caller side calls the Open without argument
2. Caller side calls the Close afterwards when `err != nil` - [this will panic](https://github.com/almighty/almighty-core/pull/344#discussion_r82363818)

- [x] Write test case


jinzhu:
Thank you for your PR

PR #1216, opened 2016-10-10
Update associations.md

naokij:
Many to many association
should use
```
db.Model(&user).Related(&languages, "Languages") 
```
instead of
```
db.Model(&user).Related(&languages)
```
http://stackoverflow.com/questions/34667199/gorm-many-to-many-select-gives-invalid-association-error


jinzhu:
Thank you for your PR

PR #1235, opened 2016-10-20
Added FindInBatches, FindEach

ronna-s:
I inquired about this in the past and in the end I implemented something on my own. I now reimplemented it to follow the code style of the project (I hope)...

The purpose is to iterate over records of a very large table without using too much memory.

Usage (very similar to RoR) - full example below:

```go
db.FindEach(&user, func() {
	fmt.Println(user.Name)
})
```
will result in:
```sql
SELECT * FROM "users"  WHERE "users".deleted_at IS NULL LIMIT 1000
SELECT * FROM "users"  WHERE "users".deleted_at IS NULL AND ((`id` > '1000')) LIMIT 1000
SELECT * FROM "users"  WHERE "users".deleted_at IS NULL AND ((`id` > '2000')) LIMIT 1000
...
```
If limit was already set, it will be used instead of defaulting to 1000. So the following:
```go
db.Limit(3).FindEach(&user, func() {
	fmt.Println(user.Name)
})
```
Will generate:
```sql
SELECT * FROM "users"  WHERE "users".deleted_at IS NULL LIMIT 3
SELECT * FROM "users"  WHERE "users".deleted_at IS NULL AND ((`id` > '3')) LIMIT 3
SELECT * FROM "users"  WHERE "users".deleted_at IS NULL AND ((`id` > '6')) LIMIT 3
...
```

Note: I think there might be an issue with different dialects support with the explicit 
`Where(fmt.Sprintf("`%v` > '%v'", scope.PrimaryKey(), scope.PrimaryKeyValue()))` I added to forward the iteration.

Let me know what you think...

```go
package main

import (
	"github.com/jinzhu/gorm"
	_ "github.com/jinzhu/gorm/dialects/sqlite"
)

type Product struct {
	gorm.Model
	Price uint
}

func main() {
	var product Product
	var i uint

	db, err := gorm.Open("sqlite3", "test.db")
	if err != nil {
	}
	defer db.Close()
	db.LogMode(true)
	db.DropTable(&product)
	// Migrate the schema
	db.AutoMigrate(&Product{})
	for i = 0; i < 10000; i++ {
		db.Save(&Product{Price: i})
	}
	db.FindEach(&product, func() {
	})
}

```

### DIFF
--- a/main.go
+++ b/main.go
@@ -264,6 +264,58 @@ func (s *DB) Find(out interface{}, where ...interface{}) *DB {
 	return s.clone().NewScope(out).inlineCondition(where...).callCallbacks(s.parent.callbacks.queries).db
 }
 
+// FindInBatches find records in batches for large tables and allow callbacks to operate
+// on each batch
+func (s *DB) FindInBatches(out interface{}, cb func(), where ...interface{}) *DB {
+	var limit interface{}
+	newDB := s.clone()
+	if len(newDB.search.orders) != 0 {
+		//ordering by non primary col not supported - can cause infinite loop
+		newDB.search.orders = []interface{}{}
+	}
+	value := indirect(reflect.ValueOf(out))
+	if value.Kind() != reflect.Array && value.Kind() != reflect.Slice {
+		newDB.AddError(ErrUnsupportedType)
+		return newDB
+	}
+	if newDB.search.limit == -1 {
+		limit = 1000
+	} else {
+		limit = newDB.search.limit
+	}
+	err := newDB.Limit(limit).Find(out, where...).Error
+	records := indirect(reflect.ValueOf(out))
+	for err == nil && records.Len() > 0 {
+		cb()
+		lastRecord := records.Index(records.Len() - 1).Interface()
+		scope := s.NewScope(lastRecord)
+		err = newDB.Limit(limit).
+			Where(fmt.Sprintf("`%v` > '%v'", scope.PrimaryKey(), scope.PrimaryKeyValue())).
+			Find(out, where...).Error
+		records = indirect(reflect.ValueOf(out))
+	}
+	if err != nil {
+		newDB.AddError(err)
+	}
+	return newDB
+}
+
+// FindEach find records in batches for large tables and allow callbacks to operate
+// on each of them between iterations
+func (s *DB) FindEach(out interface{}, cb func(), where ...interface{}) *DB {
+	c := s.clone()
+	v := indirect(reflect.ValueOf(&out)).Interface()
+	slice := makeSlice(reflect.TypeOf(v))
+	c = c.FindInBatches(slice, func() {
+		valueOfSlice := indirect(reflect.ValueOf(slice))
+		for i := 0; i < valueOfSlice.Len(); i++ {
+			reflect.ValueOf(out).Elem().Set(indirect(valueOfSlice.Index(i)))
+			cb()
+		}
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -264,6 +264,58 @@ func (s *DB) Find(out interface{}, where ...interface{}) *DB {
...
+		limit = newDB.search.limit
+	}
+	err := newDB.Limit(limit).Find(out, where...).Error
+	records := indirect(reflect.ValueOf(out))
+	for err == nil && records.Len() > 0 {
+		cb()
+		lastRecord := records.Index(records.Len() - 1).Interface()
+		scope := s.NewScope(lastRecord)
+		err = newDB.Limit(limit).
+			Where(fmt.Sprintf("`%v` > '%v'", scope.PrimaryKey(), scope.PrimaryKeyValue())).
```
zardak: `.PrimaryKeyValue()` may return a non-integer value, as `string` primary keys are a common case.

ronna-s: tldr; Yup, I'm not sure how to handle this. Any suggestions?
I guess I can first check if the field type is `Comparable()`.
The important thing is that if a > b and b > c than c cannot be < a, otherwise we will end up with endless loop, assuming that the ordering is also consistent with the sign, it won't lose records.

zardak: The traditional way to do batching is `OFFSET` with `LIMIT`.
Still, I'm not sure whether this will get merged, because basic batching is pretty trivial to implement and it often needs to be tailored to specific use-cases.

ronna-s: Offset is a very slow way to iterate over a table,  especially a large one.
This is for instance how rails implemented it: http://apidock.com/rails/ActiveRecord/Batches/find_in_batches.
Can you explain why you think it won’t work with strings?

ronna-s: As I wrote before, assuming the ordering is by primary key (maybe just add it to the query explicitely) we won't lose any records. We don't need to iterate over the table by order of creation.

zardak: Yep, you'll either have to force the ordering, or fall back to `OFFSET` for non-integer PKs.

jinzhu:
Hi @ronna-s 

Thank you for your PR, but I would like to suggest you to use this http://jinzhu.me/gorm/advanced.html#scan-sqlrows-in-iteration

ronna-s:
May I ask why? 
As far as I can tell the advantages of this approach are: 
1. Iterating in bulks over large tables is bug prone and a tough operation enough to offer a generic solution to. 
2. Since go doesn't offer generics, if we take your approach it will require implementing a bulk finder per struct (or duplicate your internal makeSlice and indirect methods anyway). I assumed that was one of the problems gorm was attempting to solve.
3. The advantage in the RoR find_each solution is that when the table grows you don't need to rewrite everything, only replace your calls to `each` with `find_each`.
4. Bonus: You basically had everything in place for me to do this, especially makeSlice.

ronna-s:
@jinzhu ?

jinzhu:
I think the current solution is pretty generic already, and It won't load all data into structs once to reduce your application's cost.

ronna-s:
Hi @jinzhu , the big advantage of this approach is to just allow changing the function call `Find([]SomeData)` to FindEach([]SomeData, function()) - so not to refactor the code too drastically. This will help in promoting this package to user with existing large tables.

XuruiPro:
If the primary key is not queried, only the first time will be queried.I don't know how to solve it.

PR #1242, opened 2016-10-25
Make gorm.Errors available for use outside gorm

calebthompson:
gorm.Errors, which usefully implements `error` for an `[]error` as returned by
`DB.GetError()` was already exported, but because it used a private field
`errors`, it was not able to be created due to the compile-time error:

```
implicit assignment of unexported field 'errors' in gorm.Errors literal
```

The trivial solution would be to export the `errors` field on
`gorm.Errors`, but this led to the issue that the common pattern of checking
`err != nil` failed because a struct{error: nil} != nil.

We can take advantage of type aliasing here to make Errors an []error, which
can in fact be nil and would pass `err != nil` on the happy path.
- Remove `(Errors) GetErrors()`, as it's less useful when Errors is an
  []error which can be iterated over. While this is technically a
  breaking change, we never expose an Errors and its difficult to build
  one (it can be done with the existing `(Errors) Add(error)`), but
  awkwardly. This removal can be reverted without issue and we can make
  it an identity method, but it seemed an opportune time to reduce API
  surface area on something that likely isn't used.
- Remove errorsInterface, as it's not useful without `(Errors)
  GetErrors()`
- Change `(*Errors) Add(error)` => `(Errors) Add(error...) Errors`
  because we can't modify even a *Errors when it's a type alias. This is
  more idiomatic as it follows the pattern of `slice = append(slice,
  element)` Go developers are familiar with.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -18,40 +18,38 @@ var (
 	ErrUnaddressable = errors.New("using unaddressable value")
 )
 
-type errorsInterface interface {
-	GetErrors() []error
-}
-
 // Errors contains all happened errors
-type Errors struct {
-	errors []error
-}
+type Errors []error
 
-// GetErrors get all happened errors
+// GetErrors gets all happened errors
 func (errs Errors) GetErrors() []error {
-	return errs.errors
+	return errs
 }
 
-// Add add an error
-func (errs *Errors) Add(err error) {
-	if errors, ok := err.(errorsInterface); ok {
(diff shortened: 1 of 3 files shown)


iwz:
Nice improvement. LGTM!

jinzhu:
Thank you for your PR.

calebthompson:
🎉

PR #1243, opened 2016-10-26
db.Raw().First() makes wrong sql fix #1214

### LINKED ISSUE #1214: db.Raw().First() makes wrong sql
db.Raw().First() add `)` to sql query

example

```
package main

import (
    "github.com/jinzhu/gorm"
    _ "github.com/go-sql-driver/mysql"
)

type User struct {
    ID   string `gorm:"primary_key"`
    Name string
}

func main() {
    db, _ := gorm.Open("mysql", "login:login@tcp(192.168.33.10:3306)/login?parseTime=true")
    defer db.Close()
    db.LogMode(true)

    var user User

    //OK select * from users WHERE id = '1'
    db.Raw("select * from users WHERE id = ?", 1).Scan(&user)

    //Fail select * from users WHERE id = '1') ORDER BY `users`.`id` ASC LIMIT 1
    db.Raw("select * from users WHERE id = ?", 1).First(&user)

}


```

smacker:
subj. fix #1214

### DIFF
--- a/query_test.go
+++ b/query_test.go
@@ -55,6 +55,24 @@ func TestFirstAndLastWithNoStdPrimaryKey(t *testing.T) {
 	}
 }
 
+func TestFirstAndLastWithRaw(t *testing.T) {
+	user1 := User{Name: "user", Emails: []Email{{Email: "user1@example.com"}}}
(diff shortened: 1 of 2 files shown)


jinzhu:
Thank you for your PR

PR #1251, opened 2016-11-04
Force UTC on db times

slockij:
In some cases, time fields are created without timezone.
Then, forgetting to use UTC in code can result in time related issues, and gorm.Model time fields will not have correct data.

Solution - force UTC on save

### DIFF
--- a/forced_utc_test.go
+++ b/forced_utc_test.go
@@ -1,0 +1,86 @@
+package gorm_test
+
+import (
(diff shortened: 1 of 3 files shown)


jinzhu:
I think database drivers could handle that

jinzhu:
eg https://github.com/go-sql-driver/mysql#loc

PR #1252, opened 2016-11-04
Block global updates / deletes

slockij:
In some cases (Error not checked, missed data) one can perform very harmful operation - global update or delete (all records)

DB.BlockGlobalUpdate(true) prevents it.


jinzhu:
Thank you for your PR

Issue #1249, opened 2016-11-03
Incorrectly formatted SQL when using numeric placeholders


fobispo-link:
When using the LogMode(true), the SQL produced during the log message, places the bounded variables in the wrong location, this is due to the fact that the current implementation splits the values with a regex similar to: `($\d+)|?`.

That works well for  `?` placeholders because they should be replaced in order.

However, the numeric placeholders should be replaced based on their value, for instance:

```
SELECT name,last FROM contact WHERE name=$2 and age=$1
```

Should NOT be replaced in order, because it will yield an incorrectly formed SQL command, which is not very useful for debugging.

The attached patch is capable of correctly replacing placeholder values using their contextual order.

[logger.go.diff.txt](https://github.com/jinzhu/gorm/files/570215/logger.go.diff.txt)

PR #1256, opened 2016-11-10
Issue  #1249 fix - replacing $ placeholders in order when logging

fobispo-link:
When using the LogMode(true), the SQL produced during the log message…, places the bounded variables in the wrong location, this is due to the fact that the current implementation splits the values with a regex similar to: ($\d+)|?.

That works well for ? placeholders because they should be replaced in order.

However, the numeric placeholders should be replaced based on their value, for instance:

SELECT name,last FROM contact WHERE name=$2 and age=$1
Should NOT be replaced in order, because it will yield an incorrectly formed SQL command, which is not very useful for debugging.

Fixes issue: #1249

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -12,8 +12,9 @@ import (
 )
 
 var (
-	defaultLogger = Logger{log.New(os.Stdout, "\r\n", 0)}
-	sqlRegexp     = regexp.MustCompile(`(\$\d+)|\?`)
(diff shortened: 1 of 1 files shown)


fobispo-link:
@jinzhu  have you had a chance to look into this? it would be great if you could merge this in, thanks

jinzhu:
Hi @fobispo-link 

Thank you for your PR.

PR #1284, opened 2016-12-20
Compile regexp ahead of time

sandalwing:
@jinzhu Please take a look.

This change pulls regexp compilation out of some (potentially) hot functions to be done ahead of time.

This leads to a small performance increase with little cost.

I can wrangle up some benchmarks if you really want proof but it should be fairly self-evident, this work will now only be done one time vs. several times. 

Signed-off-by: Xavier Sandal <sandalwing@sandalwing.com>

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -447,7 +447,12 @@ func (scope *Scope) callMethod(methodName string, reflectValue reflect.Value) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jinzhu:
Thank you for your PR.

sandalwing:
👍

Issue #1281, opened 2016-12-16
Callback after scope.row/score.rows call


smacker:
I implemented measurement of sql queries to NewRelic ( https://github.com/smacker/newrelic-context ) using callbacks. And it works just fine. But it misses DB queries like db.Rows(), db.Count() and maybe few others because Gorm executes callbacks only before call to underlying db driver. Right now I don't see any possibility to add callbacks after such calls without breaking changes to Gorm :(

I can create PR with breaking changes to support it, but I suppose it won't be merged ever.
@jinzhu can you help? maybe you have any ideas how to do it without modification of gorm or you see any smart way to make changes that won't affect existing code which already rely on `rowQueries` callback?

PR #1289, opened 2016-12-27
Callback for tracing sql queries

smacker:
To solve #1281 I came up with this solution...

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -947,6 +947,9 @@ func (scope *Scope) typeName() string {
 
 // trace print sql log
 func (scope *Scope) trace(t time.Time) {
+	scope.Set("gorm:trace-time", t)
+	scope.callCallbacks(scope.db.parent.callbacks.trace)
+
 	if len(scope.SQL) > 0 {
 		scope.db.slog(scope.SQL, t, scope.SQLVars...)
 	}
--- a/scope.go
+++ b/scope.go
@@ -8,7 +8,6 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
-	"time"
 
 	"reflect"
 )
@@ -346,9 +345,10 @@ func (scope *Scope) Raw(sql string) *Scope {
 
 // Exec perform generated SQL
 func (scope *Scope) Exec() *Scope {
-	defer scope.trace(NowFunc())
-
 	if !scope.HasError() {
+		scope.callCallbacks(scope.db.parent.callbacks.beforeSQL)
+		defer scope.callCallbacks(scope.db.parent.callbacks.afterSQL)
+
 		if result, err := scope.SQLDB().Exec(scope.SQL, scope.SQLVars...); scope.Err(err) == nil {
 			if count, err := result.RowsAffected(); scope.Err(err) == nil {
 				scope.db.RowsAffected = count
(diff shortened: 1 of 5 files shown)


jinzhu:
Hi @smacker 

Thank you for your PR, sorry just noticed your issue report, I think you could just make it works with set a new logger, so this might be not necessary.

Refer: https://github.com/jinzhu/gorm/blob/master/logger.go

Thank you.

smacker:
No, it won't work with a logger. I have to start newrelic transaction before sql started execution and finish it after. It measures not only time, but call stack, how much cpu it burn and other things.

smacker:
please reconsider trace callback. Code instrumentation is very useful. If you don't like callback, can you suggest any other way to do it? For example go-redis provides WrapProcess method: https://github.com/go-redis/redis/pull/352

scope.go
```diff
@@ -947,6 +947,9 @@ func (scope *Scope) typeName() string {
 
 // trace print sql log
 func (scope *Scope) trace(t time.Time) {
+	scope.Set("gorm:trace-time", t)
```
zardak: IMO, this is a sign of hacky approach. The more or less right way would be to introduce lifecycle hooks, like `BeforeX` and `AfterX` callbacks.

smacker: agreed. I'll try to improve it.

ipadavic:
I'm trying to add sql tracing to my app via datadog apm, this PR makes it possible. Any chance this could be merged?

zardak:
@jinzhu could you take a look at the pr please?

jinzhu:
Hi @smacker 

Sorry for the delay, just have time to review it and your code in https://github.com/smacker/newrelic-context/blob/master/gorm.go

I think for your problem you could resolve it with `RawQuery` callback, and it will be call in `Row`, `Rows`, `Count`...   Refer https://github.com/jinzhu/gorm/blob/master/callback.go#L78

smacker:
@jinzhu it was my first intention to use raw callback, but you call it before sql.DB methods, there is no way to know when it finish. For Query, Create, Delete and Update you call sql.DB inside callbacks, so I can add my callback before and after real query. But for RawQuery you do sql query inside scope, not callback. I also tried to change Row, Rows, Count, ... methods and move sql related code from scope to inside callbacks, but code becomes quite messy.

jinzhu:
Hi @smacker 

Thank you for your report, just fixed that problem. https://github.com/jinzhu/gorm/commit/c62e9bcabea2b3ed5e3a8bc1602cd38f8ad477f9

smacker:
@jinzhu your commit broke the code of anybody who altered the query in row callbacks. (your new callback registered before any existed one) JFK

jinzhu:
Yes, you are right, just pushed another commit to fix the behaviour with a warning message https://github.com/jinzhu/gorm/commit/7fb9b62c17d90320e0582a4720db025c1652fd6a

Thank you for the report.

ipadavic:
Sorry for commenting on closed PR, but I'm having trouble getting this to work with Raw(). Does "row_query" callback include Raw queries?

jinzhu:
No, `Raw` query is really raw, no callback included.

Only `Row`, `Rows`, `Count`, `Plack`... those methods include the callback.

PR #1308, opened 2017-01-08
BeforeSave test failed

arduanov:
BeforeSave event called but struct not changed
Sql update is broken, because doesn't have id column
```
INSERT INTO "uuid_structs" ("name") VALUES ('')
```

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -95,6 +95,27 @@ func TestStringPrimaryKey(t *testing.T) {
 	}
 }
 
+type UUIDStruct struct {
+	ID   string `gorm:"primary_key"`
+	Name string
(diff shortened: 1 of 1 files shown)


zardak:
`BeforeSave` is supposed to return an error, your function signature is wrong, i.e.

```
func (u *UUIDStruct) BeforeSave() (err error) {
	u.ID = "new_uuid"
	return
}
```

arduanov:
thanks `scope.SetColumn("ID", uuid.New())` works fine

@jinzhu please add this information in section 1.5 Callbacks
http://jinzhu.me/gorm/callbacks.html

PR #1309, opened 2017-01-10
Enable fractional seconds in Time column for Mysql

jiulongw:
Mysql (5.6.4 and up) supports up to 6 digits fractional seconds in
Time values. This patch adds "frac" field annotation to support it.

http://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html

Sample usage:
Foo { ts time.Time `sql:"frac:3"` }

### DIFF
--- a/dialect_mysql.go
+++ b/dialect_mysql.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 	"unicode/utf8"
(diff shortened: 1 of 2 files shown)


dialect_mysql.go
```diff
@@ -80,10 +81,18 @@ func (mysql) DataTypeOf(field *StructField) string {
 			}
 		case reflect.Struct:
 			if _, ok := dataValue.Interface().(time.Time); ok {
-				if _, ok := field.TagSettings["NOT NULL"]; ok {
```
jiulongw: The "NOT NULL" tag checking is removed because I don't see it is necessary here.  Caller will properly insert NOT NULL if tag exists.  The new test case added below also confirms "not null" tag works as expected.

jiulongw:
@jinzhu any thoughts about this patch?

jinzhu:
why not just write data type with tag? like this one:

image

refer:

http://jinzhu.me/gorm/models.html#model-definition

jiulongw:
@jinzhu `type` is a viable workaround but since gorm supports type reflection it looks cumbersome to have sql type annotating struct members.  `size:100` is better than `type:varchar(100)`, for example.

Having additional level of abstraction is also useful to handle difference in sql implementations.  For example, `type:timestamp(6)` doesn't work in all SQL implementations.  With a custom tag gorm can handle different dialects properly.

Please give a second thought :-)

jinzhu:
As this works only for mysql, so not going to support this tag, but just added an API for you to extend GORM to support it.

Some sample code:

```go
var defaultParser = gorm.ParseFieldStructForDialect
gorm.ParseFieldStructForDialect = func(field *StructField, dialect Dialect) (fieldValue reflect.Value, sqlType string, size int, additionalType string) {
	fieldValue, sqlType, size, additionalType = defaultParser(field, dialect)
	if dialect.GetName() == "mysql" {
		// if fieldValue is type is time...
		sqlType = "timestamp(3)"
	}

	return
}
```

PR #1311, opened 2017-01-11
solve issue scope deleted_at

xamu:


rolies106:
+1 Need this one

arifwiwoho2:
+1 idem

zardak:
Aren't you misusing the `DeletedAt` field? It's supposed to be of `*time.Time` type.

What's the purpose of this change?

xamu:
While i use Save function to update a record, its also changes the value of DeletedAt field to '0000-00-00 00:00:00' so when i query it using First function the record considered soft deleted by gorm because the condition only get the record when the column value is null.

actually it can be solve by either modify the query scope or the Save function to not touch deleted_at field.

jiulongw:
@xamu This happens when you use `time.Time` type for DeletedAt.  As @zardak mentioned it should be `*time.Time` type instead.

jinzhu:
If using *time.Time, its value should be NULL.

xamu:
Yup it works with *time.Time type.

It was my own mistake, thanks a lot guys.

PR #1315, opened 2017-01-16
Clarification on `DB.First(...)` method

xcoulon:
Documenting the limitation on the `func (s *DB) First(out interface{}, where ...interface{}) *DB` method where the column name is optional if the primary key is numeric. In other cases, the name of the column must be provided.


jinzhu:
thank you for your pr.

xcoulon:
Thanks, @jinzhu !

PR #1325, opened 2017-01-22
Add back-reference example to many-to-many doc

gmolveau:


jinzhu:
Hello @gmolveau 

Does below works for you?

```go
db.Model(&languages).Related(&users)
 //// SELECT * FROM "users" INNER JOIN "user_languages" ON "user_languages"."user_id" = "users"."id" WHERE  ("user_languages"."language_id" IN ('111','112,'113'))
```

jinzhu:
merged

PR #1326, opened 2017-01-22
Update models.md

JoeFerrucci:
Incorrect syntax.


zardak:
The sole purpose of this comment is to indicate that `UpdatedAt` field will be set to current time whenever user `Save`s or `Update`s one or more fields on existing record. Everything is correct, although it probably needs clarification.

PR #1341, opened 2017-02-06
Fix "Unsupported destination" error when value is pointer of pointer

andreynering:
Happens when you give `**User` instead of `*User` to `db.Save`, `db.Create`.


jinzhu:
Fixed this, thank you for your PR. https://github.com/jinzhu/gorm/commit/c730b30a7830a04a4a3d536edcb1f8c5f77d3482#diff-f66d859193e052cd2460e67ddc1f49e3L33

PR #1358, opened 2017-02-16
Add AddColumn method

mlanin:
I know that there is an automigration process and it works perfect, but it is sometimes sooo handy to have a common add column ability.
It doesn't break anything and made exactly like ModifyColumn or DropColumn. 

I will be very glad if you will add this feature.

### DIFF
--- a/main.go
+++ b/main.go
@@ -552,6 +552,13 @@ func (s *DB) AutoMigrate(values ...interface{}) *DB {
 	return db
 }
 
+// AddColumn add column with type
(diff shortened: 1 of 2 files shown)


jinzhu:
Hello @mlanin 

Thank you for your PR, and it looks good.

But not going to merge it, will add it as a dialect method in v2 (will also move ModifyColumn, DropColumn to Dialect also)

Thank you.

belek:
@jinzhu, hi, any updates on that? Can't find AddColumn in Dialect.

PR #1365, opened 2017-02-22
Added a zero value update test

shirayu:
Added a zero value update test, but it fails. 
This seems to be a bug.
Though I tried to fix it, I haven't found where to fix.
Can you tell the reason?

### DIFF
--- a/query_test.go
+++ b/query_test.go
@@ -594,6 +594,11 @@ func TestFindOrCreate(t *testing.T) {
 		t.Errorf("user should be created with search value and assigned attrs")
 	}
 
+	DB.Where(&User{Name: "find or create 4"}).Assign(User{Age: 0}).FirstOrCreate(&user4)
+	if user4.Name != "find or create 4" || user4.Id == 0 || user4.Age != 0 {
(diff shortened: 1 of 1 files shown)


zardak:
This is known and intendent behaviour, per http://jinzhu.me/gorm/crud.html#update
```go
// WARNING when update with struct, GORM will only update those fields that with non blank value
// For below Update, nothing will be updated as "", 0, false are blank values of their types
db.Model(&user).Updates(User{Name: "", Age: 0, Actived: false})
```

shirayu:
Thank you for the comment.
I understood the value is updated when I use
```go
DB.Where(&User{Name: "find or create 4"}).Assign("age", 0).FirstOrCreate(&user4)
```
instead of 
```go
DB.Where(&User{Name: "find or create 4"}).Assign(User{Age: 0}).FirstOrCreate(&user4)
```

But, it is inconvenient when there are several values to update like ``User{Name: "", Age: 0, Actived: false}``.
Are there any easy way to do like this?

jinzhu:
Use map like `map[string]interface{}{"Name": "", Age: 0}`

PR #1379, opened 2017-03-07
Remove SET_IDENTITY_INSERT for transactions in MS SQL

jbfm:
SET_IDENTITY_INSERT should be handled by each individual developer to avoid extra queries to the database. 

It also breaks most common use cases by being enabled by default.

Current implementation is also broken since it never disables SET_IDENTITY_INSERT, and it can only be enabled for one table at a time.

### DIFF
--- a/dialects/mssql/mssql.go
+++ b/dialects/mssql/mssql.go
@@ -12,14 +12,7 @@ import (
 	"github.com/jinzhu/gorm"
 )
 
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @jbfm 

Thank you for your PR, I have did some changes for mssql, we should fixed this issue.

Thank you.

PR #1384, opened 2017-03-14
Replace all use of *sql.DB with sqlCommon

ansel1:
Exporting sqlCommon as SQLCommon.

This allows passing alternate implementations of the database connection, or wrapping the connection with middleware.  This change didn't change any usages of the database variables.  All usages were already only using the functions defined in SQLCommon.

This does cause a breaking change in Dialect, since *sql.DB was referenced in the interface.

### DIFF
--- a/dialect.go
+++ b/dialect.go
@@ -14,7 +14,7 @@ type Dialect interface {
 	GetName() string
 
 	// SetDB set db for dialect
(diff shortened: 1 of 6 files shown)


jinzhu:
thank you for your PR.

Issue #1037, opened 2016-05-23
Composite primary keys can't be created on SQLite tables


djkaty:
Example table

type Account struct {
  foo    uint64 `"gorm:primary_key"`
  bar    uint64 `"gorm:primary_key"`
}

db.AutoMigrate(&Account{})

Problem:

Gorm creates an SQL statement like:

CREATE TABLE .... (foo primary key .... bar primary key) ...

Needed output:

SQLite only support the format:

CREATE TABLE ... (foo, bar, ...., primary key (foo, bar))

Is there a workaround?

abe-winter:
Ran into the same issue. I'm assuming it's because the sqlite dialect code is over-eager about assigning "primary key" to integer-typed primary key fields.

https://github.com/jinzhu/gorm/blob/master/dialect_sqlite3.go#L31

fuyufjh:
Ran into the same issue +1

apelisse:
Same. Actually the problem is even more nasty:

```
type Foobar struct {
   FieldA string `gorm:"primary_key"`
   FieldB int `gorm:"primary_key"`
}
```
Creates a auto-increment primary key on FieldB, and not FieldA. Just because FieldB is an int.

The following works though:
```
type Foobar struct {
   FieldA string `gorm:"primary_key"`
   FieldB string `gorm:"primary_key"`
}
```

jinzhu:
```go
package main

import (
	"github.com/jinzhu/gorm"
	_ "github.com/jinzhu/gorm/dialects/sqlite"
)

var db *gorm.DB

type Account struct {
	Foo uint64 `gorm:"primary_key;auto_increment:false"`
	Bar uint64 `gorm:"primary_key;auto_increment:false"`
}

func main() {
	var err error
	db, err = gorm.Open("sqlite3", "test.db")
	if err != nil {
		panic(err)
	}
	db.LogMode(true)

	db.AutoMigrate(&Account{})
}
```

d4l3k:
Is this behavior documented somewhere? I just ran into this issue and the behavior is very unexpected. 

I'd argue that this is a bug that should be fixed instead of just using workarounds. If it is expected, I think it should probably throw an error on AutoMigrate that the `primary_key` flag is being ignored.

PR #1399, opened 2017-03-24
sqlite3: Never set primary key in-line

apelisse:
Never set the primary key flag as part of the column definition. When
using sqlite3, this prevents mixed int/other type composite primary
key (because the int column always wins and get the index for itself).

Rather, let's just set the primary key at the end, whether it's composite
or not. Note that the int key will automatically be AUTOINCREMENT if
its type is "integer". For that reason, always use that type, rather
than "bigint", as bigint maps to "integer" eventually anyway.


apelisse:
I don't think this fix is correct, because it doesn't set the AUTO_INCREMENT field on the integer. It should do so only if it is a single-column primary key.

apelisse:
Potentially fixes #1037

PR #1402, opened 2017-03-27
fix a wrong sql

huwenchao:
`time.Now` is a function, it should be `time.Now()`


jinzhu:
Thank you for your pr

PR #1403, opened 2017-03-27
Add lost backslash

liubin:
Format is broken without the three backslasees


jinzhu:
thank you for your pr.

PR #1408, opened 2017-03-30
Fix ToDBName method

Konboi:
Thank you very much for letting me use it very conveniently

The behavior of the ToDBName method when the last capital letter comes in like FieldX has been corrected
FieldX becomes fieldx if it wants to become field_x

I am glad if you check it


jinzhu:
Thank you for your PR

PR #1423, opened 2017-04-06
Allow open to take transaction.

emil2k:
Need to skip the ping, otherwise results in a nil dereference.


jinzhu:
Maybe we can just use `dbSQL`, not necessary to call the `DB()` method.

Could you make the change? thank you.

emil2k:
Good call. I changed it.

jinzhu:
Thank you for your PR.

PR #1424, opened 2017-04-07
Add gorm.Gormw, an interface wrapper for gorm.DB

### LINKED ISSUE #729: Make the DB an interface for easier testing?

littledot:
Fixes #729. This PR adds an interface-equivalent of gorm.DB so that those who prefer mocking gorm.DB under test can easily do so. It also adds 2 functions Wrap() and Openw(), to facilitate creating gorm.Gormw instances.

This PR aims to achieve the same results as #805 by introducing a gorm.DB wrapper interface instead of making changes to gorm.DB. It introduces less code changes and ensures backwards-compatibility.

Unit tests are basic because I'd like to know your thoughts first.


Azuka:
This looks really great. @jinzhu do you think we'll be getting this in any time soon?

Selvatico:
Hello guys, please check my own bicycle which we use with **gorm** to mock for tests https://github.com/Selvatico/go-mocket . Maybe it could be useful too.

AgrimPrasad:
@jinzhu This will be very helpful indeed for writing unit tests... Any plans on getting this in soon?

jinzhu:
I think this is not necessary to be merged into GORM, you could extract those code and make a wrapper for gorm?

Thank you for your work.

kacperjurak:
Hi,

I took @littledot code and create an external package with this interface and wrapper - https://github.com/kacperjurak/igorm. All credit goes to @littledot.

PR #1435, opened 2017-04-13
Fixed build error

dand4:
I am currently unable to build this library (go 1.8). This fix resolves my issue.


jinzhu:
This syntax should works with go 1.8, maybe your local shell env, for example `GOROOT`, `GOPATH` is incorrectly configured.

Issue #1020, opened 2016-05-18
Add support for check constraints


ababo:
As far as I know currently we cannot simply add check constraints without using raw SQL. I would like GORM to support that capability.

hugows:
+1. My use case specifically is the "singleton row" like http://stackoverflow.com/questions/33104101/ensure-sqlite-table-only-has-one-row

jinzhu:
There is a `AddForeignKey` method, for auto migrate the constraints, it is duplicated with https://github.com/jinzhu/gorm/issues/450

andreynering:
I opened a PR which adds support to check contraints, but unfortuanaly not for SQLite:

https://github.com/jinzhu/gorm/pull/1452

PR #1452, opened 2017-04-27
Add methods do add and drop check constraints

andreynering:
Ref #1020

- [x] Do only one thing
- [x] No API-breaking changes

Adds `AddCheckConstraint` and `DropCheckConstraint` methods.

Just like `AddForeignKey`, these two methods won't work on SQLite, which require them to be created in the `CREATE TABLE` sentense, but is not possible on a `ALTER TABLE`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -602,6 +602,22 @@ func (s *DB) AddForeignKey(field string, dest string, onDelete string, onUpdate 
 	return scope.db
 }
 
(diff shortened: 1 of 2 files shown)


jinzhu:
Not going to merge this one rightnow, but will add it as a dialect method in v2, thank you for your PR.

PR #1462, opened 2017-05-05
Removed unnecessary cloning.

emil2k:
`NewScope` clones `DB` no need to chain a call to clone with `NewScope`.


jinzhu:
Thank you for your PR.

PR #1465, opened 2017-05-07
Add support for removing foreign keys

samgavinio:

Add `RemoveForeignKey` functionality to drop foreign keys. Adding a foreign key manually is already supported so it makes sense to have the inverse operation. 

Tested to work with mysql and postgresql. Should also work with mssql but I couldn't test it out since I didn't have the time to set up a windows machine. Dropping foreign keys for sqlite isn't supported without a [work-around](https://www.techonthenet.com/sqlite/foreign_keys/drop.php), so I left this one out.

Added some tests for the foreign key migrations. Expect these tests to fail when run against sqlite, should I just not include these tests?

- [x] Do only one thing
- [x] No API-breaking changes

### DIFF
--- a/dialect.go
+++ b/dialect.go
@@ -29,6 +29,8 @@ type Dialect interface {
 	HasForeignKey(tableName string, foreignKeyName string) bool
 	// RemoveIndex remove index
 	RemoveIndex(tableName string, indexName string) error
+	// RemoveForeignKey remove foreign key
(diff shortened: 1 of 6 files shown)


jinzhu:
Thank you for your work, it looks good, but not going to merge it, but will make it as a dialect method in v2.

Thank you.

PR #1470, opened 2017-05-11
[WIP] Batch insert through a new callback

drgomesp:
References #255

**DISCLAIMER**: This is a work in progress.

Allow batch insert operations through the introduction of a new callback type. 

### Proof of Concept

```go
func main() {
	type Foo struct {
		ID   int64  `gorm:"column:id;primary_key"`
		Text string `gorm:"column:text"`
	}

	db := connect()
	defer db.Close()

	foos := []*Foo{}

	for i := 0; i < 10000; i++ {
		foo := &Foo{Text: "Foo"}
		foos = append(foos, foo)
	}

	start := time.Now()
	for _, foo := range foos {
		db.Create(foo)
	}
	elapsed := time.Since(start)

	log.Printf("Multiple queries took %s", elapsed)

	foos = []*Foo{}

        for i := 10001; i <= 20000; i++ {
		foo := &Foo{Text: "Foo"}
		foos = append(foos, foo)
	}

	start = time.Now()
	db.CreateBatch(foos)
	elapsed = time.Since(start)

	log.Printf("Single query took %s", elapsed)
}
```

Output:

```bash
2017/05/11 08:13:27 Multiple queries took 10.361453551s
2017/05/11 08:13:27 Single query took 81.625904ms
```

### DIFF
--- a/callback.go
+++ b/callback.go
@@ -15,12 +15,13 @@ var DefaultCallback = &Callback{}
 //   Field `rowQueries` contains callbacks will be call when querying object with Row, Rows...
 //   Field `processors` contains all callback processors, will be used to generate above callbacks in order
 type Callback struct {
-	creates    []*func(scope *Scope)
-	updates    []*func(scope *Scope)
-	deletes    []*func(scope *Scope)
-	queries    []*func(scope *Scope)
-	rowQueries []*func(scope *Scope)
-	processors []*CallbackProcessor
+	creates      []*func(scope *Scope)
+	createsBatch []*func(scope *Scope)
+	updates      []*func(scope *Scope)
+	deletes      []*func(scope *Scope)
+	queries      []*func(scope *Scope)
+	rowQueries   []*func(scope *Scope)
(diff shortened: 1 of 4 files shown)


jinzhu:
Hello @drgomesp 

Thank you for your work, I would prefer to extend `Create` to support batch insert.

Also we should call `Before/After Create/Save` callbacks when do the bulk insert.

So we could change the `createCallback` method in `callback_create.go` to do the work.

callback.go
```diff
@@ -58,6 +60,11 @@ func (c *Callback) Create() *CallbackProcessor {
 	return &CallbackProcessor{kind: "create", parent: c}
 }
 
+// CreateBatch could be used to register callbacks for creating objects in bulk operations
+func (c *Callback) CreateBatch() *CallbackProcessor {
+	return &CallbackProcessor{kind: "creates_batch", parent: c}
```
mchampaneri: kind is create_batch  .  you have written creates_batch

PR #1517, opened 2017-06-30
Updated scope.go to always quote when adding index

bbstk:
I am using numbers for column names (to be compatible with protobuf) and adding unique index to them does not work since they are not quoted. I do not see a reason to check if the column name is a string in order to quote it. Correct me if I am wrong.

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -448,7 +448,7 @@ func (scope *Scope) callMethod(methodName string, reflectValue reflect.Value) {
 }
 
 var (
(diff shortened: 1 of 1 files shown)


jinzhu:
Hello @bbstk 

I think you should change scope.go#451 to:

```go
columnRegexp        = regexp.MustCompile("^[a-zA-Z\d]+(\\.[a-zA-Z\d]+)*$") // only match string like `name`, `users.name`
```

Could you check if it works for you?

bbstk:
Thank you for the quick response, @jinzhu 

I tested the change you proposed and it works as intended. I updated the PR.

jinzhu:
Cool, thank you for your PR.

PR #1536, opened 2017-07-14
Changed the type of uint32 from integer to bigint in postgres

bbstk:
The integer type in postgres is 4 bytes. Since it is also signed, when using uint32 with high bit set you will get:
`pq: value "2854263694" is out of range for type integer`
To prevent this uint32 should be bigint in postgres.

### DIFF
--- a/dialect_postgres.go
+++ b/dialect_postgres.go
@@ -30,14 +30,14 @@ func (s *postgres) DataTypeOf(field *StructField) string {
 		switch dataValue.Kind() {
 		case reflect.Bool:
 			sqlType = "boolean"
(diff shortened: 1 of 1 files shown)


bbstk:
The same is true for uint64 being converted to bigint(8 bytes). If the high bit for uint64 is set it will not fit in bigint. However, go's database/sql does not allow using uint64 with high bit set (https://github.com/golang/go/issues/20238, https://github.com/golang/go/issues/9373).
This is why it doesn't make sense to have uint64 be numeric since it will never be allowed with high bit set.

jinzhu:
Thank you for your PR.

PR #1541, opened 2017-07-21
i think it's better,please think about it

liu-xuewen:

- [] Do only one thing
- [] No API-breaking changes
- [] Write good commit message, try to squash your commits into a single one

Thank you.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


liu-xuewen:
The sql print  missing return result,can i add it ?thanks very much!

jinzhu:
Thank you for your PR.

andreynering:
Nice feature! But maybe red is not the right color for it? Red is the color that currently print errors. Maybe change to another one?

PR #1546, opened 2017-07-26
 support to fills data to nil struct pointer

icoder528:
支持类似这样的操作：
func userByID(id int)(user *User,err error){
    err = db.Find(&user,id).Error
    return
}


jinzhu:
seems like it broken tests.

jinzhu:
```go
package main

import (
	"fmt"
	"reflect"

	"github.com/jinzhu/gorm"
	_ "github.com/jinzhu/gorm/dialects/sqlite"
)

type User struct {
	gorm.Model
	Name string
}

func FindUser() (user *User) {
	fmt.Println(reflect.ValueOf(user).CanSet())
	return
}

func main() {
	FindUser()
}
```

Seems like the value is always not settable?

PR #1558, opened 2017-08-06
Fix postgres dialect for databases with multiple schemas

lukasdietrich:
If a postgres database contains more than one
schema methods like HasTable(...) would return
true even if the current schema does not contain a
table with that name.

The [INFORMATION_SCHEMA](https://www.postgresql.org/docs/9.6/static/infoschema-tables.html) tables have a column 
named `table_schema` and the [pg_indexes](https://www.postgresql.org/docs/9.6/static/view-pg-indexes.html) table a 
similar one named `schemaname`.

I added a check to the `where` clause of the respective
queries to only look at the [current schema](https://www.postgresql.org/docs/9.6/static/functions-info.html).

### DIFF
--- a/dialect_postgres.go
+++ b/dialect_postgres.go
@@ -85,7 +85,7 @@ func (s *postgres) DataTypeOf(field *StructField) string {
 
 func (s postgres) HasIndex(tableName string, indexName string) bool {
 	var count int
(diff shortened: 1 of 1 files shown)


jinzhu:
thank you for your PR.

PR #1566, opened 2017-08-13
Erros skip nil in Add function

knqyf263:

- [x] Do only one thing
- [x] No API-breaking changes

Thank you.

To skip nil in `Add` .

I wrote the following code.
```
conn *gorm.DB
...
tx := conn.Begin()
var errs gorm.Errors
errs = errs.Add(tx.Delete(&foo).Error)
errs = errs.Add(tx.Delete(&bar).Error)

if errs.Error() != "" {
  tx.Rollback()
  return
}
...
```

I got the error.
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x43e6f75]
```

I checked `errs` variable.
```
gorm.Errors{
  nil,
}
```

I think `Errors` doesn't need to have `nil`.
Or should I always write like this?
```
if err := tx.Delete(&foo).Error; err != nil {
  errs = errs.Add(err)
}
```

Is not it supposed that `nil` will be added?

Thanks!

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -29,6 +29,10 @@ func (errs Errors) GetErrors() []error {
 // Add adds an error
 func (errs Errors) Add(newErrors ...error) Errors {
 	for _, err := range newErrors {
+		if err == nil {
+			continue
+		}
+
 		if errors, ok := err.(Errors); ok {
 			errs = errs.Add(errors...)
 		} else {


jinzhu:
thank you for your PR.

Issue #1572, opened 2017-08-15
Support for cloudsqlpostgres driver/dialect type:feature


paulwalker:
Use of google's cloud sql proxy requires a special driver named `cloudsqlpostgres`. This requires the postgres dialect, but the postgres dialect instance itself is not public and cannot be accessed in a call to RegisterDialect.

If the individual dialects or the dialectsMap were public, it would allow client code of this package to register custom drivers such as cloudsqlpostgres: https://github.com/GoogleCloudPlatform/cloudsql-proxy#to-use-inside-a-go-program. Alternatively, consider adding support directly for cloudsqlpostgres by simply registering the existing postgres dialect with an additional key.

cgestes:
Or just add in dialect_postgres.go:

```go
	RegisterDialect("postgres", &postgres{})
+	RegisterDialect("cloudsqlpostgres", &postgres{})
```

cgestes:
I guess your patch and this one would be best. I'am making a PR.

Your patch allow to access already defined dialect, but it does add support for cloudsqlpostgres.

paulwalker:
Right, my thinking was just to make the dialect map public and client code could use pre-registered dialects, as ours does now. But I won't complain about native support for `cloudsqlpostgres` as well. :-)

tommg:
I'm really interested in this, but I have tried this patch and it doesn't seem to be working for me. I get the following problem:
`panic: sql: Register called twice for driver postgres goroutine 1 [running]: database/sql.Register(0xb5e0e5, 0x8, 0xe26e80, 0xe82800) /usr/local/go/src/database/sql/sql.go:50`

Any idea what I'm doing wrong here, having huge problems trying to get this to work with Postgres cloudsql

PR #1573, opened 2017-08-15
makes DialectsMap public for easier registration of dialects using ex…

paulwalker:
…isting values refs #1572

- [X] Do only one thing
- [X] No API-breaking changes
- [X] Write good commit message, try to squash your commits into a single one
- [NA] Run `./build.sh` in `gh-pages` branch for document changes

Thank you.

Makes DialectsMap public so that its existing values can be referenced when registering a new driver/dialect that uses the same underlying driver, e.g. 
`gorm.RegisterDialect("cloudsqlpostgres", gorm.DialectsMap["postgres"])`

### DIFF
--- a/dialect.go
+++ b/dialect.go
@@ -48,6 +48,7 @@ type Dialect interface {
 	CurrentDatabase() string
 }
 
+// DialectsMap map of registered dialects
 var DialectsMap = map[string]Dialect{}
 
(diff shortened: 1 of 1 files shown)


jinzhu:
Going to merge PR https://github.com/jinzhu/gorm/pull/1577 for now, if there are another use case like this, will consider to have a `GetDialect` to get registered dialect.

PR #1584, opened 2017-08-23
1. add structField.Offset 

bigpigeon:
[with this issue](https://github.com/jinzhu/gorm/issues/1583)


bigpigeon:
I Used unsafe.OffsetOf to calculate field offset ,but sub embeded struct filed offset  was relative with sub embeded struct

jinzhu:
Hi @bigpigeon 

This PR won't be merged, as the Offset is not used in GORM, and I am not really sure if it is commonly used or not...

Sorry for that.

PR #1587, opened 2017-08-24
bugfix for LimitAndOffsetSQL

SenZhangAI:
if limit = -1 and offset >= 0, sql with only OFFSET but not LIMIT ahead will cause syntax error (test in mysql 5.6) as OFFSET is based on LIMIT.


jinzhu:
limit can't be negative? so I think this should not be handled by GORM. and some databases doesn't require `limit` when using `offset`

SenZhangAI:
Thanks

PR #1598, opened 2017-08-31
Fixed handling of driver.Valuer 

### LINKED ISSUE #1596: Use Valuer before trying to cast to byte slice
https://github.com/jinzhu/gorm/blob/e5432b14d2f28ad759d8d6262c1f8a167d517f73/scope.go#L561

It should use Valuer interface prior to trying to cast it in byte slice.

Currently it's impossible to use Valuer interface to define custom serialization for byte slice fields in models.

I.e. you have some data you want to encrypt/decrypt upon saving/retrieving from DB.

destel:
This PR fixes #1596 

It changes how slices are expanded when building WHERE conditions
```
db.Where("a = ?", param)
```

Current logic:
- If param is non byte slice - expand it
- If param is byte slice - use it as is (as binary data)
- If param is driver.Valuer - call param.Value() without checking for error
- Otherwise - use param as is

Slice types, even those that implement driver.Valuer were always expanded. So it was not possible to use custom slice types. For example: 

```
type EncryptedValue []byte

func (v EncryptedValue) Value() (driver.Value, error) {
...
}

func (v *EncryptedValue) Scan(value interface{}) error {
 ...
}
```

I replaced the logic with the following
- If param is non byte slice and non driver.Valuer - expand it
- If param is driver.Valuer - call param.Value() and **check for errors**
- Otherwise - use param as is

Also this change affects custom byte slice types which don't implement driver.Valuer
```
type BinData []byte
```

Previously such params were always expanded as slices. After the changes I made such param will be passed as is (as binary data)

### DIFF
--- a/migration_test.go
+++ b/migration_test.go
@@ -32,6 +32,7 @@ type User struct {
 	CompanyID         *int
 	Company           Company
 	Role              Role
+	Password          EncryptedData
 	PasswordHash      []byte
 	IgnoreMe          int64                 `sql:"-"`
 	IgnoreStringSlice []string              `sql:"-"`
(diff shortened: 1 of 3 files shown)


jinzhu:
Merged, thank you for your PR.

PR #1599, opened 2017-08-31
add type created_at deleted_at update_at

newgoo:
增加在创建时间，删除时间的类型，可以是int64和string的类型


jinzhu:
Hi @newgoo 

Thank you for your work, it looks good.

But I am not going to merge it, think it might be better as a GORM plugin.

Thank you.

PR #1614, opened 2017-09-15
Correct ModifyColumn SQL syntax.

perrito666:
The generated SQL for ModifyColumn was:

`ALTER TABLE "tablename" MODIFY "columname" type`

But should have been:

`ALTER TABLE "tablename" ALTER COLUMN "columname" TYPE type`

since Modify does not seem to be entirely compatible with all Engines

- [] Do only one thing
- [] No API-breaking changes
- [] Write good commit message, try to squash your commits into a single one

Thank you.

### DIFF
--- a/migration_test.go
+++ b/migration_test.go
@@ -5,6 +5,7 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
+	"os"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


jinzhu:
Hi @perrito666 

I have written a test for ModifyColumn in branch https://github.com/jinzhu/gorm/tree/perrito666-master, seems your changes haven't pass the test, could you look at it?

perrito666:
Ok the issue is actually with Sqlite, it does not fully support `ALTER TABLE` it does not support specifically changing a column type (when reverting my change you will notice the error persists) https://sqlite.org/lang_altertable.html is there a way to tell the test to only run when using certain dialects?

jinzhu:
Hi @perrito666 

You could use https://github.com/jinzhu/gorm/blob/master/create_test.go#L97 , could you take care of it?

perrito666:
Yes I’ll get it sorted and up in the PR tomorrow morning it’s a bit late
here, thank you for the test and sorry I did not include one with the PR
originally.

perrito666:
There you go, I made it to check explicitly the dialects that fully support `ALTER TABLE` with `ALTER COLUMN` let me know if I should fix something else on the PR, cheers.

PR #1619, opened 2017-09-18
Add the ability to set the default Logger

glb:
Sometimes you want to have a default logger that is used for all of your database connections; you can now set the default logger using `gorm.SetDefaultLogger`.

- [x] Do only one thing
- [x] No API-breaking changes
- [x] ~~Run `./build.sh` in `gh-pages` branch for document changes~~ not applicable?

This adds a `SetDefaultLogger` method to allow you to replace the default logger that `gorm` uses.

I did this for two reasons: first, I wanted to be 100% certain that any logs created during `gorm.Open()` got sent to the right place; second, I wanted to just set the logger once and have it be used for all DB connections.

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -18,6 +18,11 @@ var (
 	numericPlaceHolderRegexp = regexp.MustCompile(`\$\d+`)
 )
 
+// SetDefaultLogger replaces the default Logger instance.
+func SetDefaultLogger(l Logger) {
+	defaultLogger = l
+}
+
 func isPrintable(s string) bool {
 	for _, r := range s {
 		if !unicode.IsPrint(r) {


jinzhu:
Hi @glb 


You can just set logger after initailize `DB`?

like 

```go
DB, err := gorm.Open("...")
DB.SetLogger(logger)
```

For your code, you have to set default logger before initailize DB, or it will have execute order issue.

glb:
Hi @jinzhu, thank you for your thoughts & this package. Is there anything that can cause logging in the `gorm.Open` call though? This is what I worry about -- if any logging happens during `gorm.Open` (now or in the future -- is it possible that someone will want to add debugging logs in the `Open` call flow?) then logs will go to the wrong place. Also, I would like to be able to set the default logger once similar to how I can set default callbacks. Something like:
```go
gorm.SetDefaultLogger(...)
gorm.DefaultCallback.Query().Register(...)
```
in an application initialization routine, and then I don't need to set it elsewhere.

jinzhu:
Hi @glb 

GORM won't print any log during `Open`.

For register callbacks, you can register it to a DB instance like:

```go
DB.Callback().Query().Register(...)
```

So you could initialize a DB instance first, then set it's logger, callbacks when initialize your application.

Change the global variable is not recommended, or its behavior will depends on your code's order, and hard to debug it...  (maybe it's better to hide `DefaultCallback`...)

glb:
Thanks @jinzhu ... maybe I'm misunderstanding the intended usage of `Open()`; is it safe to keep a `DB` reference and use it in multiple goroutines?

Or are you saying that I should approach this by writing a method that calls `Open` and sets the logger + callbacks, and use that method everywhere I need to get a database connection to perform a transaction? In fact I do have a method that wraps `Open`, but I would like to keep the contents of that method simple and avoid re-registering callbacks and doing other initialization every time I need a database connection.

There is already a default logger that gets copied into the `DB` instance, all this PR does is makes it possible for someone to set that value and reduce the work done every time they call `Open`. People who wanted to keep working the other way still could do so.

I also struggle with having `defaultLogger` set to something that I will _never_ want for my application; I need logs to be written & formatted differently. Knowing it is set to write to stdout makes me nervous that some critical log could get missed if someone forgets to call `DB.SetLogger`.

I'm sorry to be difficult, I am just trying to understand the intended workflow.

PR #1622, opened 2017-09-22
Call Query callback chain when preloading many2many

glb:
When using `Preload` on a `many2many` association, the `Query` callback chain was not being called. This made it difficult to write a plugin that could reliably get called regardless of how objects were being queried.

Now `handleManyToManyPreload` will call the `Query` callback chain for each object that is retrieved by following the association.

Since the data has already been read by the`handleManyToManyPreload` method, a new scope setting called `gorm:skip_queryCallback` is set to `true` before calling the callbacks. Callbacks can check for the presence of this setting if they should not be run; the default `queryCallback` is an example of this case.

Fixes jinzhu/gorm#1621.

- [x] Do only one thing
- [x] No API-breaking changes
- [ ] ~~Run `./build.sh` in `gh-pages` branch for document changes~~ not applicable?

### DIFF
--- a/callback_query.go
+++ b/callback_query.go
@@ -15,6 +15,10 @@ func init() {
 
 // queryCallback used to query data from database
 func queryCallback(scope *Scope) {
+	if _, skip := scope.Get("gorm:skip_query_callback"); skip {
+		return
+	}
+
 	defer scope.trace(NowFunc())
 
 	var (
--- a/callback_query_preload.go
+++ b/callback_query_preload.go
@@ -324,6 +324,10 @@ func (scope *Scope) handleManyToManyPreload(field *Field, conditions []interface
 
 		scope.scan(rows, columns, append(fields, joinTableFields...))
 
+		scope.New(elem.Addr().Interface()).
+			Set("gorm:skip_query_callback", true).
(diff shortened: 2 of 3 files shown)


callback_query.go
```diff
@@ -15,6 +15,10 @@ func init() {
 
 // queryCallback used to query data from database
 func queryCallback(scope *Scope) {
+	if _, skip := scope.Get("gorm:skip_queryCallback"); skip {
```
glb: @jinzhu I don't know if this name is right; I would be happy to change it.

jinzhu:
Hi @glb

Thank you for your PR, could you change `gorm:skip_queryCallback` to `gorm:skip_query_callback`?

glb:
Hi @jinzhu, thanks for taking a look at it! I've made the requested change and rebased to latest `master`.

glb:
Hi @jinzhu, is everything OK to proceed with this PR? Would you like me to make any other changes?

jinzhu:
Hi @glb 

Thank you for your PR, and sorry for the delay.

glb:
Thank you @jinzhu ! Much appreciated!

PR #1626, opened 2017-09-30
add JSONB type

Aetheus:
Add a JSONB type to the postgres dialect. I'm aware of @Niondir raising an issue for this (https://github.com/jinzhu/gorm/issues/1183), but it's been nearly a year since then and as I don't see any current PRs for this so I have decided to go ahead and make a PR for it so future users of the ORM won't have to constantly reinvent this struct.

### DIFF
--- a/dialects/postgres/postgres.go
+++ b/dialects/postgres/postgres.go
@@ -6,6 +6,9 @@ import (
 
 	_ "github.com/lib/pq"
 	"github.com/lib/pq/hstore"
+	"encoding/json"
+	"errors"
+	"fmt"
 )
 
 type Hstore map[string]*string
@@ -51,4 +54,22 @@ func (h *Hstore) Scan(value interface{}) error {
 	}
 
 	return nil
+}
+
+// for Postgresql's JSONB data type
+type Jsonb struct {
+	json.RawMessage
+}
+
+func (j Jsonb) Value() (driver.Value, error){
+	return j.MarshalJSON()
+}
+
+func (j *Jsonb) Scan (value interface{}) error {
+	bytes, ok := value.([]byte);
+	if !ok {
+		return errors.New(fmt.Sprint("Failed to unmarshal JSONB value:", value))
+	}
+
+	return json.Unmarshal(bytes, j)
 }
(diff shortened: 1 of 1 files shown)


Aetheus:
Not entirely certain why codeclimate is failing this PR on the very first line of postgres.go ("package postgres"), as this line was unmodified, and seems perfectly legal anyway considering that its the intended package name. 

The failed check has no description either, so I'm not entirely sure what needs to be modified to pass the check. Any possible advice, @jinzhu ?

jvoorhis:
+1

I cannot reproduce the codeclimate warning with the latest tools.

```
$ go fmt  // no output
$ echo $?
0
$ go version
go version go1.9 linux/amd64
```

jinzhu:
thank you for your PR

SongCF:
How do I use it?Can you give me an example?

Aetheus:
I saw a comment here earlier asking about how to use this. It looks like it's since been deleted, but for other folks who might be curious:


Assuming you have a "documents" table, and in that documents you have a "metadata" jsonb column, your resulting model struct definition may look something like this:

````
package yourPackageName

import (
	"time"
	"github.com/jinzhu/gorm/dialects/postgres"
)

type Document struct {
	Metadata postgres.Jsonb		`json:"requirements"`
	Body string				        `json:"body"`

	//housekeeping attributes
	ID        int					        `json:"id"`
	CreatedAt time.Time			`json:"created_at"`
	UpdatedAt time.Time			`json:"updated_at"`
}

````

Assuming you wanted to access the value of "Metadata" in a specific "Document" as a string ... 

````
func GetMetadataOfDocument(dbConn *gorm.DB, documentId int) string {
    doc := Document{}
    dbConn.Where("id = ?", documentId).First(&doc)

    stringVal := string(doc.Metadata.RawMessage)
    return stringVal;
}
````

januszm:
@Aetheus thanks for this, but it seems that you have to manually convert the `Metadata` from bytes to string. What if you simply want to use that `Jsonb` column in the API response built with `json.NewEncoder(w).Encode(result)`  (where, let's say that `result` contains the `Document` from your example).

In the above case what I see in the response is the array of bytes `[123 34 56 ...]` and I'd like to have them converted to string somehow.

PR #1647, opened 2017-10-20
corresponds timestamp precision for mysql

miya-masa:

- [x] Do only one thing
- [x] No API-breaking changes

Thank you.

corresponds timestamp precision for mysql

### DIFF
--- a/dialect_mysql.go
+++ b/dialect_mysql.go
@@ -96,9 +96,9 @@ func (s *mysql) DataTypeOf(field *StructField) string {
 		case reflect.Struct:
 			if _, ok := dataValue.Interface().(time.Time); ok {
 				if _, ok := field.TagSettings["NOT NULL"]; ok {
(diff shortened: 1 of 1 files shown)


ozkar99:
An alternative is adding `gorm:"type:timestamp(6)"` (or the precision you want)  to your timestamps on the struct annotations.

miya-masa:
If precision is specified, sqlite or postgres will not be timestamp type.
Only mysql needs precision.

TwistedHardware:
Can you also add a default size in case the size tag was not passed. This can some in handy when you are dealing with `UpdatedAt` from `gorm.Model` where you don't really have cannot really edit that tag. I good default size would be 6 because it is the largest possible precision. By default in SQLite, you will get precision of 9 to match the nano second precision of Golang time.

jinzhu:
Merged, but you need to use tag `precision` to set timestamp's precision

PR #1658, opened 2017-10-25
CreatedAt,UpdatedAt,DeletedAt  from time.time to time.Unix

wangzhipeng:
CreatedAt,UpdatedAt,DeletedAt  from time.time to time.Unix


andreynering:
I'm not the maintainer, just having a look...

Interesting idea, but this will break existing applications.

A possible solution is:

- If type is `time.Time`, set `time.Now()`
- If type is `int64`, set `time.Now().Unix()`

jinzhu:
Not going to merge this one, but you could write some callbacks to do the job, maybe you could extract it as a GORM plugin.

PR #1662, opened 2017-10-30
check column exists modify

inkbx:

- [x] Do only one thing
- [x] No API-breaking changes

Thank you.

when auto migrate, check column exists and process one by one, that is not efficient. 

i change a another way do this.


jinzhu:
seems it doesn't support all dialects?

PR #1629, opened 2017-10-04
Allows foreign keys to be saved without saving the assoication when s…

joe-at-startupmedia:
…pecified #1628

### What did this pull request do?

Allow foreign keys to be saved when `save_assocations` is false.

Issue #1628, opened 2017-10-04
save_associations annotation preventing associating from being saved but not saving foreign key


joe-at-startupmedia:
### What version of Go are you using (`go version`)?

go version go1.8.3 linux/amd64

### Which database and its version are you using?
```
  if config.Config.DB.Adapter == "postgres" {
    DB, err = gorm.Open("postgres", fmt.Sprintf("postgres://%v:%v@%v/%v?sslmode=disable", dbConfig.User, dbConfig.Password, dbConfig.Host, dbConfig.Name))
```

### What did you do?

Model File

```
type Program struct {
  ID                       uint `gorm:"primary_key"`
  CauseId                  uint
  Cause                    Cause `gorm:"save_associations:false"`
  SponsorId                int
  Sponsor                  Cause 
}
```

When I save the model:
- SponsorId foreign key is populated respectfully but the association is also saved.
- CauseId foreign key is not populated and association is not saved

Why doesn't this code just return the foreign key instead of returning nothing?

`github.com/jinzhu/gorm/callback_save.go`
```
func saveBeforeAssociationsCallback(scope *Scope) {
  if !scope.shouldSaveAssociations() {
    return
  }
  for _, field := range scope.Fields() {
    if ok, relationship := saveFieldAsAssociation(scope, field); ok && relationship.Kind == "belongs_to" {
      fieldValue := field.Field.Addr().Interface()
      scope.Err(scope.NewDB().Save(fieldValue).Error)
      if len(relationship.ForeignFieldNames) != 0 {
        // set value's foreign key
        for idx, fieldName := range relationship.ForeignFieldNames {
          associationForeignName := relationship.AssociationForeignDBNames[idx]
          if foreignField, ok := scope.New(fieldValue).FieldByName(associationForeignName); ok {
            scope.Err(scope.SetColumn(fieldName, foreignField.Field.Interface()))
          }
        }
      }
    }
  }
}
```

PR #1664, opened 2017-11-01
Association foreign key save

ezk84:
Allow foreign keys to be saved when `save_assocations` is false (More details: #1628)

This is an update to PR #1629, with the tests needed to hopefully get the code merged. 

Wasn't able to squash as there are commits here from @joe-at-startupmedia.


jinzhu:
Hi @ezk84 

Thank you for your work, have merged this PR, and make it more advanced.

Like you could disable updating associations have a primary key, and keep the ability to create new values.

More details please refer: http://gorm.io/crud.html#associations

Thank you.

PR #1683, opened 2017-11-17
Add `Expecter` for easier unit testing and DB mocking

iantanwx:

- [x] Do only one thing
- [x] No API-breaking changes

Thank you.

This PR is a proposal to address unit testing issues like https://github.com/jinzhu/gorm/issues/1525 and https://github.com/jinzhu/gorm/issues/711.

Since `gorm.DB` is a `struct`, it cannot be easily mocked. As a result, unit testing code that depends on `gorm` requires one of:

1.  Writing a wrapper and injecting that wrapper into client code instead of directly using `*gorm.DB`
2. Using https://github.com/DATA-DOG/go-sqlmock or https://github.com/erikstmartin/go-testdb in compatibility mode

The first option involves a high level of complexity since wrappers cannot access private `gorm` methods, and might easily break even for non-breaking changes in `gorm`.

The second option requires verbose code, since both `go-sqlmock` and `go-testdb` expose a very low level API for stubbing queries/execs and their results. This requires client code to guess what queries `gorm` is generating, the arguments it is passing to prepared statements, and what is coming back from the `database`, which `gorm` should be hiding. This makes the abstraction very leaky when unit testing, for example, complex DAOs.

As a result, this PR proposes to hook directly into private methods for generating SQL, and, through an `Adapter` interface, expose an `Expecter` struct that mirrors the `gorm.DB` API, and can be used like so:

```
expect, db, err := gorm.NewDefaultExpecter()
dao := SomeDao{db: db}
// generate SQL and pass it to sqlmock through `Adapter` interface
q := expect.Model(&User{}).Where("name = ?", "jinzhu").Find(&{user}) // this returns an `ExpectedQuery`
q.Returns(&User{Name: "jinzhu")

dao.FindUserByName("jinzhu")
// err will contain failure message from the mock db library if expectation is not met
err := expect.ExpectationsMet()

```

This seems to me to be the best solution for now, as it exposes extra functionality without touching any of the existing API surface area. As you can say there is quite some way to go. Please let me know what you think.


expecter_test.go
```diff
@@ -144,3 +144,29 @@ func TestMockPreload(t *testing.T) {
...
+	out := User{Id: 1, Languages: languages}
+
+	expect.Preload("Languages").Find(&in).Returns(out)
+	db.Preload("Languages").Find(&in)
+
+	if err := expect.AssertExpectations(); err != nil {
+		t.Error(err)
+	}
+
+	// if !reflect.DeepEqual(in, out) {
```
iantanwx: @jinzhu For some reason, when returning mock rows from sqlmock, the values are not being scanned into the struct as expected. I can't find the reason why and need some help.

iantanwx:
Closing in favour of writing as plugin

januszm:
@iantanwx could you add here a link to the plugin when it's ready?

1saifj:
6 years and still no one solved the gorm mocking issue?
is this project really considered stable?

PR #1688, opened 2017-11-22
Do not call TableName function via interface{}.

ReSTARTR:

- [x] Do only one thing
- [x] No API-breaking changes

Thank you.

If scope.Value implemts tabler interface,
DefaultTableNameHandler is never called.
But, if scope.Value is a slice of Model,
DefaultTableNameHandler is called.
DefaultTableNameHandler should always be called.

So, I removed the type assertion.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strconv"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jinzhu:
the `TableName` method is used to dynamically change current struct's table name, if you want to always call the DefaultTableNameHandler method, you could include it in the `TableName` method

Issue #1194, opened 2016-09-21
Many to many table with self referencing field


iyedg:
I have this model that represents a course, which might have many courses as prerequisites and be a prerequisite of many courses

```
type Course struct {
    gorm.Model
    Name          string
    Prerequisites []Course `gorm:"many2many:course_prerequisite"`
}
```

this code raises an error

`duplicate column name: course_id`

the logger shows that the generated SQL is as follows:

`CREATE TABLE "course_prerequisite" ("course_id" integer,"course_id" integer, PRIMARY KEY ("course_id","course_id"))`

Is there a way to rename one of the columns or any other way around this error ?

danhardman:
Seems like a very bad data structure. Obviously I'm not sure on the contract, but could `Prerequisites` be a different type? For example, `Qualification` instead of `Course`?

iyedg:
I figured it is bad data structure, but I could not come up with a logical way to model this relationship. It is a tree where all the nodes have the same type, therefore `Prerequisites` must be of type `Course`

rajan61005co:
Its definitely not a bad data structure. In database you can reference own table as many to many relationship. I am pretty surprised that this library does not address critical like this. This should have been escalated and fixed long time back. Similar issue has made me stuck as well and I definitely don't want to fix this in hacky way.

alapini:
@jinzhu

Arman92:
Any fix to this issue? I definitely need many to many with self entity!
I tried this [http://stackoverflow.com/a/28189798/1420189](url) but does not work

nathan-osman:
Good news - I think I have found a way to fix this without breaking anything else. Please see the pull request immediately above this comment.

jinzhu:
Hi All,

Thanks @nathan-osman 's works, we have this feature now, please refer http://gorm.io/associations.html#self-referencing-many-to-many-relationship for usage.

nekketsuuu:
New URL: http://gorm.io/docs/many_to_many.html#Self-Referencing

brunoro:
Is there any easy way to make symmetrical self-referencing associations? I have a `Post` model that could have many references to other duplicate `Post` entries, using a `post_duplicates` table with two columns, `post_a` and `post_b`.
[...]

PR #1709, opened 2017-12-23
Allow name of column to be customized to support self-referencing many2many fields

nathan-osman:
This pull request attempts to solve issue #1194 without breaking anything else. This is accomplished by allowing the `AssociationForeignKey` tag to specify the name for the column in the database:

    type User struct {
        //...
        Friends []*User `gorm:"many2many:user_friends;AssociationForeignKey:ID=friend_id"`
    }

Notice the addition of the `=friend_id`. This addition is optional, which allows existing code to continue to work exactly the way it did before. Models that require a self-referencing many-to-many field can use the expanded syntax.

In `join_table_handler.go`, it became clear that `getSearchMap()` wasn't going to work in `Add()` since the `s.Source.ModelType == modelType` expression would evaluate to `true` for both the source and the destination models. Therefore, I adapted `Add()` to loop through the source and destination fields individually, ensuring that the correct scope was used.

The new changes also include a test.

### DIFF
--- a/model_struct.go
+++ b/model_struct.go
@@ -289,11 +289,24 @@ func (scope *Scope) GetModelStruct() *ModelStruct {
 									}
 
 									for _, name := range associationForeignKeys {
+
+										// In order to allow self-referencing many2many tables, the name
+										// may be followed by "=" to allow renaming the column
+										parts := strings.Split(name, "=")
+										name = parts[0]
+
 										if field, ok := toScope.FieldByName(name); ok {
 											// association foreign keys (db names)
 											relationship.AssociationForeignFieldNames = append(relationship.AssociationForeignFieldNames, field.DBName)
+
+											// If a new name was provided for the field, use it
+											name = field.DBName
+											if len(parts) > 1 {
+												name = parts[1]
+											}
+
 											// join table foreign keys for association
-											joinTableDBName := ToDBName(elemType.Name()) + "_" + field.DBName
+											joinTableDBName := ToDBName(elemType.Name()) + "_" + name
 											relationship.AssociationForeignDBNames = append(relationship.AssociationForeignDBNames, joinTableDBName)
 										}
 									}
--- a/customize_column_test.go
+++ b/customize_column_test.go
@@ -279,3 +279,25 @@ func TestBelongsToWithPartialCustomizedColumn(t *testing.T) {
 		t.Errorf("should preload discount from coupon")
 	}
 }
+
+type SelfReferencingUser struct {
+	gorm.Model
+	Friends []*SelfReferencingUser `gorm:"many2many:UserFriends;AssociationForeignKey:ID=friend_id"`
+}
+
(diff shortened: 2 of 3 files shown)


model_struct.go
```diff
@@ -289,11 +289,24 @@ func (scope *Scope) GetModelStruct() *ModelStruct {
...
+
+											// If a new name was provided for the field, use it
+											name = field.DBName
+											if len(parts) > 1 {
+												name = parts[1]
+											}
+
 											// join table foreign keys for association
-											joinTableDBName := ToDBName(elemType.Name()) + "_" + field.DBName
+											joinTableDBName := ToDBName(elemType.Name()) + "_" + name
```
jchavannes: Why not just `name`?

nathan-osman: This particular line will run even if an override is not specified. If I were to only assign `name` to `joinTableDBName`, then existing code would behave differently.

jchavannes: Yes, but that can be changed.  If I set `ID=friend_id` then the column name should be `friend_id` not `user_friend_id`.

jchavannes: I'm using your branch and made this change: https://github.com/jchavannes/gorm/commit/26342e

Also, thanks for creating this branch.  Self-referencing objects definitely needed to be supported.

nathan-osman: Oh, I see what you mean. Yes, I'd be fine with doing it that way and I can easily update this pull request if that is the preferred behavior.

shenshouer:
Any update for this ?

jinzhu:
Hi @nathan-osman 

Merged it and changed the API a little bit, please refer http://gorm.io/associations.html#self-referencing-many-to-many-relationship for how to. 

Thank you for your awesome works.

jchavannes:
@jinzhu as mentioned in my comment above, I much prefer the interface used in this branch otherwise all field names will always be prepended with the table name, which isn't ideal and loses flexibility: https://github.com/jchavannes/gorm/commit/26342e

jinzhu:
@jchavannes GORM read name from tag `association_jointable_foreignkey` and won't prepend with table name. refer https://github.com/jinzhu/gorm/blob/gh-pages/documents/associations.md#self-referencing-many-to-many-relationship

jchavannes:
Unless you changed something somewhere else, if you have `ID=friend_id` the column name in the database will be named `user_friend_id`, it is automatically prefixed by the table name.  It's caused by this line:

```
joinTableDBName := ToDBName(elemType.Name()) + "_" + name
```
https://github.com/jinzhu/gorm/pull/1709/files#diff-a608ea2eb85962d34d7b6e0bf6010af9R309

I've made a change that doesn't prepend the table name (linked in my previous comment).

Perhaps you prefer the columns being prefixed by the table name, but I'd rather have the flexibility to control that.

jinzhu:
Hi @jchavannes 

Yes, I have changed the logic and its usage.

PR #1712, opened 2017-12-26
Add pgx dialect for github.com/jackc/pgx

vutung2311:
Add new dialect for library github.com/jakc/pgx which claims to be faster than github.com/lib/pq.

### DIFF
--- a/association_test.go
+++ b/association_test.go
@@ -7,7 +7,7 @@ import (
 	"sort"
 	"testing"
 
-	"github.com/jinzhu/gorm"
+	"gorm"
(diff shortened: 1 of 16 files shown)


jinzhu:
Hi @vutung2311 

Don't get why change import path from `github.com/jinzhu/gorm` to `gorm`?

vutung2311:
Hi @jinzhu 
It's because of the directory structure that I used. Correct me if I'm wrong because I think in that way tests can use the implementation regardless of directory structure.

PR #1714, opened 2017-12-27
When using postgres, don't use prepare statement, instance of compose sql manually

wmin0:
When using gorm with pgbouncer transaction mode, we face a problem.
Pgbouncer transaction mode doesn't support prepare statement, and gorm always use it.
So in this pr, I compose postgres sql instead of using prepare statement.
Therefore gorm can work fine with pgbouncer transaction mode.

In the meanwhile, I change "?" replacing logic.
Because of if the previous variable contain with "?", the next variable may be replaced in wrong position.

### DIFF
--- a/create_test.go
+++ b/create_test.go
@@ -12,7 +12,7 @@ import (
 func TestCreate(t *testing.T) {
 	float := 35.03554004971999
 	now := time.Now()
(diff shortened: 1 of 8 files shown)


jinzhu:
Hi @wmin0 

Thank you for your PR, I think the `?` replacing logic is kind of duplicated with PR https://github.com/jinzhu/gorm/pull/1655, which has been merged.

But I am not clear about the `prepare statement` issue, would you mind create another PR that only describe this issue or just ping me on gitter?

Thank you.

PR #1721, opened 2018-01-13
Implementatoin of possibility to reconnect

ruslanfedoseenko:

- [*] Do only one thing
- [] No API-breaking changes
- [*] Write good commit message, try to squash your commits into a single one

Thank you.

Implement possibility to reconnect in case of failture.
Currently dont have any time to test it - will do it later.

### DIFF
--- a/callback_create.go
+++ b/callback_create.go
@@ -50,6 +50,8 @@ func updateTimeStampForCreateCallback(scope *Scope) {
 // createCallback the callback used to insert data into database
 func createCallback(scope *Scope) {
 	if !scope.HasError() {
+		scope.db.reconnectGuard.Wait()
+
 		defer scope.trace(NowFunc())
 
 		var (
--- a/callback_delete.go
+++ b/callback_delete.go
@@ -28,6 +28,8 @@ func beforeDeleteCallback(scope *Scope) {
 // deleteCallback used to delete data from database or set deleted_at to current time (when using with soft delete)
 func deleteCallback(scope *Scope) {
 	if !scope.HasError() {
+		scope.db.reconnectGuard.Wait()
+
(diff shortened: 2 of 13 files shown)


jinzhu:
Hi @ruslanfedoseenko 

I have merged it to branch `ruslanfedoseenko-master`, and refactored it as plugins

please check it out.

https://github.com/jinzhu/gorm/tree/ruslanfedoseenko-master/plugins/reconnect

Let me know if you found any problems.

ruslanfedoseenko:
Problem I see that you need to block execution of any SQL if one thread failed. Thats why I created a WaitGroup and use it in all default call backs.  I use wait group as it allows concurent execution of normal sql in case there is now error. I f you use mutex it will blocck acces to db during 1 query.
Why you need to block execution of sql during reconnect:
Lets imagine situation - 3 Thread all failed. All thread starts reconnect and due to mutex only one will create new connection after that another  will create new connection etc.

So I think it may be done with anther before callback 

Also It is possible that one thread already starts reconnect and another just finished executeion of sql and detectred that it need reconnect

jinzhu:
>  3 Thread all failed. All thread starts reconnect and due to mutex only one will create new connection after that another will create new connection etc

No, this won't happen, if you noticed this line `connected := db.DB().Ping() == nil` after the lock, it will check current db is alive or not, and won't reconnect db if it is already connected.

> Also It is possible that one thread already starts reconnect and another just finished executeion of sql and detectred that it need reconnect

Yes, I understand, so there is a TODO left, it will rerun all failed SQL due to the bad connection issue... I will look into it later.

jinzhu:
Not going to merge it, will move it as a GORM plugin

Issue #1700, opened 2017-12-05
UpdateColumns(map[string]interface{}) generated Update sql columns should be sorted.


sunfmin:
```
UPDATE "orders" SET "shipping_address_country" = $1, "shipping_address_given_name" = $2, "shipping_address_zipcode" = $3, "shipping_address_family_name" = $4, "shipping_address_katakana_given_name" = $5, "shipping_address_city" = $6, "shipping_address_phone" = $7, "shipping_address_address2" = $8, "shipping_address_prefecture" = $9, "shipping_address_address1" = $10, "shipping_address_phone2" = $11, "shipping_address_name" = $12, "shipping_address_katakana_family_name" = $13  WHERE (order_code = $14 AND state = $15)
```

Right now it generate different sql each time.

pedromorgan:
a map is intentionally unsorted, 
- https://nathanleclaire.com/blog/2014/04/27/a-surprising-feature-of-golang-that-colored-me-impressed/
- https://stackoverflow.com/questions/12108215/golang-map-prints-out-of-order

what is our case example ?  like mine probably of having same query...for debug eyes

sunfmin:
Yes, just for sql logs to be consistent each time.

smithjessk:
@sunfmin @pedromorgan I took a stab at this with #1734. Thoughts?

sunfmin:
@smithjessk That looks great.

PR #1734, opened 2018-01-22
Sort column names before generating SQL in `DB.UpdateColumns`

smithjessk:

- [x] Do only one thing
- [x] No API-breaking changes
- [x] Run `./build.sh` in `gh-pages` branch for document changes (I couldn't get `gitbook` to install the plugins necessary. I will try again tomorrow)

Thank you.

Addresses #1700. 

I tested it by looking at the `SQL` field of the scope in `DB.UpdateColumns`. Across multiple runs, the generated SQL did not change.


smithjessk:
`build.sh` failed because I didn't have gitbook's plugins. The error message implied that, to get the plugins, I should run `gitbook install` from the current directory. However, this doesn't fix the problem; `gitbook` install must be run from `documents/`. Maybe we should put this either in the `gh-pages` README or in the PR instructions? I'd be happy to do so, and would put it in a new PR that I would tag here.

PR #1736, opened 2018-01-22
Use square brackets for quoting identifiers on MSSQL

gm42:

- [x] Do only one thing
- [x] No API-breaking changes

Use square brackets for quoting identifiers on MSSQL


gm42:
For the records, master GORM does not work with MSSQL 2012 when the table identifier is quoted, hence I made this PR.

PR #1751, opened 2018-02-02
Use table name to guess current database if none is given

ofpiyush:

If no database is selected, we try to guess based on the tableName.


ofpiyush:
@jinzhu this PR completes the work you started to support multitenancy. 

It doesn't break old behaviour, adds new feature

If you need any changes, please feel free to ask 👍

PR #1789, opened 2018-03-07
Add "gorm:query_prefix"

vutung2311:
In our project we're using pgool and sometimes we want to send query with prefix `/*NO LOAD BALANCE*/` so that pgpool will ignore load balancing and send the query directly to master node. Using callback was not possible for us to we want to have this option so do this.


jinzhu:
Hello @vutung2311 

Thank you for your PR.

I am going to support this in the coming v2 release. only accept bug fix for v1 for now.

Thank you.

pedromorgan:
@jinzhu Why not create a `next` or `develop` branch as the next upcoming version..
so
- `master` = latest stable
- `develop/next` = next upcoming version

means this could potentially be merged into upcoming v2

PR #1796, opened 2018-03-14
lock TagSettings structure when modified

posener:
The map is modified in different places in the code which results in race conditions
on execution.
This commit locks the map with read-write lock when it is modified

- [x] Do only one thing
- [x] No API-breaking changes


posener:
@jinzhu why is this in v2.0 milestone? I can't see any API break here.
Can this be merged to master?
Thanks!

mtfelian:
+1

PR #1823, opened 2018-03-30
suport date time '2070-03-30 00:00:00',timestamp can't support lar…

lotus-wu:
when i use time.Time to store '2070-03-30 00:00:00' to mysql 5.7.21, it will failed , and change to DATETIME can solve this problem.


lotus-wu:
@jinzhu   werker/build 是因为环境没起来导致的，你看看这个什么时候可以合并呢？

PR #1923, opened 2018-05-30
Wrapping field value with DBMS functions on insert/update

askretov:
**New feature:**
Allow using user-defined field value wrapper functions on insert/update queries **avoiding raw SQL**.
**_Why avoiding raw SQL?_**
If you want to minimize your code with a single db.save() operation, especially when you have a lot of nesting related objects, you can just set a wrapper tag instead of moving to raw SQL.

Usage:
```
type Object struct {
    ...
    field1 string `gorm:"wrapper:ST_GeomFromText(?)"`
    field2 string `gorm:"wrapper:UserFunc(?)"`
}
```

Use case:
1. Spatial functions. If you have geometry type fields but want to use WKT instead of WKB to write values into DB.
2. Date/string functions. If you want to perform extra actions on DBMS side (eg. extract only a part of value, transform a value etc.) to avoid in-app implementation.
3. User-defined DBMS functions to pre-process values.

### DIFF
--- a/callback_create.go
+++ b/callback_create.go
@@ -65,13 +65,13 @@ func createCallback(scope *Scope) {
 						scope.InstanceSet("gorm:blank_columns_with_default_value", blankColumnsWithDefaultValue)
 					} else if !field.IsPrimaryKey || !field.IsBlank {
 						columns = append(columns, scope.Quote(field.DBName))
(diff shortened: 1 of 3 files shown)


askretov:
Closed for improvements

PR #1924, opened 2018-05-30
Wrapping field value with DBMS functions on insert/update

askretov:
**New feature:**
Allow using user-defined field value wrapper functions on insert/update queries **avoiding raw SQL**.
**_Why avoiding raw SQL?_**
If you want to minimize your code with a single db.Save() operation, especially when you have a lot of nesting related objects, you can just set a wrapper tag instead of moving to raw SQL.

Usage:
```
type Object struct {
    ...
    field1 string `gorm:"wrapper:ST_GeomFromText(?)"`
    field2 string `gorm:"wrapper:UserFunc(?)"`
}
```

Use case:
1. Spatial functions. If you have geometry type fields but want to use WKT instead of WKB to write values into DB.
2. Date/string functions. If you want to perform extra actions on DBMS side (eg. extract only a part of a value, transform a value etc.) to avoid in-app implementation.
3. User-defined DBMS functions to pre-process values.

### DIFF
--- a/callback_create.go
+++ b/callback_create.go
@@ -65,13 +65,13 @@ func createCallback(scope *Scope) {
 						scope.InstanceSet("gorm:blank_columns_with_default_value", blankColumnsWithDefaultValue)
 					} else if !field.IsPrimaryKey || !field.IsBlank {
 						columns = append(columns, scope.Quote(field.DBName))
-						placeholders = append(placeholders, scope.AddToVars(field.Field.Interface()))
+						placeholders = append(placeholders, scope.WrapPlaceholder(field, scope.AddToVars(field.Field.Interface())))
 					}
 				} else if field.Relationship != nil && field.Relationship.Kind == "belongs_to" {
 					for _, foreignKey := range field.Relationship.ForeignDBNames {
 						if foreignField, ok := scope.FieldByName(foreignKey); ok && !scope.changeableField(foreignField) {
 							columns = append(columns, scope.Quote(foreignField.DBName))
-							placeholders = append(placeholders, scope.AddToVars(foreignField.Field.Interface()))
+							placeholders = append(placeholders, scope.WrapPlaceholder(foreignField, scope.AddToVars(foreignField.Field.Interface())))
 						}
 					}
 				}
--- a/callback_update.go
+++ b/callback_update.go
@@ -76,12 +76,12 @@ func updateCallback(scope *Scope) {
 			for _, field := range scope.Fields() {
 				if scope.changeableField(field) {
 					if !field.IsPrimaryKey && field.IsNormal {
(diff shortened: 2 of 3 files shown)


pedromorgan:
@askretov is there a test case for this, eg with sqlite (got spatial) ?

askretov:
@pedromorgan Fortunately, it doesn't matter which backend is being used, sqlite or Postgre or something else. This feature allows using any of backend-specific embedded functions or user-defined functions by adding a wrapper function explicitly within a gorm/wrapper tag for a struct's field.
As for a test case, I didn't add it within this pull request, sorry, do you want me to add a test case for this feature?

irees:
I would also be very interested in this feature.

jinzhu:
Hello, 

You could use `driver.Valuer` type and returns `gorm.Expr("dbfunc")`

askretov:
@jinzhu Assume you have a further case...
You have such object:
```
type SomeData struct {
    id int64
    field1 string
    field2 string
}
var sd SomeData
```
And you want to fetch data from a DB in the simple ORM way like:
`db.Find(&sd).Where('id = ?', 1)`

But field1 stores in DB as WKB value (it doesn't matter, it could be any other case... JSON value etc.) and you need to fetch that field using DB func ST_GeomFromText (it could be JSON_EXTRACT in case of JSON etc.).

If you use wrapper tag in the struct as I proposed, it comes as easy as the same `db.Find(&sd).Where('id = ?', 1)`.
So, could you tell how can I use gorm.Expr in this case and how it could seems like?

jinzhu:
`driver.Scanner` can do some of those work, and I think this is a better way to handle it as which would compatible with golang's `sql/database` package

askretov:
> `driver.Scanner` can do some of those work, and I think this is a better way to handle it as which would compatible with golang's `sql/database` package

Unfortunately, that's definitely another way to solve such a case. If you try to use drive.Scanner you need to implement your own client logic to convert WKB to string, but it's already implemented by DBMS.

But the much more bad thing in this way is that you have huge traffic overhead... assume you have JSON value in DB... big object enough... let's assume this JSON value in DB has 20 props, but your struct needs the only one of them. First, you need to implement your own client logic to extract JSON field and it's really bad... Second, you need to fetch the full raw value from a DB with 20 props and then extract the one only locally... why?

Why don't you want to pull wrapper tag feature? It solves tons of cases in the right and efficient way. More than this, I tried to implement it in your own architecture style so it seems absolutely natural in the project.

t2wu:
@jinzhu I finally understand what you meant by a valuer. However, It didn't seem to work with MySQL since the dbfunc is quoted. I posted the question [here](https://stackoverflow.com/questions/60520863/working-with-spatial-data-with-gorm-and-mysql/60537524#60537524) on Stackoverflow. Anything I can do to escape that quote? Or is it really not within Gorm's control?

PR #1926, opened 2018-05-31
Support only preloading 

gstvg:
Add function to only preload an already populated/initialized model. Very small and simple change. Using to preload simple fields of an model with custom sql on a performance-critical context.

### DIFF
--- a/callback_query.go
+++ b/callback_query.go
@@ -19,6 +19,10 @@ func queryCallback(scope *Scope) {
 		return
 	}
 
+	if _, skip := scope.InstanceGet("gorm:only_preload"); skip {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


jinzhu:
Hello @gstvg 

Could you let me know your use case?  and can the `Related` method help you?

gstvg:
Hello @jinzhu, thanks for replying.
Use it for querying products on category page. Its a custom query working on a slice of Products.
So, i couldn't use `Related` since it would require a loop with 1 query for every item, right? Using this pull, i can use performance optimizations from Preload. Thanks again!

jinzhu:
Thank you for your explanation, that makes sense.

PR #1939, opened 2018-06-08
Edit DB.clone(), DB.Dialect(), and Scope.Dialect() preserve transactions

cavemanpi:
#### The problem I'm trying to solve

When running a migration in a transaction, the DB.AutoMigrate() function queries the database for the existence of tables outside of the context of the transaction. This can result in the attempt to create a table which already exists in the transaction. For instance, the following migration results in a failure.

``` go
package models

import (
	"time"

	"github.com/jinzhu/gorm"
	gormigrate "gopkg.in/gormigrate.v1"
)

func Migrate() error {
	db, openErr := gorm.Open("postgres", "host=127.0.0.1 port=5432 user=ingress dbname=ingress password=password sslmode=disable")

	if openErr != nil {
		return openErr
	}

	options := &gormigrate.Options{
		TableName:      "migrations",
		IDColumnName:   "id",
		UseTransaction: true,
	}

	if db.Error != nil {
		return db.Error
	}

	m := gormigrate.New(db, options, []*gormigrate.Migration{
		{
			ID: "initial",
			Migrate: func(tx *gorm.DB) error {
				type JsonifiedModel struct {
					ID        uint       `json:"id" gorm:"primary_key"`
					CreatedAt time.Time  `json:"created_at"`
					UpdatedAt time.Time  `json:"updated_at"`
					DeletedAt *time.Time `json:"deleted_at" sql:"index"`
				}

				type First struct {
					JsonifiedModel
					Name string `json:"name" gorm:"not null"`
				}

				err := tx.CreateTable(&First{}).Error
				if err != nil {
					return err
				}

				return nil
			},
		},
		{
			ID: "add_ttl",
			Migrate: func(tx *gorm.DB) error {
				// when table already exists, it just adds fields as columns
				type First struct {
					TTL int `gorm:"not null"`
				}

				return tx.AutoMigrate(&First{}).Error
			},
			Rollback: func(tx *gorm.DB) error {
				err := tx.Table("firsts").DropColumn("ttl").Error
				if err != nil {
					return err
				}

				return nil
			},
		},
	})

	err := m.Migrate()

	if err != nil {
		return err
	}

	return nil
}
```

The migration above results in an error: Could not perform database migrations: pq: relation "firsts" already exists.

#### Proposed Fix

When a DB connection is cloned, I propose that it stores a Dialect based off the parent's dialect, but using the db SQLCommon of the connection it was cloned from. This will make the current transaction available to a newly created scope so that when it interrogates the DB for the existence of a table, it has the context of the current running transaction.

Additionally, when a transaction is begun, I think it should update the DB connection of the dialect stored in the newly cloned connection.

Feedback is very welcome.


cavemanpi:
Have you had a chance to look at this PR and if so, do you have any feedback or needs from me on this PR?

PR #1958, opened 2018-06-19
Fix Race condition in tablename

jontonsoup:
Fixes https://github.com/jinzhu/gorm/issues/1407

### DIFF
--- a/model_struct.go
+++ b/model_struct.go
@@ -17,6 +17,8 @@ var DefaultTableNameHandler = func(db *DB, defaultTableName string) string {
 	return defaultTableName
 }
 
+var mutex sync.Mutex
+
 type safeModelStructsMap struct {
 	m map[reflect.Type]*ModelStruct
 	l *sync.RWMutex
@@ -50,6 +52,9 @@ type ModelStruct struct {
 
 // TableName get model's table name
 func (s *ModelStruct) TableName(db *DB) string {
+	mutex.Lock()
+	defer mutex.Unlock()
+	
 	if s.defaultTableName == "" && db != nil && s.ModelType != nil {
 		// Set default table name
 		if tabler, ok := reflect.New(s.ModelType).Interface().(tabler); ok {


model_struct.go
```diff
@@ -17,6 +17,8 @@ var DefaultTableNameHandler = func(db *DB, defaultTableName string) string {
 	return defaultTableName
 }
 
+var mutex sync.Mutex
```
mtfelian: Why you added this to the package and not into ModelStruct type?

mtfelian:
I issued another [PR](https://github.com/jinzhu/gorm/pull/1967) with mutex in the proper scope.

model_struct.go
jontonsoup: @mtfelian doesn't this depend on what context the go routines are called in? If I want a global mutex but the scope is inside the struct it might not actually work as intended. 

That said, I am not familiar with this code base and used the 'nuclear' option to be safe. I was thinking about this PR as a discussion point more than an actual PR. The state of the linked bug was "can't reproduce" but the issue clearly exists. This PR fixes the bug, but there are other concerns I have with this PR in general: how this might impact performance and if there are any deadlock scenarios. 

Thanks!

jontonsoup: maybe @jinzhu can chime in here

mtfelian: @jontonsoup think this should not have a visible affect on performance due to no "hard" methods are inside. To make it faster, you can set/get the `defaultTableName` field with mutex. 

I tested my fork (i hurried to make a PR, but the Mutex there is inside the struct) with our production code with transactions, multithreading and other things and it fixes the race. 

So i think you can move it inside the struct: 

```
// ModelStruct model definition
type ModelStruct struct {
	PrimaryFields    []*StructField
	StructFields     []*StructField
	ModelType        reflect.Type
	defaultTableName string

	l sync.Mutex
}
```

jinzhu: I think we could just use sync.Map to avoid the race condition

mtfelian: @jinzhu i created another PR https://github.com/jinzhu/gorm/pull/2022 where it's changed to sync.Map

jinzhu:
fixed with https://github.com/jinzhu/gorm/pull/2022

dnathe4th:
@jinzhu #2022 does not resolve this race condition within `TableName` solved for in this PR (and #1958). The race condition resolved by this PR is due to the reading and writing of the `defaultTableName` attribute on the `ModelStruct`. #2202 provides no protections around this. Please consider opening and merging this PR in to resolve the issue of concurrent reads to `TableName` with an unset `defaultTableName`

PR #1967, opened 2018-06-29
fix race condition in ModelStruct.TableName()

mtfelian:
- [x] Do only one thing
- [x] No API-breaking changes

Fix race condition in `ModelStruct.TableName()`: https://github.com/jinzhu/gorm/issues/1407


mtfelian:
sorry, i was too fast

PR #1988, opened 2018-07-16
Fix issue updating models with foreign key constraints

fillup:

- [x] Do only one thing
- [x] No API-breaking changes

We found an issue where Gorm was trying to write the zero value of a has_one foreign key into the table and when using actual database foreign keys, the value `0` was invalid. This pull request adds a check in the `updateCallback` to ensure the field being update is not a blank foreign key with a default value. In our case the has_one relationship has a default value of null in the event that the model does not have the child yet. 

All gorm tests pass and this makes it safe to use foreign keys with a default value of null.

### DIFF
--- a/callback_update.go
+++ b/callback_update.go
@@ -76,7 +76,9 @@ func updateCallback(scope *Scope) {
 			for _, field := range scope.Fields() {
 				if scope.changeableField(field) {
 					if !field.IsPrimaryKey && field.IsNormal {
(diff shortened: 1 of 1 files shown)


jinzhu:
thank you for your PR.

Scott-Chen:
Is it possible to allow update NULL value to foreign key?  It was work before this.

PR #1995, opened 2018-07-19
support association Find and Count with customized func, change buildCondition TableName to struct's TableName  when query is interface{}

hectorqin:

- [-] Do only one thing
- [+] No API-breaking changes
- [+] New code/logic commented & tested

* support association Find and Count with customized function
* change buildCondition TableName to struct's TableName  when query is interface{}

### DIFF
--- a/association.go
+++ b/association.go
@@ -15,8 +15,9 @@ type Association struct {
 }
 
 // Find find out all related associations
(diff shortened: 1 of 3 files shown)


jinzhu:
We don't need to change the public API to accepts the `Scopes`, we can make it like:

`db.Scopes(funcs). Association("User").Count()`

PR #2021, opened 2018-08-03
Fixes nested-preload for rows with nil pointers

brunoro:
The `IsValid` sanity check in the `getValueFromFields` would return `true` even if `fieldValue` was nil. Wrapping `fieldValue` in a `reflect.Indirect` fixes the issue.


jinzhu:
thank you for your PR.

sulthonzh:
I have problem with this condition, 

mysql, with UUID

PR #2040, opened 2018-08-16
Naming Strategy

ademozay:
Instead of tagging every single struct field, this PR is useful when you want to add prefix/suffix or alter the database/table/column names. 
Default naming strategy(snake case) is being applied without any configuration but if you like to set custom naming strategy you can use below 

```go
func main() {
	db, err := gorm.Open("sqlite3", "test.db")
	if err != nil {
	}
	defer db.Close()

	gorm.AddNamingStrategy(&gorm.NamingStrategy{
		DB: func(name string) string {
			return "db_" + name
		},
		Table: func(name string) string {
			return "table_" + name
		},
		Column: func(name string) string {
			return "col_" + name
		},
	})

	db.AutoMigrate(&Product{})
}
```


pedromorgan:
This look great, and what stopped my using GORM on a recent project

jinzhu:
Looks great, thank you.

fieryorc:
Thanks!

PR #2049, opened 2018-08-22
Fix gorm:"-" When the query is not working to  issues 2048

aixiaoxiang:
Fix https://github.com/jinzhu/gorm/issues/2048

Fix Fix gorm:"-" When the query is not working

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -112,6 +112,9 @@ func (scope *Scope) Fields() []*Field {
 		)
 
 		for _, structField := range scope.GetModelStruct().StructFields {
+			if structField.IsIgnored {
+				continue
+			}
 			if isStruct {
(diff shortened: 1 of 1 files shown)


pedromorgan:
@jinzhu mayve this should be merged as its a bug.

jinzhu:
Hi, fixed that issue with commit https://github.com/jinzhu/gorm/commit/dc3b2476c4eb61c37424a1ca2f46859e4e6fcd81

We don't want to ignore those ignored fields in the `Fields` method as we still would like to try to fill those values into our models, it would be helpful in some cases.

Thank you for your PR.

aixiaoxiang:
@jinzhu When the query use `gorm:"column:-`  ignored fields.

PR #2058, opened 2018-08-27
Improve preload speed

### LINKED ISSUE #1672: Improve preload speed
When doing HasOne and BelongsTo Preloads (and maybe others also), GORM is quite slow when the there are many query results and many referenced results.

Example scenario:
- Table User and table Profile, where User "belongs to" Profile (i.e. User has a ProfileID FK referencing a column ID in Profile)
- Query all users in a table with Preload("Profile")
- If there is large number of users (and large number of profiles), GORM is slow

I've looked at the code, and found that handleBelongsToPreload and handleHasOnePreload in callback_query_preload.go are inefficient when assigning results of relation queries. 
Currently, ForeignField values from one set of results and AssociatedForeignField values from the other set are compared as strings in a nested for loop. This is quadratic complexity, which is quite bad if there is large number of rows (more than a few thousand).

This can be improved by storing field values of one set of results in a map with a string key, iterating through the field values/objects of the other set and looking for the same values in the map.

maddie:
All credits to @vanjapt who came up with this patch.

Closes #1672

It would be great if @vanjapt can help improve this PR a little bit (adding comments etc.) since I don't 100% understand the code here, but I've tested locally with production data and it seems to work correctly and sped up the `Preload`s from 7~8s to around 2s

### DIFF
--- a/callback_query_preload.go
+++ b/callback_query_preload.go
@@ -161,14 +161,17 @@ func (scope *Scope) handleHasOnePreload(field *Field, conditions []interface{}) 
 	)
 
 	if indirectScopeValue.Kind() == reflect.Slice {
+		foreignValuesToResults := make(map[string]reflect.Value)
+		for i := 0; i < resultsValue.Len(); i++ {
+			result := resultsValue.Index(i)
+			foreignValues := toString(getValueFromFields(result, relation.ForeignFieldNames))
+			foreignValuesToResults[foreignValues] = result
+		}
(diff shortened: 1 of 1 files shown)


andrea3bianchi:
A similar implementation to this brought down queries on many objects(20k+) with many preloads(20+) from minutes to seconds for us also!

jinzhu:
Thank you for your PR.

PR #2064, opened 2018-08-29
Use sync.Map for DB.values

akhleung:

- [x] Do only one thing
- [x] No API-breaking changes

When using Gorm in a production environment, we experienced fatal errors caused by concurrent map reads/writes. We traced this down to `DB.values`, and when we put a mutex around this map, the concurrent reads/writes stopped. However, we noticed that other parts of Gorm seem to prefer `sync.Map`, so this pull request replaces `DB.values` with a `sync.Map`.

(Relevant issue: https://github.com/jinzhu/gorm/issues/2065)

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,7 +22,7 @@ type DB struct {
 	logMode           int
 	logger            logger
 	search            *search
-	values            map[string]interface{}
+	values sync.Map
 
 	// global db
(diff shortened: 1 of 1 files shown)


jinzhu:
Thank you four your PR.

PR #2080, opened 2018-09-04
Fix scanning identical column names occurring >2 times

xyziemba:

- [x] Do only one thing
- [x] No API-breaking changes

Fix a bug in `Scope.scan` where fields with repeated names are not incorrectly skipped. Specifically, values in `selectedColumns` must be indexed relative to `fields`, not `selectFields`.

Add a test demonstrating how a SELECT query with three ID fields from a joined table can be used to populate a single struct with the correct values. (The test fails without this fix.)

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -581,6 +581,60 @@ func TestJoins(t *testing.T) {
 	}
 }
 
+type JoinedIds struct {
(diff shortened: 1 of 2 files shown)


jinzhu:
thank you for your PR.

PR #2096, opened 2018-09-18
Avoid conflict between variable and package names

fifsky:

- [x] Do only one thing
- [x] No API-breaking changes

Avoid conflict between variable and package names


jinzhu:
Close due to failed to pass tests after any tries.

PR #2098, opened 2018-09-19
adding function to get column's value

geedchin:
adding function to get column's value


jinzhu:
I think it is not necessary, you could use Scope to get field by name, and get its value

PR #2099, opened 2018-09-19
adding function to get column's value

geedchin:
adding function to get column's value

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -238,6 +238,43 @@ func (scope *Scope) SetColumn(column interface{}, value interface{}) error {
 	return errors.New("could not convert column to field")
 }
 
+// GetColumn to get the column's value, column could be field or dbname
+func (scope *Scope) GetColumn(column interface{}) (interface{}, error) {
+
+	return scope.getColumn(column, false)
+}
+
+// GetColumnByModelName to get column's value,column should be filed's name
+func (scope *Scope) GetColumnByModelName(column string) (interface{}, error) {
+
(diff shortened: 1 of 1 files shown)


scope.go
```diff
@@ -238,6 +238,43 @@ func (scope *Scope) SetColumn(column interface{}, value interface{}) error {
...
+// getColumn to get the column's value, column could be field or field's name/dbname
+func (scope *Scope) getColumn(column interface{}, isModelName bool) (v interface{}, e error) {
+
+	if field, ok := column.(*Field); ok {
+
+		v = field.Field.Interface()
+		return
+	}
+
+	for _, field := range *scope.fields {
```
geedchin: here should be 
```go
for _, field := range *scope.fields {
```

geedchin:
when we need to add a prefix or change it ,we can use these function to get the value.
## for example
i want to add a 'prefix_' on name column :
```go
type User struct {
	ID      string `gorm:"primary_key" json:"id"`
	Account string `json:"account"`
	Name    string `json:"name"`
}

func (u *User) BeforeCreate(scope *gorm.Scope) error {

	// add prefix
	name, err := scope.GetColumn("name")
	if err != nil {
		return err
	}
	name = "prefix_" + name.(string)
	scope.SetColumn("name", name)
	return nil
}
```

jinzhu:
field, ok := scope.FieldByName("FieldName")
field.Field.Interface()

PR #2126, opened 2018-10-09
Add RollbackUnlessCommitted()

AdamSLevy:

- [x] Do only one thing
- [x] No API-breaking changes

This adds a function called `RollbackUnlessCommitted()` which is intended to be `defer`ed at the beginning of a transaction to avoid needing to call Rollback in every error case. `RollbackUnlessCommitted`  calls sql.Tx.Rollback() and ignores `sql.ErrTxDone` errors, indicating that the transaction was already committed.

Below is an example of its intended usage. In the example only exiting the function through the final return will result in a commit. Any other return will result in a rollback. 

```golang
func doTransaction(DB *gorm.DB) error {
        tx := DB.Begin()
        defer tx.RollbackUnlessCommitted()

        u := User{Name: "test"}
        if err != tx.Save(&User).Error; err != nil {
                return err
        }
        return tx.Commit().Error
}
```


rubensayshi:
@jinzhu this seems like a no brainer to merge and would be very welcome to see in the next release :)

hewenyang:
The function name seems a bit long. I believe it would suffice to make Rollback a no-op when Commit() has been successful, like [Badger](https://godoc.org/github.com/dgraph-io/badger#Txn.Discard) does.

pwli0755:
What's the difference between `defer tx.RollbackUnlessCommitted()` and `defer tx.Rollback()`?

rubensayshi:
`defer tx.Rollback()` will log an error if you called `tx.Commit()` anywhere before the defer occurs.  
it doesn't panic but just does a `log.Error` so it was never harmful, but it was anoying and look worrying for many devs. 

but other than the log line the `defer tx.Rollback()` won't have any effect if you did `tx.Commit()` before the defer occurs.

PR #2142, opened 2018-10-22
add an override on the DB instance instead of using the global NowFunc.

rubensayshi:
- [x] Do only one thing
- [x] No API-breaking changes

when using multiple instances of `gorm.DB` inside a testsuite that runs in parallel model (using `t.Parallel`) it becomes impossible to use the global `NowFunc` that is in `util.go`, by moving it onto the `gorm.DB` instance it becomes much easier to control because it won't affect all instances of `gorm.DB`.

the old `NowFunc` in `util.go` still remains for backwards compatibility and because most likely for most developers they actually like it that they can control it globally.


pedromorgan:
This looks cool, aint run it, but makes sense.. much like olde mysql bug and calling now() function causes issues

jinzhu:
Hello @rubensayshi 

Could you let me know the reason you need different `NowFunc` in one application, thank you.

rubensayshi:
we run tests in parallel (using the `t.Parallel()` and `-parallel=8` of `go test`).  
during the tests we mock the time to be able to properly test code that depends on time (using https://github.com/jonboulle/clockwork), so in some test cases we move the time forward and having a global `NowFunc` results in these parallel tests sometimes messing with each-others results.

the `NowFunc` is the only thing in gorm that actually is still a global, everything else works perfectly in parallel with multiple `gorm.DB` instances.

PR #2143, opened 2018-10-23
Fixed: rows affected when get row

cuiweixie:
https://github.com/jinzhu/gorm/issues/2136


cuiweixie:
@jinzhu  what do you think about it?

jinzhu:
I think the RowsAffected is not necessary for the row... maybe better to remove it from log for `Row`?

PR #2155, opened 2018-10-27
Check valuer interface before scan value

nanjj:

- [x] Do only one thing
- [x] No API-breaking changes

convert via valuer interface before do scan.
Scan interface only accept int64, float64, bool, []byte, string,
time.Time or nil. When do scan, it's better to check whether the type
support valuer interface and do convert.

### DIFF
--- a/field.go
+++ b/field.go
@@ -2,6 +2,7 @@ package gorm
 
 import (
 	"database/sql"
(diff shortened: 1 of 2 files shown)


Tsuki:
This change affecting the where cause using struct 
#2245

PR #2156, opened 2018-10-29
Argument that can be set on DB structure

craigmj:

- [X] Do only one thing
- [X] No API-breaking changes

This PR adds a `SetArg(interface{})` and a `Arg() interface{}` method to a DB. The argument is a user-settable value that can be retrieved from the DB in the various callbacks, allowing a user to extend the DB class in their own code, and then access the extended class in the callbacks.

#### Rationale
I extend the *DB and create a transaction for each web request. I then do request-scoped caching for the request: for example, I will only use the ORM to lookup a user once during the request, and cache that user result: I do not want the cache globally scoped, because of cache eviction issues. However, in the various callbacks (AfterFind especially), I need access to the request-scoped cache, so scoping the cache to the transaction (the *DB) works perfectly. EXCEPT, when I am inside a callback, I no longer have access to the extended *DB class that contains my cache.
Using the `SetArg(interface{})` and `Arg() interface{}` methods, I can link my own class to the `*DB` and then retrieve this class inside the callbacks.

### DIFF
--- a/main.go
+++ b/main.go
@@ -30,17 +30,56 @@ type DB struct {
 	dialect       Dialect
 	singularTable bool
 
-	// Arguments you can add to a db
+	// Argument you can add to a db
 	arg interface{}
 }
 
 // SetArg allows you to set an application defined argument to
-// a DB object.
+// a DB object. You can then access this argument in the
+// various callback functions.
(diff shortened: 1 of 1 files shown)


jinzhu:
you can use `db.Set`, `db.Get`

PR #2161, opened 2018-11-02
Log adds context features

e421083458:
Log adds context features

If I want to set up a separate context context for a HTTP session request, You can use this feature.

### Testing steps
1. export DEBUG=true
2. export LOGCTX=true 
3. go test -v -run "TestLogger"


jinzhu:
I think it is not necessary to do that?

e421083458:
Well, the coupling is too serious. I want to consider printing SQL into the context of the request. Is there a way to get the query SQL now?

e421083458:
> I think it is not necessary to do that?

Well, the coupling is too serious. I want to consider printing SQL into the context of the request. Is there a way to get the query SQL now?

PR #2164, opened 2018-11-04
LintFix: Make receiver name of structField consistent

crispgm:

- [x] Do only one thing
- [x] No API-breaking changes

Fix:
```
receiver name structField should be consistent with previous receiver name s for StructField
```


crispgm:
@pedromorgan ok, all the receivers are `s` by now. I have to change them all.

Issue #2176, opened 2018-11-12
New feature 'CopyIn for postgres'  


fwhezfwhez:
New feature:
Allow postgres dialect using CopyIn function to insert a big amount of data in the same time.

Usage:
```go
// CopyIn execute a big amount of insertions
//   closeAfterUsed: After finished or stopped by accident, the copyDB close itself  if 'closeAfterUsed' has been set true
//   table: which table you want to operate
//   args: what you want to insert
//   columns: what column you want to operate
// For instance:
//   db.CopyIn(false, "user", [][]string{["tom", 9],["sara", 10],["jim", 19]}, "name","age")
// This stands for 'insert into user(name,age) values('tom', 9),('sara',10),('jim',19)'
// 				or 'COPY 'user' ('name','age') FROM STDIN`'
func (s *DB) CopyIn(closeAfterUsed bool,table string, args [][]interface{}, columns ...string) error {
	if s.Dialect().GetName() != "postgres" {
		return errors.New("CopyIn only supports postgres")
	}

	pdb, err := s.CopyDB()
	if closeAfterUsed{
		defer pdb.Close()
	}

	if err != nil {
		return err
	}
	txn, err := pdb.Begin()
	if err != nil {
		return err
	}

	stmt, err := txn.Prepare(pq.CopyIn(table, columns...))
	if err != nil {
		return err
	}
	for _,v:= range args {
		_, err = stmt.Exec(v...)
		if err != nil {
			txn.Rollback()
			return err
		}
	}

	_, err = stmt.Exec()
	if err != nil {
		txn.Rollback()
		return err
	}

	err = stmt.Close()
	if err != nil {
		txn.Rollback()
		return err
	}
	err = txn.Commit()
	if err != nil {
		txn.Rollback()
		return err
	}
	return nil
}
```
Use case:
```sql
-- postgres
create table if not exists example(
name varchar,
age integer
)
```
```go
...
	source := fmt.Sprintf("host=%s user=%s dbname=%s sslmode=%s password=%s",
		"localhost", "postgres", "test", "disable", "123")
	db, er := gorm.Open("postgres", source)
	if er != nil {
		t.Fatal(er.Error())
	}
        	var args = make([][]interface{}, 0)
	args = append(args, []interface{}{
		"tom", 9,
	}, []interface{}{
		"sara", 10,
	}, []interface{}{
		"jim", 11,
	})
       e := db.CopyIn(true, "example", args, "name", "age")
...
```

Test case:
```go
[...]

PR #2178, opened 2018-11-12
New feature, add CopyIn for postgres dialect

fwhezfwhez:
Refer:
#2176 
New feature:
Allow postgres dialect using CopyIn function to insert a big amount of data in the same time.
Added apis:
```go
// DataSource returns its opened dataSource:
//    gorm.Open("postgres", "host=localhost user=root dbname=test sslmode=disable password=123")
// Then dataSource = "host=localhost user=root dbname=test sslmode=disable password=123"
func (s *DB) DataSource() string

// CopyDB returns its origin db engine whose driver is github.com/lib/pq
func (s *DB) CopyDB() (*sql.DB, error)

// CopyIn for postgres
func (s *DB) CopyIn(closeAfterUsed bool, table string, args [][]interface{}, columns ...string) error
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 documents
 _book
 .idea/workspace.xml
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 documents
 _book
+.idea/workspace.xml
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"github.com/lib/pq"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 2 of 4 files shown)


fwhezfwhez:
hello ,when adopted,I'll add CopyIn for struct like
`DB.Model(&Example{}).CopyIn2([]Example{example1, example2 ...})`

fwhezfwhez:
I used to think adding CopyIn into *gorm.DB as receiver,however find the db instance in the 'gorm.Open()' returns a db interface.I realize it's bad to modify the interface rule itself, and also it's ugly to use db interface{} assertion.Finally decide to put a CopyDB which uses gorm's origin github/lib/pq driver.Basic postgres copy in feature has been done and it shall be judged by users.

PR #2219, opened 2018-12-11
Custom Type Encoding

rodriguise:

- [X] Do only one thing
- [X] No API-breaking changes

This pull request add a custom field encoder for structs that allows for handling field encoding manually. This is similar to using sql.Scanner/driver.Valuer, but provides more flexibility when the types are more complex or external to the source (i.e. from another library). 

The example uses simple json types, but this was primarily written to support even more complex types like binary geometry data objects (EWKB, etc) in platforms like Postgis. 

This also allows for fields that have `interface` types, supporting more complex polymorphism. We use this with Swagger (OpenAPI) auto-generated models that use the `discriminator` tag, which results in fields with an interface{} type. 

The containing struct needs to implement the simple interface: 

```
// Encoder is a value encoding interface for complex field types
type Encoder interface {
	EncodeField(column string) (interface{}, error)
	DecodeField(column string, value interface{}) error
}
```

The example in ```encoder_test.go``` is pretty self explanatory.

The code ensures that NormalFields are populated prior to calling the interface, so other fields can be utilized in the decoding of the interface.


jinzhu:
Hello, this should be not necessary, refer https://github.com/qor/serializable_meta

rodriguise:
I disagree, the qor method is far less flexible and is a bit of a gorm anti pattern and introduces an additional dependency.

PR #2227, opened 2018-12-18
Add BeginTx for parity with sql.DB.BeginTx

### LINKED ISSUE #1870: Begin with txOptions
It would be nice to be able to use the BeginTx variant introduced in Go 1.8. It allows to specify transaction isolation levels and read-only/read-write behavior together with a BEGIN instruction, at least in Postgres. One can do a subsequent SET TRANSACTION query but it means one extra roundtrip per transaction.

tylerb:

- [x] Do only one thing
- [x] No API-breaking changes

This PR introduces `DB.BeginTx(context.Context, sql.TxOptions)` for parity with the `sql` package. This call is necessary to enable isolation levels and readonly transactions.

Fixes #1870


tylerb:
One thing I wasn't sure about is whether or not we want to wrap the `sql` specific types in order to avoid needing to import the `sql` package directly in user-programmer code.

For example, right now the method expects an `sql.TxOptions`. Additionally, the isolation levels are defined inside the `sql` package.

We could create our own versions of both of those and translate to the `sql` types internally.

PR #2279, opened 2019-02-01
go modules

istsh:

- [x] Do only one thing
- [x] No API-breaking changes

corresponded to go modules

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/istsh/gorm
+
+require github.com/jinzhu/inflection v0.0.0-20180308033659-04140366298a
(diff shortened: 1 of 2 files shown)


konradreiche:
Without this change it seems gorm cannot easily be added using Go Modules.

inliquid:
GORM works with modules just fine, without this change.

jonathaningram:
@jinzhu it looks like the module import path you committed is not `jinzhu` rather it's `istsh`? Was that intentional?

See https://github.com/jinzhu/gorm/blob/4b13e079fcea637fcb166ee1752c8d80601e3ef0/go.mod#L1

```
module github.com/istsh/gorm
```

Also, the go.mod and go.sum in my PR #2307 was generated automatically by Go and so it includes quite a lot more than what you merged into master. Is there any reason for only including `github.com/jinzhu/inflection` in `go.mod`?

cc @istsh

inliquid:
That's mistake. Please revert this commit @jinzhu

PR #2298, opened 2019-02-19
create table add column comment

itcuihao:

- [x] Do only one thing
- [x] No API-breaking changes

for example:

```
package main

import (
	c "tests/gorm/comment"
)

func main() {
	db := c.InitTestDB()
	db.AutoMigrate(&TableComment{})
}

type TableComment struct {
	Name string `sql:"type:varchar(10);not null;default:'';comment:'姓名'"`
}
```

result:
```
CREATE TABLE `table_comments` (`name` varchar(10) NOT NULL  DEFAULT '' COMMENT '姓名' )
```


dragonflylee:
This commit doesn't work with PostgresSQL

quexer:
> This commit doesn't work with PostgresSQL

hope this patch could be merged.

Issue #2326, opened 2019-03-12
Many2Many syntax error


Clowning:
GoVersion : 1.12
Database : PostgreSQL 11.2

Hi, i'm trying to make a many2many relation but I've got this error ->
```pq: relation "'offers_operating_systems'" does not exist``` 

And it's true, this table doesn't exist!

But in my (offer)model i set it to 
```gorm:"many2many:offers_operatingSystems;foreignkey:id;association_foreignkey:id;association_jointable_foreignkey:operatingSystemsId;jointable_foreignkey:offersId;" ```

I guess this issue is because I use a capital for 'Systems' but I can't change my database.

So, can you know have to fix this table name? 

Thanks

emirb:
Fixed in https://github.com/jinzhu/gorm/pull/2331

PR #2331, opened 2019-03-14
fix the table name of many2many

### LINKED ISSUE #2322: version 1.9.2  manytomany:tableName not correct
I define a many to many relation:
type Xxx struct {
UserAlerts      []*UserAlert      `gorm:"many2many:userAlert;association_autoupdate:false"`
}
type Yyy struct {
        Name                   string
	WebhookAlerts    []*Xxx`gorm:"many2many:userAlert"`
}
then the code 
<pre> 
db.Preload("Xxx").Where(&Yyy{Name: name}).First(u)
</pre>
in version 1.9.1 
<pre>
SELECT * FROM `Xxx` INNER JOIN `userAlert` ....
</pre>
but in version 1.9.2
<pre>
SELECT * FROM `Xxx` INNER JOIN `user_alert` ....
</pre>

The tables userAlert and user_alert is auto create, how to keep use userAlert?

xwjdsh:

- [x] Do only one thing
- [x] No API-breaking changes

fixes #2322 #2326  
these two issues are the same.

### DIFF
--- a/customize_column_test.go
+++ b/customize_column_test.go
@@ -289,6 +289,9 @@ type SelfReferencingUser struct {
 func TestSelfReferencingMany2ManyColumn(t *testing.T) {
 	DB.DropTable(&SelfReferencingUser{}, "UserFriends")
 	DB.AutoMigrate(&SelfReferencingUser{})
+	if !DB.HasTable("UserFriends") {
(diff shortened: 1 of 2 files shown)


emirb:
Thanks @xwjdsh.

PR #2411, opened 2019-04-19
Remove old elements from the output parameter of Pluck()

zenovich:

- [x] Do only one thing
- [x] No API-breaking changes

Currently, the Pluck() method keeps old elements in the output parameter. So, if someone calls Pluck() twice, the results would be doubled. The patch fixes this incorrect behaviour.


emirb:
Thanks @zenovich.

PR #2434, opened 2019-05-01
Don't AddError for Rollback on ErrTxDone

j16r:
When db.Rollback fails, check the error code. If it matches sql.ErrTxDone, we won't treat it as an error. This is expected and normal behavior of using Rollback with defer.

e.g: 

```go
tx := db.Begin()
defer tx.Rollback()
...
return tx.Commit().Error
```

See more discussion here: https://github.com/go-gormigrate/gormigrate/issues/36


j16r:
@javie2007 what do you think it would take to get this merged? The coverage could be a little tricky, not sure how to trigger an appropriate error in Rollback.

andreynering:
Ping @jinzhu @emirb

PR #2435, opened 2019-05-01
Pass logger into Callback{} so that logs are printed consistently

### LINKED ISSUE #1825: enhancement: GetLogger() for callbacks
Callbacks need to log into the same logger as gorm. 

If callbacks are in a different package (outside gorm), then,
they need a gorm.GetLogger() to log using the same logger as gorm.

j16r:
Pass the db's logger into Callback{} so that the same logger is used for callback registration and removal.

[Fixes #1825]


emirb:
Thanks @j16r!

PR #2619, opened 2019-08-16
add a field(LastInsertId) in scope to getLastInsertId

wushuangxiaoyu:
- [x] Do only one thing
- [x] No API-breaking changes

add a filed to solve cannot get lastInsertid when insert by Exec(“insert into...”) 

add a field: LastInsertId  appointed by result.LastInsertId()


emirb:
Tests failing: 
`No error should happen when create table, but got LastInsertId is not supported by this driver`

PR #2635, opened 2019-08-28
fix gorm logger insert timezero 0001-01-01

itcuihao:

- [x] Do only one thing
- [x] No API-breaking changes

```
logger:

[sql D:/mygo/src/ch/t/run.go:309 1.9954ms INSERT  INTO `a` (`user_id`,`vip_expire_time`,`message_type`,`click_tab_time`) VALUES (?,?,?,?) [0 0001-01-01 00:00:00 +0000 UTC  0001-01-01 00:00:00 +0000 UTC] 1]

?[33m[2019-08-28 10:53:01]?[0m  ?[36;1m[0.99ms]?[0m  INSERT  INTO `a` (`user_id`,`vip_expire_time`,`message_type`,`click_tab_time`) VALUES (3,'0001-01-01 00:00:00','hello','0001-01-01 00:00:00')
```

check timezero:
```
INSERT  INTO `a` (`user_id`,`vip_expire_time`,`message_type`,`click_tab_time`) VALUES (3,'0000-00-00 00:00:00','hello','0000-00-00 00:00:00')
```

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -49,7 +49,11 @@ var LogFormatter = func(values ...interface{}) (messages []interface{}) {
 				if indirectValue.IsValid() {
 					value = indirectValue.Interface()
 					if t, ok := value.(time.Time); ok {
-						formattedValues = append(formattedValues, fmt.Sprintf("'%v'", t.Format("2006-01-02 15:04:05")))
+						if t.IsZero() {
+							formattedValues = append(formattedValues, fmt.Sprintf("'%v'", "0000-00-00 00:00:00"))
+						} else {
+							formattedValues = append(formattedValues, fmt.Sprintf("'%v'", t.Format("2006-01-02 15:04:05")))
+						}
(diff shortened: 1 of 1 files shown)


vzool:
Hi, I think if you just make the time object a pointer it will be `nil` and that will be reflected as a `NULL` value in the database.

```golang
StartDate  *time.Time
```
This is what result I got with **SQLite**, I hope it behave similar in other databases.

Thanks

itcuihao:
> Hi, I think if you just make the time object a pointer it will be `nil` and that will be reflected as a `NULL` value in the database.
> 
> ```go
> type Contract struct {
>     gorm.Model
> 
>     StartDate  *time.Time
> }
> ```
> 
> This is what result I got with **SQLite**, I hope it behave similar in other databases.
> 
> Thanks

NO! This is the gorm logger operation,please see the code.

inliquid:
This was always confusing, I think makes sense to fix.

PR #2640, opened 2019-08-31
Fixed Ignore Blank Fields On Create

jamesh1999:
Removed erroneous IsPrimaryKey check which caused fields that were not part of the Primary Key but were Blank to be inserted.

(Particularly a problem because DeletedAt would be set to the zero value.)

- [x] Do only one thing
- [x] No API-breaking changes

Fixed an issue where zero valued fields would be inserted instead of NULL/default values. This includes the DeletedAt field which would cause all rows to be automatically marked as deleted


jinzhu:
blank primary key should not be inserted when insert.

PR #2702, opened 2019-10-15
support custom namer 

theoneLee:
hi.
I had a problem that the mapping of existed db tables column and models，just like：

db table define:
```
CREATE TABLE `tags` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `name` varchar(100) ,
  `createdOn` int(10) ,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;

``` 
model define：
```
type Tag struct{
    Id int
    Name string
    CreateOn int 
}
```

if i can not set `gorm:"column:createdOn"` for CreateOn field ,it will be mapping to created_on,
so i check the source code about this mapping method and try to custom.
but i can not do it because the `smap` is not public access.

this pr is solve it,and gorm users can custom column namer and use `smap` to cache.

### DIFF
--- a/naming.go
+++ b/naming.go
@@ -66,7 +66,7 @@ func ToColumnName(name string) string {
 	return TheNamingStrategy.ColumnName(name)
 }
 
-// mapping cache map
+// Smap is mapping cache map
 var Smap = newSafeMap()
 
 func defaultNamer(name string) string {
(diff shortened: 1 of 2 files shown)


theoneLee:
https://github.com/jinzhu/gorm/pull/2040
it can solve this mapping problem.

theoneLee:
#2040 can solve it,but it can not use `smap` to cache the mapping result.

jinzhu:
Hello @theoneLee 

It's better to don't cache the result as it might be dynamic, thank you.

PR #2713, opened 2019-10-20
feat: support go1.13 errors.Is

chyroc:

- [x] Do only one thing
- [x] No API-breaking changes

support go 1.13 `errors.Is`

```go
type BookPO struct {
	// ...
}

func GetBookPO(id int64) (*BookPO, error) {
	po := new(BookPO)
	if err := db.Find(po, id).Error; err != nil {
		return nil, fmt.Errorf("get book fail: %w", err)
	}
	return po, nil
}

func main() {
	book, err := GetBookPO(1234)
	if err != nil {
		if gorm.IsRecordNotFoundError(err) {
			fmt.Println("not found")
		} else {
			panic(err)
		}
	}

	fmt.Println("book", book)
}
```

### DIFF
--- a/error_12.go
+++ b/error_12.go
@@ -1,0 +1,7 @@
+// +build !go1.13
+
+package gorm
+
+func isError(err, target error) bool {
+	return err == target
+}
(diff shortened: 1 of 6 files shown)


chyroc:
how to fix `unrecognized printf verb 'w'` in error_13_test ...

PR #2722, opened 2019-10-23
Add warning comment

kohago:

- [x] Do only one thing
- [x] No API-breaking changes

Update will not update zero value fields , And Delete will not delete record with deleteAt filed.
I was not aware of that  until i made some bugs .

I think it is better to add some warnings to the update and delete 's comment, so people will know that more directly and quickly.

 in Chinese maybe:　注意，有坑


emirb:
Thanks @kohago.

Issue #2703, opened 2019-10-15
where("id in (?)", []uint8{1,2})  got unexpected result


rbdoer:
### What version of Go are you using (`go version`)?
go version go1.11.4 darwin/amd64

### Which database and its version are you using?
mysql

### Please provide a complete runnable program to reproduce your issue. **IMPORTANT**

 When I was using `where("id in (?)", []uint8{1,2})`, I set LogMode=2 and I got the sql of "SELECT * FROM `test_table`  WHERE (id IN ('< binary>') ) ".   

After checking the code, I find out the code in scope.go line:612
```
case reflect.Slice: // For where("id in (?)", []int64{1,2})
	if scanner, ok := interface{}(arg).(driver.Valuer); ok {
		arg, err = scanner.Value()
		replacements = append(replacements, scope.AddToVars(arg))
	} else if b, ok := arg.([]byte); ok { 

		// HERE: if the arg is type of []byte,  it will be parsed to string
		replacements = append(replacements, scope.AddToVars(b))
	} 
```

I think it may be a bug to fix

zaneli:
Hmm...It may be unexpected, but I think this is a valid specification.

`uint8` is an alias for `byte`.

```go
vs := []uint8{97, 98}
s := string(vs) // ab
```

`97`, `98` slice or `ab` string seems to be indistinguishable.
You should use `[]int8`, `[]uint16` and so on.

rbdoer:
@zaneli  Yeah, we fix it by using []int8.  But I think gorm should be compatibility with []uint8 .
how about stop using "[]uint8{97, 98}" , instead of using string type "ab" ?

zaneli:
@rbdoer 
I agree with your opinion.
However, maybe not good to change default behavior.

This just came to my mind, I tried to send #2745.
If this PR is accepted, we can use as below.
```go
db.Set("gorm:bytes_value_as_array", true).Where("id IN (?)", []byte{1, 2})
```
or
```go
db = db.Set("gorm:bytes_value_as_array", true)
db.Where("id IN (?)", []byte{1, 2}).Find(&user)
```

jinzhu:
https://github.com/jinzhu/gorm/pull/2745#issuecomment-562183691

jinzhu:
https://github.com/jinzhu/gorm/pull/2745#issuecomment-562183691

PR #2745, opened 2019-11-04
Add `gorm:bytes_value_as_array` tag

zaneli:

- [x] Do only one thing
- [x] No API-breaking changes

fix for #2703

#### default behavior
```go
db.Where("id IN (?)", []byte{1, 2}).Find(&user)
```
⬇️ 
```sql
SELECT * FROM `users`  WHERE `users`.`deleted_at` IS NULL AND ((id IN ('<binary>')))
```

#### set `gorm:bytes_value_as_array`
```go
db.Set("gorm:bytes_value_as_array", true).Where("id IN (?)", []byte{1, 2}).Find(&user)
```
⬇️ 
```sql
SELECT * FROM `users`  WHERE `users`.`deleted_at` IS NULL AND ((id IN (1,2)))
```


jinzhu:
Hi @zaneli

Thank you for your PR, but I guess this feature might not that helpful for users.

They might not realize it before using, once they realized this issue, I would suggest they change the data type....

Thank you anyway ;)

zaneli:
@jinzhu 
Thank you for review and comment on my trivial PR.
I would be happy if I could contribute GORM, but I generally agree with you.

PR #2768, opened 2019-11-19
Better "Duplicate column name" handling

edv1n:
struct fields with column tag will force other fields under same `DBName` with no column tag be ignored to avoid duplicated column name issue

- [x] Do only one thing
- [x] No API-breaking changes

The original implementation of GORM cannot handle two fields with the same column name (DBName), even when one field has tag gorm:"column:DBName", the other field will also get processed and appeared in the SQL statement.

This PR tried to solve the issue by checking if a field has a DBName assigned by struct tag "column", if yes, then other fields that share the same DBName (while having no "column" tag will be ignored automatically.

This enables developers to embed an external struct and override fields with same name easily.

example usage:

```go
type User struct {
    gorm.Model
    ID int64 `gorm:"column:id"`
}
```

Before applying this PR, there will be sql error of "Error 1060: Duplicate column name 'id'".

After applying this PR, GORM will ignore the ID field inside `gorm.Model`. 

This PR might resolve https://github.com/jinzhu/gorm/issues/1504

---

P.s. This PR MIGHT be API-breaking as I am changing the behavior of modeling struct. Feel free to close this PR if a new Issue is required to discuss the implementation.

### DIFF
--- a/model_struct.go
+++ b/model_struct.go
@@ -633,6 +633,28 @@ func (scope *Scope) GetModelStruct() *ModelStruct {
 		}
 	}
 
+	// look for field which has tag COLUMN assigned, if found any, set other fields which has the same DBName, while have no COLUMN tag assigned, to IsIgnored = true
+	for _, v := range modelStruct.StructFields {
+		if column, ok := v.TagSettingsGet("COLUMN"); ok {
+			for k, field := range modelStruct.StructFields {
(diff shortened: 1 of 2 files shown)


jinzhu:
Hello @edv1n

Thank you for your PR, but I think this might introduce more trouble than its value, it would be better to raise the error to let users know there is an issue, they need to handle it carefully...

Thank you.

PR #2774, opened 2019-11-22
transaction blocks: don't swallow panics

cstrahan:

- [x] Do only one thing
- [x] No API-breaking changes

This improves upon #2767.

Previously, the code would swallow any panics, which isn't ideal;
panic is intended to be used when a critical error arises,
where the process should fail fast instead of trying to limp along.

This now defers the any recovery (if desired) to the client code.


huacnlee:
@jinzhu It's looks correct, we need this changes.

Issue #2796, opened 2019-12-06
Is this a bug ?


YangKian:
### What version of Go are you using (`go version`)?

go 1.13

### Which database and its version are you using?

Mysql 8.0

```go
//here is my struct
type Task struct {
    Id int
    Book string
    CreateTime time.Time `gorm:"column:create_time"`
}

func SelectTasks(start, end time.Time) []time.Time {
	var tasks []time.Time

	res := db.Table("task").Where("create_time BETWEEN ? AND ?", start, end).Pluck("DISTINCT(create_time) ", &tasks)

	if res.RecordNotFound() {
		log.Println("get RecordNotFound err")
	}
	log.Printf("SelectTasks find %d result", len(tasks))
	return tasks
}

func main() {

    ...
    //here omit the connection operation to mysql database
    
	start, _ := time.ParseInLocation("2006-01-02 15:04:05", "2019-11-25 00:00:00", time.Local)
	end, _ := time.ParseInLocation("2006-01-02 15:04:05", "2019-12-02 00:00:00", time.Local)
    
	res := models.SelectTasks(start, end)
    
	for _, v := range res {
		fmt.Println(v)
	}
}
```

I have create a table named `task`，When I use `Pluck` combined with `Where` to select the data in the time durations, I get the following result:

![image-20191206223701812](https://user-images.githubusercontent.com/45479280/70331524-95693c00-187a-11ea-8f51-fb5e154af0c5.png)

the number of `rows affected or returned` are wrong, I get five result, not zero.

And when I try to select data in a duration which not include in my database, the error detect also not work . Here is an example:

```go
func main() {

    ...
    //here omit the connection operation to mysql database
    
    //I modify the start and end time, and make sure there are no related data in my database
	start, _ := time.ParseInLocation("2006-01-02 15:04:05", "2019-12-25 00:00:00", time.Local)
	end, _ := time.ParseInLocation("2006-01-02 15:04:05", "2019-12-26 00:00:00", time.Local)
    
	res := SelectTasks(start, end)
    
	for _, v := range res {
		fmt.Println(v)
	}
}
```

The example should have to print `get RecordNotFound err`，but it not

![image-20191206224344155](https://user-images.githubusercontent.com/45479280/70331554-a3b75800-187a-11ea-9f8c-f6d1a5b541b5.png)

Are these conditions bugs?

zaneli:
> When I use Pluck combined with Where to select the data in the time durations, I get the following result: ...

[...]

PR #2800, opened 2019-12-08
Fix rows affected log for Pluck, and remove rows affected log uncountable case and getting single row case

zaneli:

- [x] Do only one thing
- [x] No API-breaking changes

* fixed: #2796
When use `Pluck` fix [n rows affected or returned ] log.
* https://github.com/jinzhu/gorm/pull/2143#issuecomment-435590448
remove it from log for `Row`

#### before
* use `Find`
```sql
[2019-12-08 23:59:45]  [1.22ms]  SELECT * FROM `users`  WHERE `users`.`deleted_at` IS NULL AND ((id IN (1,2)))
[2 rows affected or returned ]
```
* use `Pluck`
```sql
[2019-12-08 23:59:45]  [1.22ms]  SELECT name FROM `users`  WHERE (id IN (1,2))
[0 rows affected or returned ]
```
* use `Count`
```sql
[2019-12-08 23:59:45]  [1.17ms]  SELECT count(*) FROM `users`  WHERE (id IN (1,2))
[0 rows affected or returned ]
```

#### after
* use `Find`
```sql
[2019-12-09 00:00:24]  [1.34ms]  SELECT * FROM `users`  WHERE `users`.`deleted_at` IS NULL AND ((id IN (1,2)))
[2 rows affected or returned ]
```
* use `Pluck`
```sql
[2019-12-09 00:00:24]  [1.08ms]  SELECT name FROM `users`  WHERE (id IN (1,2))
[2 rows affected or returned ]
```
* use `Count`
```sql
[2019-12-09 00:00:24]  [1.25ms]  SELECT count(*) FROM `users`  WHERE (id IN (1,2))
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -366,12 +366,16 @@ func (s *DB) Scan(dest interface{}) *DB {
 
 // Row return `*sql.Row` with given conditions
 func (s *DB) Row() *sql.Row {
-	return s.NewScope(s.Value).row()
+	scope := s.NewScope(s.Value)
+	defer scope.trace(NowFunc(), false)
+	return scope.row()
 }
 
 // Rows return `*sql.Rows` with given conditions
 func (s *DB) Rows() (*sql.Rows, error) {
-	return s.NewScope(s.Value).rows()
+	scope := s.NewScope(s.Value)
+	defer scope.trace(NowFunc(), false)
+	return scope.rows()
 }
 
 // ScanRows scan `*sql.Rows` to give struct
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -366,12 +366,16 @@ func (s *DB) Scan(dest interface{}) *DB {
...
+	scope := s.NewScope(s.Value)
+	defer scope.trace(NowFunc(), false)
+	return scope.row()
 }
 
 // Rows return `*sql.Rows` with given conditions
 func (s *DB) Rows() (*sql.Rows, error) {
-	return s.NewScope(s.Value).rows()
+	scope := s.NewScope(s.Value)
+	defer scope.trace(NowFunc(), false)
```
zaneli: `Rows` returns multiple records, but in this case `scope.db.RowsAffected` cannot increment.
So I think it's better to remove it.

PR #2834, opened 2020-01-05
add: disable default logger color with NO_COLOR support.

MarAvFe:

- [X] Do only one thing
- [X] No API-breaking changes

Add [NO_COLOR](https://no-color.org/) compliance to gorm.

When logging, search for the NO_COLOR environment variable. If its not empty, suppress the color from the formatter. Else, log with the usual colors.

Also, a bit of refactor was done to achieve the task simpler than rewritting a new formatter:
- Export colors to variables
- Using string templates in all prints as opposed to cat strings

This behavior can be tested as:

```shell
$ NO_COLOR= go test  # same as $ go test
PASS
ok  	github.com/jinzhu/gorm	26.724s
$ NO_COLOR=1 go test
PASS
ok  	github.com/jinzhu/gorm	26.760s
```

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -13,13 +13,18 @@ import (
 )
 
 const (
+	// RESET is the shell code to disable output color
 	RESET = "\033[0m"
-
-	CYAN    = "\033[36;1m"
+	// CYAN is the shell code for CYAN output color
+	CYAN = "\033[36;1m"
+	// MAGENTA is the shell code for MAGENTA output color
 	MAGENTA = "\033[35m"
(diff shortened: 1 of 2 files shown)


jinzhu:
Thank you for your PR, I think it would be better to be an option of the default logger https://github.com/jinzhu/gorm/blob/e2ed6ffb4c96a5293f83e6be94a163298e72e348/logger.go#L179

jinzhu:
Thank you for your PR, implemented in V2.

PR #2880, opened 2020-02-12
Makes order of values deterministic for IN(?, ?..) style query

jitendra-1217:

- [x] Do only one thing
- [x] No API-breaking changes

#### What did this pull request do?

- [x] Makes order of values deterministic for IN(?, ?..) style query

  Currently during preloads of associations the order of parameterized/comma-
separated values for IN(?, ?..) style query is not deterministic. This commit
fixes that which removes random failure when writing tests which asserts exact
queries run (e.g. using DATA-DOG/go-sqlmock library or such).

### DIFF
--- a/scope.go
+++ b/scope.go
@@ -1324,7 +1324,7 @@ func (scope *Scope) autoIndex() *Scope {
 }
 
 func (scope *Scope) getColumnAsArray(columns []string, values ...interface{}) (results [][]interface{}) {
-	resultMap := make(map[string][]interface{})
+	resultMap := make(map[string]bool)
 	for _, value := range values {
 		indirectValue := indirect(reflect.ValueOf(value))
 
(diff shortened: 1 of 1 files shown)


scope.go
```diff
@@ -1324,7 +1324,7 @@ func (scope *Scope) autoIndex() *Scope {
 }
 
 func (scope *Scope) getColumnAsArray(columns []string, values ...interface{}) (results [][]interface{}) {
-	resultMap := make(map[string][]interface{})
+	resultMap := make(map[string]bool)
```
Dasio: use struct{} instead of bool, you are not using that value anywhere

jitendra-1217: Yes the map's value is not used. Using this(i.e. resultMap) as a set.
Didn't know about struct{} thing. Just asking is that common?

I will make the changes. But it does not seem this will get into master. [Ref](https://github.com/jinzhu/gorm/issues/2886)

PR #2921, opened 2020-03-09
Add RollbackUnlessCommitted

DoubleDi:

- [x] Do only one thing
- [x] No API-breaking changes

Hi @jinzhu ! Awesome refactoring with v2 going on.
Added RollbackUnlessCommitted func for transactions. Helps not to write a lot of rollback code on error checking!

Example
```
tx := db.Begin()
defer tx.RollbackUnlessCommitted()
if somerrhappened {
     // no need for rollback here
     return
}
if othererrhappened {
     // no need for rollback here
     return
}
tx.Commit()
```

also please keep in mind that you should write commiTTer with double T.

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -214,8 +214,8 @@ func (db DB) Begin(opts ...*sql.TxOptions) (tx DB) {
 
 // Commit commit a transaction
 func (db DB) Commit() DB {
-	if comminter, ok := db.Statement.ConnPool.(TxCommiter); ok && comminter != nil {
-		db.AddError(comminter.Commit())
+	if commiter, ok := db.Statement.ConnPool.(TxCommiter); ok && commiter != nil {
+		db.AddError(commiter.Commit())
 	} else {
 		db.AddError(ErrInvalidTransaction)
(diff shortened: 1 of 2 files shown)


jinzhu:
Hello,

Thank you for your PR.

It won't work after adding the `RollbackUnlessCommitted` method to `TxCommiter ` as `database/sql` doesn't implement this method.

And I would suggest we don't add API like this but just using method `Transaction` like this  http://gorm.io/docs/transactions.html#Transactions

Could you open another PR to fix the typo, thank you ;)

DoubleDi:
Yes, didn't think of that. 
I reviewed the changes and we actually don't  need `RollbackUnlessCommitted` in  `TxCommitter`, only need it in `DB`. So maybe this PR is still considerable? 
Or maybe we can use you callbacks and put beside `CommitOrRollback`?

jinzhu:
Have you checked out the `Transaction` method? I prefer to use it than adding a new API.

DoubleDi:
Ok it's pretty handy too. Thanks

Issue #3007, opened 2020-05-06
Method chaining corrupts data due to slice capacity increasing


tekkamanendless:
Method chaining corrupts data due to the fact that `search.clone()` doesn't actually copy its slices.

The first time that you call `Where`, `whereConditions` goes from a `nil` slice to a new slice with length=1, capacity=1.

The second time that you call `Where`, `whereConditions` goes from a slice with length=1, capacity=1 to a new slice with length=2, capacity=2.

The third time that you call `Where`, `whereConditions` goes from a slice with length=2, capacity=2 to a new slice with length=3, capacity=4.  It is here that Go's slice creation has allocated more capacity than we are using.

The fourth time that you call `Where`, `whereConditions` remains the same; its length merely grows to 4 and the fourth element is set.

If you used the first 3 `Where` calls to create a template query, and then if you create two queries that each perform one more `Where` independently of each other, then the results will be corrupt.

Here's a simple test case to show that this is what's happening.  It starts creates a "base" query with 1, 2, 3, and then 4 initial `Where` calls, and then it creates two additional queries that each use that base to perform a single additional `Where` call.

In all cases, the text `THING1` should only be present in query `q1`, and `THING2` should only be present in query `q2`.  However, this is _not_ the case when `w=3`, meaning that the base query has 3 `Where` calls to start with.

```go
package main

import (
	"fmt"
	"strings"
	"testing"

	"github.com/jinzhu/gorm"
	_ "github.com/jinzhu/gorm/dialects/sqlite"
	"github.com/stretchr/testify/assert"
)

func TestWhereClone(t *testing.T) {
	databaseType := "sqlite3"
	databaseURL := "file:no-provided-database?mode=memory"

	db, _ := gorm.Open(databaseType, databaseURL)
	db.LogMode(true)

	for whereCount := 1; whereCount <= 4; whereCount++ {
		t.Run(fmt.Sprintf("w=%d", whereCount), func(t *testing.T) {
			query := db.Table("myTable")
			for w := 0; w < whereCount; w++ {
				query = query.Where(fmt.Sprintf("w%d = ?", w), fmt.Sprintf("value%d", w))
			}

			q1 := query.Where("finalThing = ?", "THING1")
			q2 := query.Where("finalThing = ?", "THING2")

			e1 := fmt.Sprintf("%v", q1.QueryExpr())
			e2 := fmt.Sprintf("%v", q2.QueryExpr())

			assert.True(t, strings.Contains(e1, "THING1"))
[...]

PR #3040, opened 2020-06-07
Fix Statement Where clone array corruption in v2

tekkamanendless:
This is an updated fix for jinzhu/gorm #3007 for the new code base.  While the
overall structure has changed drastically, the primary issue behind
the original problem is still present.  In this case, `MergeClause` does
not create a proper copy and instead allows for two clauses to have
the same backing array.

----
    
Method-chaining in gorm is predicated on a `Clause`'s `MergeClause`
method ensuring that the two clauses are disconnected in terms of
pointers (at least in the Where case).
    
However, the original Where implementation used `append`, which
only returns a new instance if the backing array needs to be resized.
In some cases, this is true.  Practically, go doubles the size of the
slice once it gets full, so the following slice `append` calls would
result in a new slice:
    
* 0 -> 1
* 1 -> 2
* 2 -> 4
* 4 -> 8
* and so on.
    
So, when the number of "where" conditions was 0, 1, 2, or 4, method-chaining
would work as expected.  However, when it was 3, 5, 6, or 7, modifying the
copy would modify the original.
    
This also updates the "order by", "group by" and "set" clauses.

### DIFF
--- a/clause/group_by.go
+++ b/clause/group_by.go
@@ -30,8 +30,13 @@ func (groupBy GroupBy) Build(builder Builder) {
 // MergeClause merge group by clause
 func (groupBy GroupBy) MergeClause(clause *Clause) {
 	if v, ok := clause.Expression.(GroupBy); ok {
-		groupBy.Columns = append(v.Columns, groupBy.Columns...)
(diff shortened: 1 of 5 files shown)


jinzhu:
thank you for your PR.

PR #3076, opened 2020-06-23
Create supports Array / ArrayPtr

soranoba:

- [x] Do only one thing
- [x] No API-breaking changes

Create function supports batch create, but only Slice is supported.
There are cases where it would be desirable if Array could be used in some cases.
For example, when using faker to create a fixed number of records.

This change supports it.

-----

By the way, in my local PC the other test failed than before the PR, but I ignored it.
It seems that the return value of test script is always 0 (it means success)

Please use the patch below.

```diff
diff --git a/tests/tests_all.sh b/tests/tests_all.sh
index a321fe3..f85a7a4 100755
--- a/tests/tests_all.sh
+++ b/tests/tests_all.sh
@@ -1,3 +1,4 @@
+#!/bin/sh -e
 dialects=("sqlite" "mysql" "postgres" "sqlserver")
 
 if [[ $(pwd) == *"gorm/tests"* ]]; then
```

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -73,7 +73,7 @@ type Tabler interface {
 // get data type from dialector
 func Parse(dest interface{}, cacheStore *sync.Map, namer Namer) (*Schema, error) {
 	modelType := reflect.ValueOf(dest).Type()
-	for modelType.Kind() == reflect.Slice || modelType.Kind() == reflect.Ptr {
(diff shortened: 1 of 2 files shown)


jinzhu:
> By the way, in my local PC the other test failed than before the PR, but I ignored it.
> It seems that the return value of test script is always 0 (it means success)

Thank you for your report, fixed that.

PR #3101, opened 2020-07-02
fix no limit no offset.

SmallTianTian:
- [x] Do only one thing
- [x] Tested

修复当 `limit` 为非正整数时，`offset` 的设置将无效的问题。

### User Case Description

1. 仅设置 `offset`，能正常追加到 `SQL`
1. 多次设置  `offset`，将使用最后设置的值
1. 设置 `offset` 后可再次设置 `limit`，互不干涉

### DIFF
--- a/clause/limit.go
+++ b/clause/limit.go
@@ -18,11 +18,13 @@ func (limit Limit) Build(builder Builder) {
 	if limit.Limit > 0 {
 		builder.WriteString("LIMIT ")
 		builder.WriteString(strconv.Itoa(limit.Limit))
-
-		if limit.Offset > 0 {
-			builder.WriteString(" OFFSET ")
-			builder.WriteString(strconv.Itoa(limit.Offset))
+	}
+	if limit.Offset > 0 {
+		if limit.Limit > 0 {
+			builder.WriteString(" ")
 		}
+		builder.WriteString("OFFSET ")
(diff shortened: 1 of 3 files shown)


jinzhu:
thank you for your PR.

omegaatt36:
Hello, I am a user who is using gorm v1 to upgrade to gorm v2. I use `snowflake` id in part of the schema, but this needs to be INT64.

```
func (limit Limit) Build(builder Builder) {
...
    builder.WriteString(strconv.FormatInt(limit.Offset, 10))
}

```

The following changes enable Limit to support int64, but since drivers are dependent on a different repo, and drivers updated offset [11 hours ago](https://github.com/go-gorm/sqlite/commit/4fe70a2cc521c8da55599c978311701e70d2aec2), direct changes will not pass the test, and presumably will not be able to launch PR.

 If this is a feasible solution, please let me know how to contribute to gorm and the drivers that have been established.

Thanks.

SmallTianTian:
@omegaatt36 It's a good idea. but I don't think that's necessary.
1. How do you deal with `uint64`？It's bigger than `int64`.
2. Most machines are 64-bit, `int` equal `int64`.
3. Even though you're using a 32-bit machine, `int32` max value is `2,147,483,647`, it's to big for a table.

I think you can use `int(int64limit)`.
Thanks.

PR #3174, opened 2020-07-20
Woraround for some database that returns capitalized letter for non-terminal column name

stevefan1999-personal:
- [x] Do only one thing
- [ ] ~Non breaking API changes~ (potentially breaking)
- [ ] Tested

When scanning rows, just convert all the column names to lower case

### User Case Description
This is to remediate some database who actually returns capitalized name no matter what the name you specified in your table creation (Yes I mean you and perhaps only you Oracle)

But this also mean you might fuck up the namer because what if they do like capitalized name?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,4 +5,5 @@ go 1.14
 require (
 	github.com/jinzhu/inflection v1.0.0
 	github.com/jinzhu/now v1.1.1
+	github.com/thoas/go-funk v0.7.0
 )
--- a/gorm.go
+++ b/gorm.go
@@ -32,6 +32,10 @@ type Config struct {
 	DisableAutomaticPing bool
 	// DisableForeignKeyConstraintWhenMigrating
 	DisableForeignKeyConstraintWhenMigrating bool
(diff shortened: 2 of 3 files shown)


jinzhu:
Hello @stevefan1999-personal 

Thank you for your PR, good to know that.

But I would suggest you replace the default `NamingStrategy` for Oracle, and use upper case for it?

Refer http://v2.gorm.io/docs/gorm_config.html#NamingStrategy

stevefan1999-personal:
@jinzhu you cant, because the auto migrator cannot figure out the relation struct keys for some reason. A workaround is to use EqualFold on schema parsing, but that also break stuff

stevefan1999-personal:
@jinzhu https://github.com/go-gorm/playground/runs/893128227#step:6:77 see

stevefan1999-personal:
also, how do you tell go to update the dependencies for the latest commit...it lagged behind to the commit i made last day

stevefan1999-personal:
although technically you can rename every single column to its capitalized equivalent via struct tag...this is very verbose. how do i reopen this again @jinzhu?

PR #3198, opened 2020-07-25
proposal for struct scanning

pkpfr:
- [Proposal to address https://github.com/go-gorm/gorm/issues/3161] Do only one thing
- [NO] Non breaking API changes
- [YES] Tested

It is often not possible for users to specify models and relationships.  Especially when using Gorm with third party databases.

In other scenarios, defining models with several nested structs can result in Gorm creating sub-optimal queries.  In cases where performance is required, raw queries can provide a significant advantage.

Scanning results from raw queries into custom structs that contain multiple nested structs with overlapping IDs currently does doesn't work.  Results for fields with identical names will be overwritten.  This is different to behaviour in Gorm V1.

### User Case Description

```
type ProductSubscriptionBillingProfile struct {
	db.ProductSubscription
	db.BillingProfile
	db.Product
}
```

Each of the above structs share common fields for ID, CreatedAt, UpdatedAt, etc.  These structs have been defined in model generated by protocol buffers and nesting doesn't make sense due to how the same models are used elsewhere in the codebase.

Gorm V1 will successfully scan a result into the above model.  Gorm V2 will not.

### Considerations

This proposal breaks the User's ability to manipulate the order of values in a custom struct.  The order of the values would be required to match the order of the select statement.  However, this is something that can be easily documented and should not present a significant issue to users.

Performance wise, there is no significant impact, as the current implementation iterates over columns.

For any users wishing to use Gorm on databases which they do not control, or with models generated by third party functions such as protobuf the serve multiple purposes such as GRPC marshalling and transport.  The ability to retain this V1 functionality is critical.

The alternative of manually scanning rows would result in a high maintenance overhead and would negate the need to use this project at all.


jinzhu:
I think this is not acceptable by users to change struct's order to make GORM works.

https://github.com/go-gorm/gorm/pull/3198/files#diff-8c84d7b16f764dab2ba135ab6aa36cb4R147-R151

PR #3207, opened 2020-07-28
Add json tag to struct Model

15ho:
- [x] Do only one thing
- [x] Tested

Add json tag to struct Model

### User Case Description
```go
//type OldModel struct {
//	ID        uint      `gorm:"primarykey"`
//	CreatedAt time.Time
//	UpdatedAt time.Time
//	DeletedAt DeletedAt
//}

type User struct {
       OldModel
       Name string `json:"name"`
}

// json output:
//{
//	"ID":0,
//	"CreatedAt":"0001-01-01T00:00:00Z",
//	"UpdatedAt":"0001-01-01T00:00:00Z",
//	"DeletedAt":null,
//	"name":""
//}
```

### DIFF
--- a/model.go
+++ b/model.go
@@ -8,8 +8,8 @@ import "time"
 //      gorm.Model
 //    }
 type Model struct {
-	ID        uint `gorm:"primarykey"`
-	CreatedAt time.Time
-	UpdatedAt time.Time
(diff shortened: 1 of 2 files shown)


jinzhu:
Just don't use `gorm.Model` and define your own one.

PR #3247, opened 2020-08-08
refactor function CheckTruth

0x2d3c:
refactor function CheckTruth. each diff type need a muched case to handle


jinzhu:
Hello, could you fix the reviewdog issue, thank you.

0x2d3c:
> Hello, could you fix the reviewdog issue, thank you.

I can't solve it well，the previous version is better

PR #3325, opened 2020-08-28
prepare_stmt add ctx

LanLanBoom:
- [x] Do only one thing
- [x] Tested

Context are lost in PreparedStmtDB and PreparedStmtTX functions, using context.Backgroud instead.  

This pull request fixed this problem.

### User Case Description

PreparedStmt insert, update, delete tested.

PreparedStmtTX insert, update, delete tested.

### DIFF
--- a/prepare_stmt.go
+++ b/prepare_stmt.go
@@ -25,7 +25,7 @@ func (db *PreparedStmtDB) Close() {
 	db.Mux.Unlock()
 }
 
-func (db *PreparedStmtDB) prepare(query string) (*sql.Stmt, error) {
+func (db *PreparedStmtDB) prepare(ctx context.Context, query string) (*sql.Stmt, error) {
 	db.Mux.RLock()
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi, we don't include the context in the API on purpose,  so we can still cache the statement even it is a timeout context

Thank you for your PR anyway.

jinzhu:
does it cause any issue for you?

LanLanBoom:
> does it cause any issue for you?

yes, we are developing a mysql driver, wrapped around github.com/go-sql-driver/mysql with some logic packed, which relies on the ctx param being passed around.  

I think caching the statement even it is a timeout context is a rare case with little performance improvement.  However ctx passed arount is vital for extensibility. 

Perhaps passing arount ctx is more important, considering code extensibility?

jinzhu:
https://github.com/go-gorm/gorm/commit/9b0ad4730f16d6ac7cf18d1aa42d74714959745b

merged, thank you for your pull request

PR #3329, opened 2020-08-29
Woraround for some database that returns capitalized letter for non-terminal column name (reopen of #3174)

stevefan1999-personal:
- [x] Do only one thing
- [] Non breaking API changes
- [x] Tested

see #3174

### User Case Description
see #3174


jinzhu:
Hello @stevefan1999-personal

Thank you for your PR, I have checked few documents, seems for Oracle, if the column name was quoted (that what we are doing), it won't return the capitalized letter, so I don't this PR is the correct behavior.

https://stackoverflow.com/questions/24504458/to-make-entire-column-names-in-lower-case-in-oracle
https://seeq.atlassian.net/wiki/spaces/KB/pages/443088907/SQL+Column+Names+and+Case+Sensitivity#:~:text=lOo%2C%20LOO%2C%20etc-,Oracle,not%20case%20sensitive%20unless%20quoted.

But I think Oracle works best with uppercase names, so I think maybe we should change the default [NamingStrategy](https://gorm.io/docs/gorm_config.html#NamingStrategy) when initializing oracle driver and returns uppercase for the column name.

Thank you.

stevefan1999-personal:
@jinzhu Again, this is not what I wanted to, there's apparently some problem in Gorm that it cannot process the metadata, so for example that the many-to-many relationship data cannot be parsed in oracle with column name all capitalized. I've gave you the sample Action result that directly expressed this problem, geez.

Fine, I would try and see if I had a better alternative to fix this.

jinzhu:
> for example that the many-to-many relationship data cannot be parsed in oracle with column name all capitalized.

If you changed the `NamingStrategy`'s logic, this should works.

PR #3334, opened 2020-08-30
feature: using default limite and offset query for count

rainingmaster:
- [x] Do only one thing
- [x] Tested

Get `count` more convenient for pagination.

### User Case Description

Sometime we want to paginate convenient, we will write like below, but we will get a `0` for `count`: 

```golang
var (
    user []User
    count int
)

if err := DB.Where("name=?", "john").Limit(10).Offset(20).Find(&users).Count(&count).Error; err != nil {
       t.Errorf("Not error should happen, but got %v", err)
}
```

It is because if we run `Count` with `Offset`, we will get an empty data set.

After my attempts, this can achieve the `count`

```golang
var (
    user []User
    count int
)

db := DB.Where("name=?", "john")
if err := db.Limit(10).Offset(20).Find(&users).Error; err != nil {
       t.Errorf("Not error should happen, but got %v", err)
}

if err := db.Count(&count).Error; err != nil {
       t.Errorf("Not error should happen, but got %v", err)
}
```

or

```golang
var (
    user []User
    count int
)

if err := DB.Where("name=?", "john").Limit(10).Offset(20).Find(&users).Offset(0).Count(&count).Error; err != nil {
       t.Errorf("Not error should happen, but got %v", err)
}
```
**But the code will be more verbose. I think more people hope to complete the goal in one chain call, like** [here](https://jasperxu.github.io/gorm-zh/crud.html#count)

### DIFF
--- a/clause/limit.go
+++ b/clause/limit.go
@@ -6,6 +6,7 @@ import "strconv"
 type Limit struct {
 	Limit  int
 	Offset int
+	Ignore bool
 }
 
 // Name where clause name
@@ -15,6 +16,9 @@ func (limit Limit) Name() string {
 
 // Build build where clause
 func (limit Limit) Build(builder Builder) {
+	if limit.Ignore {
+		return
+	}
 	if limit.Limit > 0 {
 		builder.WriteString("LIMIT ")
 		builder.WriteString(strconv.Itoa(limit.Limit))
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -322,6 +322,9 @@ func (db *DB) Count(count *int64) (tx *DB) {
 		defer tx.Statement.AddClause(clause.Select{})
 	}
 
+	tx.Statement.AddClause(clause.Limit{Ignore: true})
+	defer tx.Statement.AddClause(clause.Limit{Ignore: false})
+
 	tx.Statement.Dest = count
 	tx.callbacks.Query().Execute(tx)
 	if tx.RowsAffected != 1 {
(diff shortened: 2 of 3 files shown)


rainingmaster:
sister PR for https://github.com/jinzhu/gorm/pull/8

jinzhu:
Write the chain like `DB.Where("name=?", "john").Count(&count).Limit(10).Offset(20).Find(&users)`

BTW, don't use the above document, it is tooo old

use https://cn.gorm.io/zh_CN/

crcms:
> Write the chain like `DB.Where("name=?", "john").Count(&count).Limit(10).Offset(20).Find(&users)`
> 
> BTW, don't use the above document, it is tooo old
> 
> use https://gorm.gitee.io/zh_CN/
I tried this method, but I always reported "unsupported data type: 0xc00088a070", and it turned out to be a call to `count()`

```
err = db.Preload(`User.Certification`).Preload(`User.Followers`).Preload(`Files.File`).
		Preload(`Likes`).Preload(`Collections`).Preload(`Comments`).
		Joins(`Pinned`).
		Limit(limit).Offset(offset).Order(`Pinned.expires_at desc`).
		Order(`feeds.id desc`).Find(&feeds).Count(&total).Error
```
There are no exceptions when I use this method，But when I move the Count position, I report an exception `unsupported data type: 0xc00088a070`
```
err = db.Preload(`User.Certification`).Preload(`User.Followers`).Preload(`Files.File`).
		Preload(`Likes`).Preload(`Collections`).Preload(`Comments`).
		Joins(`Pinned`).
                Count(&total).
		Limit(limit).Offset(offset).Order(`Pinned.expires_at desc`).
		Order(`feeds.id desc`).Find(&feeds).Error
```

PR #3341, opened 2020-08-31
db instance add method WithLogger

aapelismith:
- DB struct instance add method WithLogger()

A method WithLogger() is added to the DB instance to quickly modify the logger of the DB instance.
Add this method based on the following scenario:
    Each log line needs to print an x-request-id so that ELK can search the log to find the data containing x-request-id in the entire call chain.


jinzhu:
Hi @aapelismith

Thank you for your PR, but want to make GORM's API tiny & clean, so not going to merge this.

PR #3425, opened 2020-09-08
Fix monocolor log output inconsist with colorful log

egenchen:
- [x] Do only one thing
- [x] Tested

This pull request fixed the logger output execution time format in mono color mode was different from colorful mode.

### User Case Description

In mono color mode, the execution time output: [0]
In colorful mode, the execution time output: [0.001ms]


jinzhu:
thank you for your PR.

PR #3433, opened 2020-09-08
Release the connection when discovering the column types in the migrator

### LINKED ISSUE #3432: AutoMigrate does not close sql.Rows when retrieving the schema
## GORM Playground Link

https://github.com/go-gorm/playground/pull/122

## Description

This happens when `AutoMigrate()` is run on a database that already has data. The `AutoMigrate()` function attempts to discover the schema and it runs a `select * from <table>` query and then reads the column tables. It never reads any of the data and does not close the rows. The database drivers keep this connection open indefinitely until the application is shutdown. This causes an application that uses `AutoMigrate()` to keep open `N+1` connections at a minimum with `N` being the number of tables in the database. The `N` connections are never recycled.

This only appears to be an issue if there is data present in the database. When no data is present, the connection immediately closes and the connection is not leaked.

jsternberg:
When the migrator is used to discover the column types, such as when
used with `AutoMigrate()`, it does not close the query result. This
changes the migrator to close the query result and it also changes the
query to use `LIMIT 1` to prevent additional work against the database
when only discovering the schema.

Fixes #3432.

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -388,9 +388,10 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 
 func (m Migrator) ColumnTypes(value interface{}) (columnTypes []*sql.ColumnType, err error) {
 	err = m.RunWithValue(value, func(stmt *gorm.Statement) error {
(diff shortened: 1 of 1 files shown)


jsternberg:
The test failures for sqlserver seem to be caused by [this issue](https://github.com/denisenkom/go-mssqldb/issues/598) rather than anything related to this PR.

PR #3447, opened 2020-09-10
Feat add SetLogMode

gsxhnd:
- [x] Do only one thing
- [x] Tested

Add `logger.SetLogMode`, that will replace logger Config LogLevel, not return new logger config instance.

### User Case Description
```golang
db, err := gorm.Open(mysql.Open(dns), &gorm.Config{
	Logger: logger.Default.SetLogMode(logger.Silent),
})
```

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -49,6 +49,7 @@ type Config struct {
 // Interface logger interface
 type Interface interface {
 	LogMode(LogLevel) Interface
+	SetLogMode(LogLevel) Interface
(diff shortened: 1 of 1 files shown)


jinzhu:
Hello @gsxhnd 

Thank you for your PR, your issues was caused by callbacks are not using db's logger but using `logger.Default`

Honestly, I am not a fan for adding this API, and the above commit should fixed your issue, thank you.

Issue #3451, opened 2020-09-11
Default tag not working in embedded schema


rocinantex:
## https://github.com/go-gorm/playground/pull/136

```go
type StatusEmbed struct {
    Status string `gorm:"not null;default:''"`
}

type Bugs struct {
    ID int64 `gorm:"primaryKey"`

    Foo string
    StatusEmbed
}

fmt.Println(DB.Select("foo").Create(new(Bugs)).Error)
```

In this case will occurs an error: NOT NULL constraint failed: bugs.status

PR #3453, opened 2020-09-11
Fix default tag not working in embedded schema (#3451)

rocinantex:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


jinzhu:
Fixed in commit https://github.com/go-gorm/gorm/commit/b8a74a80d732963df95580eae3316db140a882a4 , thank you for your PR.

PR #3492, opened 2020-09-19
Fix format sql log

caelansar:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


jinzhu:
thank you for your pr

PR #3512, opened 2020-09-22
Add option to detect unique constraint errors

### LINKED ISSUE #3499: Add option to allow check unique constraint (duplicate) errors
Currently there is only RecordNotFound error check. Would be great if there was a way to check also unique errors as for different databases errors are different. Ideally it would be great if error would contain also column name in question.

Ref: #2903

lafriks:
- [x] Do only one thing
- [x] Tested

Add new error `ErrUniqueConstraint`

### User Case Description

To allow detect unique constraint errors independently of used database

Fixes #3499

### DIFF
--- a/callbacks/create.go
+++ b/callbacks/create.go
@@ -92,12 +92,12 @@ func Create(config *Config) func(db *gorm.DB) {
 										}
 									}
 								} else {
-									db.AddError(err)
+									db.AddErrorFromDB(err)
 								}
 							}
 						}
 					} else {
-						db.AddError(err)
+						db.AddErrorFromDB(err)
 					}
 				}
 			}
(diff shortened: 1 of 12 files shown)


lafriks:
To me it looks like MSSQL has error somewhere as it fails to execute statement but does not return error either. (see https://github.com/go-gorm/gorm/pull/3512/files#diff-f5a0f20e67b1e257a236d53af224d6eaR27)

Also I don't quite like that if/else error detection but moving them to specific dialect would require adding new method to dialect interface. That would allow more precise error detection also. Any suggestions?

jinzhu:
the repo of GORM won't accept the pull request, but I think it would be good to have this feature in GORM drivers to check dialector specified errors

Thank you for your PR.

lafriks:
Ok but how? Error itself should be in gorm, function to ask drivers for specific errors should also be in gorm (that means changing dialector inferface or adding new one?) I can rewrite this either way but recommendations would be welcome

jinzhu:
For example, we could add a `ParseErr` method to the mysql driver, which returns a struct like:

```go
type ErrResult struct {
  Column   string
  Error        error
  Retryable bool
}
```

If you could implement the `ParseErr` method to all dialectors, we could even add it to the `Dialector` interface, but the code itself should not implemented in GORM but in each drivers.

lafriks:
Would it be possible to add `ErrUniqueConstraint` error type to errors.go in gorm and than I could use it to add methods to drivers?

https://github.com/go-gorm/gorm/pull/3512/files#diff-76a7e0e299c7417bae32d3098e71370980157fe29e68a366671491d147d40572R38

PR #3515, opened 2020-09-23
WIP: feat: support complexity field with special tag saved as string.

j2gg0s:
This is one pr in WIPS, and want to get some advice.

With a special tag "xjson":
- before create/update, field will be json.marshaled to bytes
- before query, field will be assigned with json.unmarshaled

### User Case Description
With [GraphQL](https://graphql.org/) or [Thrift](http://thrift.apache.org/), tools can generated model by api definition.
However, i can't use these golang struct with gorm. Field type like: ``[]string``, ``[]int`` is common.

#### One Example:
I have a simple application to save my url links. The main object is ``Link``, the definition in graphql and generated model:
```graphql
type Link {
  id: ID!
  name: String!
  url: String!
  desc: String
  tags: [Tag!]
  createdAt: Time
  updatedAt: Time
}

scalar Tag
```
```golang
type Link struct {
	ID        string     `json:"id"`
	Name      string     `json:"name"`
	URL       string     `json:"url"`
	Desc      *string    `json:"desc"`
	Tags      []string   `json:"tags" gorm:"type:golang;xjson"`
	CreatedAt *time.Time `json:"createdAt"`
	UpdatedAt *time.Time `json:"updatedAt"`
}
```

Now, i want to save link in one table, and save tags as json.
And i don't want to do type alias, like: ``type TagsType []string``.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -12,7 +12,7 @@ import (
 func prepareValues(values []interface{}, db *DB, columnTypes []*sql.ColumnType, columns []string) {
 	if db.Statement.Schema != nil {
 		for idx, name := range columns {
-			if field := db.Statement.Schema.LookUpField(name); field != nil {
+			if field := db.Statement.Schema.LookUpField(name); field != nil && !field.XJSON {
 				values[idx] = reflect.New(reflect.PtrTo(field.FieldType)).Interface()
 				continue
 			}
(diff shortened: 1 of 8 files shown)


jinzhu:
Hello,

All customized field should implement the `sql.Scanner`, `value.Valuer` interface, check out the details from gorm.io

We are not going to accept any other kind of tags.

Thank you.

Issue #3542, opened 2020-09-28
Change the config assignment method to prevent initial configuration errors


godcong:
## Describe the feature

The current Open function is like this:
```go
func Open(dialector Dialector, config *Config) (db *DB, err error) {
//...
}
```

I think we should change it to function assignment to avoid problems caused by user misconfiguration of `config`
like this:
```go
type ConfigOption func(*Config)

func Open(dialector Dialector, options ...ConfigOption) (db *DB, err error) {
	//init config
	cfg := defaultConfig()
	for _,option := range options {
		//change config value
		option(cfg)
	}
}
```

## Motivation

Such changes can simplify assignment and prevent incorrect configuration.
And save some default value judgments in the code.
Considering the impact of the change, it might be better to add a new function.

PR #3543, opened 2020-09-28
example for #3542

godcong:
- [X] Do only one thing
- [X] Tested

Solve the problem I raised in #3542

### User Case Description

after fix we can open db like this:
```go
db, err = gorm.OpenD(sqlite.Open(filepath.Join(os.TempDir(), "gorm.db")))
```

### DIFF
--- a/gorm.go
+++ b/gorm.go
@@ -50,6 +50,8 @@ type Config struct {
 	cacheStore *sync.Map
 }
 
+type ConfigOption func(*Config)
+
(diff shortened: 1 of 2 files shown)


jinzhu:
doesn't sound an issue to me, usually user only need to initialize the `gorm.DB` once and I would like user clearly knows he can use the `gorm.Config` to make some changes.

jinzhu:
thank you for your PR anyway.

PR #3546, opened 2020-09-28
add a name generator util func to replace fmt.Sprintf

0x2d3c:


jinzhu:
Hi,

Thank you for your PR, but I don't think this is a better implementation, it seems make the logic more complicated without provide much value.

Thank you.

PR #3579, opened 2020-10-06
Added support Tabler interface for cached schemas

sah4ez:
- [x] Do only one thing
- [x] Tested

Added support Tabler interface for stored schemas in the cache.

### User Case Description
Dynamically table name filling.


jinzhu:
We don't want to support dynamically table name by design, use `Scopes` if has similar requirements.

Issue #3604, opened 2020-10-14
nil point transfer '<nil>' not transfer NULL 


blakejia:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/179

## Description

``` golang

type UserWithUUID struct {
	ID        uint `gorm:"primarykey"`
	CreatedAt time.Time
	UpdatedAt time.Time

	TestID     *uuid.UUID `gorm:"type:uuid"`
}

value := UserWithUUID{}

createErr := DB.Save(&value).Error

```

sql is 
``` bash
INSERT INTO "user_with_uuids" ("created_at","updated_at","test_id") VALUES ('2020-10-14 02:20:47.3','2020-10-14 02:20:47.3','<nil>') RETURNING "id"
```

```TestID``` nil point transfer ```<nil>```

**Comment by tebrizetayi** [comment]
I would like to take this issue to work on it!

**Comment by jinzhu** [comment]
> I would like to take this issue to work on it!

thank you, most likely it is a logger format issue.

**Comment by tebrizetayi** [comment]
What was expectation @blakejia ?

**Comment by blakejia** [comment]
> What was expectation @blakejia ?

sql is
``` sql
INSERT INTO "user_with_uuids" ("created_at","updated_at","test_id") VALUES ('2020-10-14 02:20:47.3','2020-10-14 02:20:47.3','<nil>') RETURNING "id"
```

this is should be
``` sql
INSERT INTO "user_with_uuids" ("created_at","updated_at","test_id") VALUES ('2020-10-14 02:20:47.3','2020-10-14 02:20:47.3', NULL) RETURNING "id"
```

tebrizetayi:
if type is fmt.Stringer() and  fmt.Sprintf("%v", v) is used , it does not trigger the Stringer() function  of the object.That is why it gives '<nil>' not NULL as we wanted. The case fmt.Stringer is removed from the switch and pushed.

PR #3626, opened 2020-10-17
nil point transfer '<nil>' not transfer NULL #3604

tebrizetayi:
- [x] Do only one thing
- [x] Tested

if the field is null, then  not "<nil>",  but "NULL" is written.

if type is fmt.Stringer() and fmt.Sprintf("%v", v) is used , it does not trigger the Stringer() function of the object.That is why it gives '' not NULL as we wanted. The case fmt.Stringer is removed from the switch and pushed.

### User Case Description
```
type UserWithUUID struct {
	CreatedAt time.Time
	UpdatedAt time.Time

	TestID     *uuid.UUID `gorm:"type:uuid"`
}

value := UserWithUUID{}

createErr := DB.Save(&value).Error

```
Before :
INSERT INTO "user_with_uuids" ("created_at","updated_at","test_id") VALUES ('2020-10-14 02:20:47.3','2020-10-14 02:20:47.3','<nil>') 
Now
INSERT INTO `user_with_uuids` (`created_at`,`updated_at`,`test_id`) VALUES ("2020-10-18 00:20:22.765","2020-10-18 00:20:22.765",NULL)

### DIFF
--- a/logger/sql.go
+++ b/logger/sql.go
@@ -48,8 +48,6 @@ func ExplainSQL(sql string, numericPlaceholder *regexp.Regexp, escaper string, a
 			} else {
 				vars[idx] = "NULL"
 			}
-		case fmt.Stringer:
-			vars[idx] = escaper + strings.Replace(fmt.Sprintf("%v", v), escaper, "\\"+escaper, -1) + escaper
 		case driver.Valuer:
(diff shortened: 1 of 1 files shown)


tebrizetayi:
@jinzhu pull request is ready!

Issue #3530, opened 2020-09-25
DO NOT use SELECT * in SQL queries


luisgomez29:
## Describe the feature

Hi, I have been going through the documentation at https://gorm.io/docs/query.html and found that the queries use SELECT * which is bad practice.

```go
// Get the first record ordered by primary key
db.First (& user)
// SELECT * FROM users ORDER BY id LIMIT 1;
```

Assuming we have a user struct with the following fields:

```go
type User struct {
	ID        uint32    `gorm:"type:serial;primaryKey;auto_increment" json:"id"`
	Username string    `gorm:"not null;size:60;unique" json:"username"`
	Email     string    `gorm:"not null;size:60;unique" json:"email"`
}
```
The above should run a query:

`SELECT "user"."id", "user"."username", "user"."email" FROM "users" AS "user" ORDER BY "user"."id" LIMIT 1;`

## Motivation
We should not use SELECT * in the SQL query anymore. It's always better to use the explicit column list in the SELECT query than a * wildcard. It not only improves the performance but also makes your code more explicit.

jinzhu:
https://gorm.io/docs/advanced_query.html#Smart-Select-Fields

PR #3731, opened 2020-11-09
Select all fields in SQL queries avoiding the SELECT * FROM

luisgomez29:
- [x] Do only one thing
- [x] Tested

Select all the fields of the structures to do the SQL query avoiding SELECT * FROM. It's always better to use the explicit column list in the SELECT query than a * wildcard, #3530

### User Case Description
Queries for the `First`, `Take`, `Last`, `Find` methods:
```go
var users []*models.User
db.Debug().Model(&models.User{}).Find(&users)
// SELECT "id","first_name","last_name","email","password","created_at","updated_at" FROM "users"

db.Debug().Find(&users)
// SELECT "id","first_name","last_name","email","password","created_at","updated_at" FROM "users"
```
#### Smart Select Fields - Advanced Query
https://gorm.io/docs/advanced_query.html#Smart-Select-Fields

```go
type APIUser struct {
	FirstName string
	LastName  string
	Email     string
}

db.Debug().Model(&models.User{}).Find(&APIUser{})
// SELECT "first_name","last_name","email" FROM "users"
```

#### Special case

The `Find` method allows the result to be scanned to `map[string]interface{}` or `[]map[string]interface{}`, so the model fields are not selected for compatibility with the `Create From Map` option of the `Create` method.

```go
var results []map[string]interface{}
db.Debug().Model(&models.User{})Find(&results)
// SELECT * FROM "users"
```

### DIFF
--- a/callbacks/query.go
+++ b/callbacks/query.go
@@ -78,7 +78,7 @@ func BuildQuerySQL(db *gorm.DB) {
 				clauseSelect.Columns = make([]clause.Column, len(stmt.Schema.DBNames))
 
 				for idx, dbName := range stmt.Schema.DBNames {
-					clauseSelect.Columns[idx] = clause.Column{Name: dbName}
+					clauseSelect.Columns[idx] = clause.Column{Table: db.Statement.Table, Name: dbName}
 				}
 			}
 		}
--- a/callbacks/query.go
+++ b/callbacks/query.go
@@ -72,23 +72,13 @@ func BuildQuerySQL(db *gorm.DB) {
 				}
 			}
 		} else if db.Statement.Schema != nil && db.Statement.ReflectValue.IsValid() {
-			smallerStruct := false
-			switch db.Statement.ReflectValue.Kind() {
-			case reflect.Struct:
-				smallerStruct = db.Statement.ReflectValue.Type() != db.Statement.Schema.ModelType
-			case reflect.Slice:
-				smallerStruct = db.Statement.ReflectValue.Type().Elem() != db.Statement.Schema.ModelType
-			}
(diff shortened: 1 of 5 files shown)


jinzhu:
Hello @luisgomez29 

Thank you for your pull request, after few days thought, I think it might better to don't accept this breaking changes, which also might cause errors for `Joins`

But the feature sounds reasonable, maybe we can have a Session Mode for it.

Any suggestions?

luisgomez29:
@jinzhu Ok, however I think this improvement can be included in a later version.
Also I think the SQL query could be returned like this to avoid errors with the joins:
```go
db.Debug().Model(&models.User{}).Find(&users)
// SELECT "user"."first_name", "user"."last_name", "user"."email" FROM "users"
```
I will update the branch with the aforementioned changes in case you want to make improvements.

jinzhu:
Hi,

If a column doesn't exists in the table, which will broken the query also, so I think this is a breaking changes we should not introduce by default.

Maybe we can add a new mode for `Session` to support it.

luisgomez29:
@jinzhu Hi,
The `QueryFields` session mode was added.
`QueryFields` executes the SQL query with all fields of the table:


```go
// globally mode, all SQL queries will executes with all fields of the table
db, err := gorm.Open(postgres.Open(config.DBURL), &gorm.Config{
    QueryFields: true,
})

//sesion mode
tx := db.conn.Session(&gorm.Session{QueryFields: true})
tx.Debug().Model(&models.User{}).Find(&users)

// SELECT "user"."first_name", "user"."last_name", "user"."email" FROM "users"
```

jinzhu:
Thank you for your pull request!

PR #3753, opened 2020-11-16
Add hooks after completion of Migrate Value

kainonly:
- [ ] Do only one thing
- [x] Tested

Increase the dialect of the driver, add hooks after completion of Migrate Value

### User Case Description

for example in the column comment of postgresql

```golang
func (m Migrator) After(value interface{}) error {
	return m.RunWithValue(value, func(stmt *gorm.Statement) error {
		for _, field := range stmt.Schema.FieldsByDBName {
			if field.Comment != "" {
				if err := m.DB.Exec(
					"COMMENT ON COLUMN " + stmt.Table + "." + field.DBName + " is " + field.Comment,
				).Error; err != nil {
					return err
				}
			}
		}
		return nil
	})
}
```

### DIFF
--- a/migrator.go
+++ b/migrator.go
@@ -33,6 +33,7 @@ type ColumnType interface {
 type Migrator interface {
 	// AutoMigrate
 	AutoMigrate(dst ...interface{}) error
+	After(dst interface{}) error
 
 	// Database
 	CurrentDatabase() string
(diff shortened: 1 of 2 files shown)


jinzhu:
Hello @kainonly 

Thank you for your PR, but I don't think this is a good implementation, I would suggest you create a pull request for the postgres driver and reimplement the `CreateTable` method to support it.

BTW, when implementing the method, you could just call the current one by using `m.Migrator.CreateTable()` and create comments after the method

Thank you for your PR.

PR #3766, opened 2020-11-19
Fix Association.Replace() error returning

Deviller:
- [x] Do only one thing
- [x] Tested

`Assocoation.Replace()` does not correctly returns error in some switch branches: `schema.HasOne, schema.HasMany and schema.Many2Many`

### User Case Description

Create such database structure:

```
type Profile struct {
	ID       uint
	Number   string
	MemberID uint `gorm:"not null"`
}

type Member struct {
	ID       uint
	Profiles []Profile
}
```

Insert some data:

```
member := &Member{
	Profiles: []Profile{{
		Number: "1",
	}, {
		Number: "2",
	}},
}
DB.Create(&member)
```

And try to `DB.Model(member).Association("Profiles").Clear()`. No error will be returned but the error actually occurs if you enable the logger


jinzhu:
thank you for your PR

PR #3794, opened 2020-11-26
test: no cache preload in goroutine

zhaozong:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


zhaozong:
@jinzhu 看一下这个case，没有缓存relations的情况下goroutine中使用preload会报错

PR #3800, opened 2020-11-27
update 支持postgres RETURNING 语法

akyowan:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


jinzhu:
driver specific features should only in the driver's codebase

PR #3818, opened 2020-12-02
Fixed creation of associated records with composite primary keys (#3817)

### LINKED ISSUE #3817: Creation of associated records fails on Postgres and SQLite when using composite primary keys
## GORM Playground Link

https://github.com/go-gorm/playground/pull/218

## Description

Using composite primary keys, I get error trying to create a record that has many associated records. I expect the main and associated records to be created without errors.

Postgres error:
> ERROR: there is no unique or exclusion constraint matching the ON CONFLICT specification (SQLSTATE 42P10)

SQLite error:
> ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint

MySQL and MS SQL Server are not affected.

baibaratsky:
- [x] Do only one thing
- [x] Tested

Fixed #3817.

### User Case Description

I’m not sure why did you use `PrioritizedPrimaryField` here, but it causes issues with composite keys, so I removed it.

### DIFF
--- a/callbacks/associations.go
+++ b/callbacks/associations.go
@@ -318,12 +318,8 @@ func onConflictOption(stmt *gorm.Statement, s *schema.Schema, selectColumns map[
 
 	if len(defaultUpdatingColumns) > 0 {
 		var columns []clause.Column
-		if s.PrioritizedPrimaryField != nil {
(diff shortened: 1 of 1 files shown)


jinzhu:
thank you for your PR

PR #3825, opened 2020-12-03
Fix schema initialization paths

abursavich:
I was reading the latest release diff and noticed an issue with schema initialization. I split this PR into two commits: the first fixes bugs and the second rearranges code to be less deeply nested.

The initialized channel was only closed if the schema's cacheStore did not contain the embeddedCacheKey and there were no errors parsing relations.  If the key existed or an error occurred, it would not be closed. This could leave other goroutines waiting for synchronization that will never occur.

Additionally, the other goroutines that wait for initialization to complete did not return the possible error.

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -92,7 +92,7 @@ func Parse(dest interface{}, cacheStore *sync.Map, namer Namer) (*Schema, error)
 	if v, ok := cacheStore.Load(modelType); ok {
 		s := v.(*Schema)
 		<-s.initialized
-		return s, nil
+		return s, s.err
 	}
 
 	modelValue := reflect.New(modelType)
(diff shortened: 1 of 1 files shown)


abursavich:
I added a commit that replaces calls to `getOrParse` with calls to `Parse` because `getOrParse` wasn't waiting for initialization/errors, but I reverted it because I'm not sure what kind of cyclical schema relation dependencies may exist that might lead to deadlocking.  Using getOrParse as-is is problematic, but the fix may not be so easy.

jinzhu:
Thank you for your PR.

jinzhu:
> I added a commit that replaces calls to getOrParse with calls to Parse because getOrParse wasn't waiting for initialization/errors, but I reverted it because I'm not sure what kind of cyclical schema relation dependencies may exist that might lead to deadlocking. Using getOrParse as-is is problematic, but the fix may not be so easy.

getOrParse is intend to don't wait, which is necessary for self-reference struct for example:

```go
type User struct {
  xxx
  Friends []User
}
```

PR #3875, opened 2020-12-21
Use tablename from provided model when querying WHERE with a struct a…

lvbesien:
Ensures that queries using `Where` with a struct argument will formulate SQL with the correct table name


jinzhu:
Some tests failed, so not good to merge the breaking change.

PR #3876, opened 2020-12-21
add RollbackUnlessComitted, to make it easy to defer a Rollback.

rubensayshi:
- [x] Do only one thing
- [x] Tested

Adding `RollbackUnlessComitted` back in (which was also in gorm 1), so you can defer a call to that and don't have to worry about not rolling back a TX and if you do commit the TX then you won't get a weird error being logged about the TX already being comitted.

People are also asking for it on the docs page; https://gorm.io/docs/transactions.html

### User Case Description

```go
func MyAction() error {
    tx := DB.Begin()
    defer tx.RollbackUnlessComitted()

    if something {
        return errors.New()
    }

    if somethingElse {
        return errors.New()
    }

    if anotherThing {
        return errors.New()
    }

    tx.Commit()

    return nil 
}
```

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -584,6 +584,21 @@ func (db *DB) Rollback() *DB {
 	return db
 }
 
+// RollbackUnlessComitted rollback a transaction unless it has already been comitted
+func (db *DB) RollbackUnlessComitted() *DB {
+	if committer, ok := db.Statement.ConnPool.(TxCommitter); ok && committer != nil {
+		if !reflect.ValueOf(committer).IsNil() {
+			err := committer.Rollback()
+			if err != nil && err != sql.ErrTxDone {
+				db.AddError(err)
+			}
+		}
+	} else {
+		db.AddError(ErrInvalidTransaction)
+	}
+	return db
+}
+
 func (db *DB) SavePoint(name string) *DB {
 	if savePointer, ok := db.Dialector.(SavePointerDialectorInterface); ok {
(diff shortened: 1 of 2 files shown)


rubensayshi:
I don't quite understand why the pipeline fails, I don't think my changes have anything to do with that...

jinzhu:
Hi, thank you for your PR, but not going to accept it, suggest everyone use the `Transaction` method but not `Begin`, `Commit`, `Rollback`...

rubensayshi:
`Transaction` doesn't work well with multiple return values :(

rubensayshi:
@jinzhu it would really help people who don't want to use the `Transaction` and it comes at a very minimal overhead for the project to add the 1 method

jinzhu:
You can wrap another method for your multiple return case if it is necessary (but I think this is still handable with `Transaction` method)

For your code,  you should add error check for the `Begin` like following code or it might cause panic in high load.

```go
    tx := DB.Begin()
    if tx.Error != nil {
    }

    defer tx.RollbackUnlessComitted()
```

> it comes at a very minimal overhead for the project to add the 1 method

I don't think this is a good practice, and don't want to encourage users use it by adding the method, it might cause more issues like the above panic one.

rubensayshi:
without the defer you risk forgetting a `tx.Rollback()` in one of your `if err != nil {}` blocks in your function ...

jinzhu:
The `Begin` method possible to returns an error, in this case, don't need to `Rollback`, please check `Transaction`'s implemention.

Issue #3891, opened 2020-12-28
MergeClause method of set doesn't work as merge but replace actually.


sdghchj:
## GORM Playground Link

```
type User struct {
	gorm.Model
	Name string `gorm:"unique index"`
}

func (user *User ) BeforeDelete(db *gorm.DB) error {
	delTime := fmt.Sprintf(`_delete_at_%d`, time.Now().Unix())
	db.Statement.AddClause(clause.Set{clause.Assignment{
		Column: clause.Column{Name:"name"},
		Value:  gorm.Expr(`concat(name,?)`, delTime),
	})
	return nil
}
```

As field `Name` has the unique index, to make it possible that  another user with the same name is able to be inserted after a user record is deleted softly, I need to implement the BeforeDelete method. But it does not work.

## Description

code [here](https://github.com/go-gorm/gorm/blob/master/clause/set.go#L34)
```
// MergeClause merge assignments clauses
func (set Set) MergeClause(clause *Clause) {
	copiedAssignments := make([]Assignment, len(set))
	copy(copiedAssignments, set)
	clause.Expression = Set(copiedAssignments)
}
```

PR #3893, opened 2020-12-28
Correct the implementation of MergeClause method of Set

sdghchj:
- [x] Do only one thing
- [x] Tested

Correct the implementation of MergeClause method of Set

### User Case Description

#3891

### DIFF
--- a/clause/clause_test.go
+++ b/clause/clause_test.go
@@ -38,6 +38,6 @@ func checkBuildClauses(t *testing.T, clauses []clause.Interface, result string, 
 	}
 
 	if !reflect.DeepEqual(stmt.Vars, vars) {
-		t.Errorf("Vars expects %+v got %v", stmt.Vars, vars)
+		t.Errorf("Vars expects %+v got %v", vars, stmt.Vars)
 	}
 }
--- a/clause/set.go
+++ b/clause/set.go
@@ -32,9 +32,14 @@ func (set Set) Build(builder Builder) {
 
 // MergeClause merge assignments clauses
 func (set Set) MergeClause(clause *Clause) {
-	copiedAssignments := make([]Assignment, len(set))
-	copy(copiedAssignments, set)
-	clause.Expression = Set(copiedAssignments)
+	if clause.Expression == nil {
+		copiedAssignments := make([]Assignment, len(set))
+		copy(copiedAssignments, set)
+		clause.Expression = Set(copiedAssignments)
+	} else if copiedAssignments, ok := clause.Expression.(Set); ok {
+		copiedAssignments = append(copiedAssignments, set...)
+		clause.Expression = Set(copiedAssignments)
+	}
 }
 
 func Assignments(values map[string]interface{}) Set {
--- a/clause/set_test.go
+++ b/clause/set_test.go
@@ -55,3 +55,25 @@ func TestAssignments(t *testing.T) {
 		t.Errorf("invalid assignments, got %v", assignments)
 	}
 }
+
+func TestMergeClause(t *testing.T) {
+	set1 := clause.Assignments(map[string]interface{}{
+		"name": "jinzhu",
+	})
+	set2 := clause.Assignments(map[string]interface{}{
+		"age": 18,
+	})
+
+	c := &clause.Clause{
+		Name:       "Set",
+		Expression: set1,
+	}
+	set2.MergeClause(c)
+
+	if set3, ok := c.Expression.(clause.Set); ok {
+		if len(set3) == 2 {
+			return
+		}
+	}
+	t.Error("invalid merge set")
+}
--- a/clause/set_test.go
+++ b/clause/set_test.go
@@ -28,7 +28,7 @@ func TestSet(t *testing.T) {
 				clause.Set([]clause.Assignment{{clause.PrimaryColumn, 1}}),
 				clause.Set([]clause.Assignment{{clause.Column{Name: "name"}, "jinzhu"}}),
 			},
-			"UPDATE `users` SET `name`=?", []interface{}{"jinzhu"},
+			"UPDATE `users` SET `users`.`id`=?,`name`=?", []interface{}{1, "jinzhu"},
 		},
 	}
 


sdghchj:
@jinzhu

jinzhu:
`Set` expects to replace but don't append, it might cause issues if have multiple same columns.

For your issue, you can use OnConflict DoUpdates to implement it.

diabloneo:
Hi, @jinzhu 

I think the **non-replacing-set-clause** is required when someone want to implement their callbacks.

Take this model as an example:

```
type User struct {
    gorm.Model
    Name string `gorm:"uniqueIndex;unique"`
}
```

It will be soft deleted when called with `db.Delete(user)` as we expected. The problem comes when we want to insert a new `user` with same name, the unique index will complain the duplication of the **name** key. So, I thought why not appending some random string to the name key to solve this problem? It can be implemented by adding a `AfterDelete` method to update the `name` field after it was soft deleted. But the hook method has two problems:

1. You have to add it on every model manually or by embedding some base model.
2. The record will be updated twice.

As GORM allows adding customize callbacks to the callbacks chains. It will be elegant and effective to implement the auto-appending scheme above.  But it does not work due to the `Set` clause's `MergeClause` method replacing the clause.

Here is the demo callback:

```

func someInitFunc() {
	basicDB.Callback().Delete().Before("gorm:delete").
		Register("update_unique_name", callbackDeleteUpdateUniqueName)
}

func callbackDeleteUpdateUniqueName(db *gorm.DB) {
	if db.Statement.Schema == nil || db.Statement.Unscoped {
		return
	}

	nameField := db.Statement.Schema.LookUpField(FieldName)
	if nameField == nil {
		return
	}
	if nameField.DataType != schema.String {
		return
	}
	if !nameField.Unique {
		return
	}

	name, _ := nameField.ValueOf(db.Statement.ReflectValue)
	newName := fmt.Sprintf("%s-%d", name, rand.Int63())
	fmt.Printf("newname: %s\n", newName)
	db.Statement.AddClause(clause.Set{{Column: clause.Column{Name: nameField.DBName}, Value: newName}})
	db.Statement.SetColumn(FieldName, newName, true)
}
```

But the `Set` clause will be replaced in `func (sd SoftDeleteDeleteClause) ModifyStatement(stmt *Statement)`:

```
func (sd SoftDeleteDeleteClause) ModifyStatement(stmt *Statement) {
	if stmt.SQL.String() == "" {
		curTime := stmt.DB.NowFunc()
                // The Set clause will be replaced here.
		stmt.AddClause(clause.Set{{Column: clause.Column{Name: sd.Field.DBName}, Value: curTime}})
		stmt.SetColumn(sd.Field.DBName, curTime, true)
...
```

omaskery:
I have also just encountered this issue whilst experimenting with writing an optimistic-locking utility for for GORM.

My approach involves a `Versioned` struct in your model struct with a `Version` field:
```go
// Versioned can be embedded in a GORM model to add optimistic locking
type Versioned struct {
	Version uint64 `gorm:"not null;default:1;"`
}
```

In the `BeforeUpdate`/`BeforeDelete` hooks it sets the `Version` field to be `version + 1`, and adds a `WHERE` clause ensuring that the update/delete applies to rows whose `version` column is equal to the old version number. E.g.:
```go
// BeforeUpdate ensures that updates to a Versioned model only apply if there has not been a concurrent modification,
// detected through an optimistic lock version, and asserts that the new object will have a new version
func (v *Versioned) BeforeUpdate(tx *gorm.DB) error {
	return v.assertLockValidity(tx)
}

// BeforeDelete ensures that deleting a Versioned model only applies if there has not been a concurrent modification,
// detected through an optimistic lock version, and asserts that the deleted object will have a new version
func (v *Versioned) BeforeDelete(tx *gorm.DB) error {
	return v.assertLockValidity(tx)
}

func (v *Versioned) assertLockValidity(tx *gorm.DB) error {
	readVersion := v.Version
	v.Version = readVersion + 1

	tx.Statement.Where("version = ?", readVersion)
	tx.Statement.AddClause(clause.Set{{Column: clause.Column{Name: "version"}, Value: v.Version}})

	return nil
}
```

This works fine for the `BeforeUpdate` call, but in `BeforeDelete` - as described by @diabloneo:
* GORM's built-in `DeletedAt` type for soft-deletion adds a `SET` clause to set its `deleted_at` column.
* GORM silently replaces existing `SET` clauses when asked to "add" more.

This means that:
* If you use GORM's `DeletedAt` type you _cannot_ add `SET` clauses in a `BeforeDelete` hook.
* And more generally: _no two `BeforeDelete` hooks_ can _ever_ add `SET` clauses.

> Set expects to replace but don't append, it might cause issues if have multiple same columns.

@jinzhu if your concern is overlapping assignments, would it make sense to at least allow merging of non-overlapping columns, and continue to have "replace with latest" behaviour where columns overlap?

E.g.:
1. A set clause that sets A=1, B=2 results in `SET A=1, B=2`
2. The next set clause sets B=3, C=4 results in `SET A=1, B=3, C=4`
   * Note that the original write to column A is preserved, as the second clause didn't reference column A
   * Note that the write to column B is replaced, as the second clause also wants to set column B
   * Note that the write to column C "just works", as it does not overlap with any previous clauses

The code for which could be something like:
```go
func (set Set) MergeClause(clause *Clause) {
	existingAssignments, ok := clause.Expression.(Set)
	
	// if the existing clause is nil or not a set, then fallback to simply replacing
	// the existing clause with our assignments
	if !ok {
		copiedAssignments := make([]Assignment, len(set))
		copy(copiedAssignments, set)
		clause.Expression = Set(copiedAssignments)
		return
	}
	
	// attempt to merge the existing assignments, overwriting assignments where
	// the referenced column is the same. The "most recently added" clause "wins".
	assignmentSet := map[Column]Assignment{}
	for _, a := range existingAssignments {
		assignmentSet[a.Column] = a
	}
	for _, a := range set {
		assignmentSet[a.Column] = a
	}
	mergedAssignments := make([]Assignment, 0, len(assignmentSet))
	for _, a := range assignmentSet {
		mergedAssignments = append(mergedAssignments, a)
	}
	clause.Expression = Set(mergedAssignments)
}
```

Is this a palatable solution? Would you accept a PR with something along these lines (plus relevant unit tests)? 🙂

PR #3931, opened 2021-01-10
use NowFunc replace time.Now

0x2d3c:


jinzhu:
Hi, the begin time here is used to calc the SQL execution time here, https://github.com/go-gorm/gorm/blob/b864a5457a59ddfca3dae0f6b11de7443633392b/logger/logger.go#L139

so it can't use the `NowFunc`, thank you for your PR.

PR #3940, opened 2021-01-12
[Relationship] Title Case schema name for many2many

lisamryl:
- [X] Do only one thing
- [X] Tested

Ensures that schema.Name is Title case before building references in buildMany2ManyRelation. 

### User Case Description
We were trying to make a many2many relation gorm tags for a field in our db model, and kept failing due to `"field" is unexported but missing PkgPath`. It seems the schema.Name for our fields were not title case, which blocked us from proceeding to build any further.

In the screenshot, the first 2 schema names already had .Title applied (from the first line of the fix), but as you can see, the 3rd is still lowercase, which lead to the error shown, once we got further along in the codepath.

Screen Shot 2021-01-12 at 12 08 48 PM

### DIFF
--- a/schema/relationship.go
+++ b/schema/relationship.go
@@ -219,7 +219,7 @@ func (schema *Schema) buildMany2ManyRelation(relation *Relationship, field *Fiel
 	}
 
 	for idx, ownField := range ownForeignFields {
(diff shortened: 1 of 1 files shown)


jinzhu:
Thank you for your PR.

PR #3943, opened 2021-01-13
extract common parts in logger and optimize code

0x2d3c:


logger/logger.go
```diff
@@ -105,6 +105,21 @@ type logger struct {
 	traceStr, traceErrStr, traceWarnStr string
 }
 
+func (l *logger) printf(ctx context.Context, str, msg string, level LogLevel, data ...interface{}) {
```
jinzhu, line 108: I think this extract is not necessary, it doesn't make things more clear, which add another func call and slow down the `FileWithLineNum`

logger/logger.go
```diff
@@ -114,52 +129,31 @@ func (l *logger) LogMode(level LogLevel) Interface {
...
-	}
+	l.printf(ctx, l.errStr, msg, Error, data...)
 }
 
 // Trace print sql message
 func (l logger) Trace(ctx context.Context, begin time.Time, fc func() (string, int64), err error) {
 	if l.LogLevel > Silent {
 		elapsed := time.Since(begin)
 		switch {
+		case l.LogLevel == Info:
```
jinzhu, line 150: This case should be the last case, and maybe the method maybe better refactor like:

```go
switch {
case: ...
  ft = xxx;
  trace = xxx;
default:
  return
}

sql, rows := fc()
if rows == -1 {
   l.Printf(ft, utils.FileWithLineNum(), trace, float64(used.Nanoseconds())/1e6, "-", sql)
} else {
  l.Printf(ft, utils.FileWithLineNum(), trace, float64(used.Nanoseconds())/1e6, rows, sql)
}
```

0x2d3c: I have not considered it, and the number of formatting parameters of Trace has not been considered.

PR #3948, opened 2021-01-15
fix error formatting

jrapoport:
config.Logger.Error does not want a format string.

- [] Do only one thing
- [] Non breaking API changes
- [] Tested


jinzhu:
I don't get why make this change, we are using `Printf` to format the message? https://github.com/go-gorm/gorm/blob/59fa07953cf43385587677f106bb5e522621dca1/logger/logger.go#L132

jrapoport:
it was an issue with a logrus adapter i was using. was tired, sry.

PR #3965, opened 2021-01-18
embedded field relationship schema do not contain relations

### LINKED ISSUE #3964: Schema in Relationship is different from the model
## GORM Playground Link

https://github.com/go-gorm/playground/pull/252

## Description

Due to the schema mismatch, the preload  will fail, because schema on Relationship do not have Relations

wenerme:
- [x] Do only one thing
- [x] Tested

add current schema to cacheStore when ParseField

### User Case Description
fix #3964

### DIFF
--- a/schema/field.go
+++ b/schema/field.go
@@ -7,7 +7,6 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/jinzhu/now"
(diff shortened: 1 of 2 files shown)


wenerme:
works for two levels, deeper relation still not exists.

wenerme:
also tested against https://github.com/go-gorm/playground/pull/252

wenerme:
this fix only works for small entity graph, large entity graph will cause supper slow parse, because too many dup fields to parse.

jinzhu:
Fixed, thank you for your report and PR

Issue #3363, opened 2020-09-02
NULL default values in v2


flusflas:
I'm trying to migrate to v2 and I'm having problems with a string field that should default to NULL if it is empty. In v1 this worked:

```go
type Invitation struct {
	ID	string	`json:"id"`
	Value	string	`json:"value" gorm:"default:null"`
}
```

In v2, using the same type, when I save a struct with empty `to_user`, it is set to `"null"` (a string with "null" value...). Is this intended? Has GORM dropped support for NULL default values?

I've read about using pointers or `sql.NullString`, but this would deeply break my code.

## Reproduction steps

```go
package main

import (
	"fmt"
	"gorm.io/driver/postgres"
	"gorm.io/gorm"
)

type Foo struct {
	ID    string `json:"id"`
	Value string `json:"value" gorm:"default:null"`
}

const DSN = "postgres://user:pass@127.0.0.1:5432/db?sslmode=disable"

func main() {
	db, _ := gorm.Open(postgres.New(postgres.Config{DSN: DSN}), &gorm.Config{})
	db.AutoMigrate(&Foo{})

	foo := &Foo{
		ID:    "001",
		Value: "",
	}

	err := db.Create(foo).Error
	if err != nil {
		fmt.Printf("Error: %s", err.Error())
	}

	db.Model(foo).First(foo)
	fmt.Printf("foo.Value -> %q", foo.Value)

	db.Delete(foo)
}
```

This should store in database a record with `value = "null"`.

flusflas:
Just added a simple code to reproduce the issue.

flusflas:
I checked my code with the new changes and now it is raising `Error: sql: Scan error on column index 0, name "value": converting NULL to string is unsupported`. I assume this is happening when the struct is being reloaded after creating the record.

jeffwalsh:
The functionality is still broken with a `sql.NullString`. It inserts a string of "null".

Bexanderthebex:
@flusflas you can use a string pointer

annlumia:
use `sql.NullString` without set default value is worked.

`

	gorm.io/driver/postgres v1.3.1
	gorm.io/driver/sqlite v1.3.1
	gorm.io/gorm v1.23.3
	
`

PR #4028, opened 2021-01-31
Add field tag to ignore migration

yrong:
- [x] Do only one thing
- [x] Tested

### User Case Description

```
import (
	"github.com/shopspring/decimal"
       ...
)

type ChainExtrinsic struct {
Fee                decimal.Decimal `json:"fee" sql:"type:decimal(30,0);"`//v1
...
}
type ChainExtrinsic struct {
Fee                decimal.Decimal `json:"fee" gorm:"type:decimal(30,0);"`//v2
...
}
db.AutoMigrate()

/*Actually there is no field definition change between v1 and v2.
for mysql migration from gorm v1 to v2 the unnessary alter column ddl as following will be generated
the worst thing is after migrating to v2, will still generate ddl each time when run automigrate
---ALTER TABLE `chain_extrinsics` MODIFY COLUMN `fee` decimal(30,0)
*/

```
we can avoid by add field tag as following:

```
type ChainExtrinsic struct {
Fee                decimal.Decimal `json:"fee" gorm:"type:decimal(30,0);ignoreMigration"`
...
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ TODO*
 documents
 coverage.txt
 _book
+.idea
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -396,7 +396,7 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 		}
 	}
 
-	if alterColumn {
+	if alterColumn && !field.IgnoreMigration {
 		return m.DB.Migrator().AlterColumn(value, field.Name)
 	}
 
--- a/schema/field.go
+++ b/schema/field.go
@@ -70,6 +70,7 @@ type Field struct {
 	ReflectValueOf         func(reflect.Value) reflect.Value
 	ValueOf                func(reflect.Value) (value interface{}, zero bool)
 	Set                    func(reflect.Value, interface{}) error
+	IgnoreMigration        bool
 }
 
 func (schema *Schema) ParseField(fieldStruct reflect.StructField) *Field {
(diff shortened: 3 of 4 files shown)


jinzhu:
Hello @yrong 

For you example, I would suggest you add a `precision` tag like:

```go
type ChainExtrinsic struct {
  Fee                decimal.Decimal `json:"fee" gorm:"type:decimal(30,0);precision:30"`
  ...
}
```

but the skip migration feature sounds interesting, maybe we could support it like following, would you mind to make the change?

```go
type ChainExtrinsic struct {
  Fee                decimal.Decimal `json:"fee" gorm:"type:decimal(30,0);-:migration"`
  ...
}
```

yrong:
@jinzhu 
Thanks for the clarifaction about `precision` tag and will take a try.  For the skip migration I think maybe there is some confliction with the exist Field-Level Permission Definition which is essentially different and better not mixed together
`Name string `gorm:"-"`  // ignore this field when write and read with struct`

yrong:
@jinzhu 
Btw, I add one minor fix in [4a1096b](https://github.com/go-gorm/gorm/pull/4028/commits/4a1096bcff22a37fb91fd86816c142d32b8eb513)  which I think maybe a bug because such model works in gorm v1 

`
Address           string          `gorm:"default: null;size:100" json:"address"`  //there is a leading space before null
`
maybe something related to #3363

jinzhu:
> @jinzhu
> Thanks for the clarifaction about `precision` tag and will take a try. For the skip migration I think maybe there is some confliction with the exist Field-Level Permission Definition which is essentially different and better not mixed together
> 
> `Name string `gorm:"-"` // ignore this field when write and read with struct`

I think it would be good to group all field permission related things into similar tags, so users can get them in one place.

Maybe we could have different permission level like the `->`, `<-` one, `-` means disable read/write by default, `-:all` means disable read/write & migration, `-:migration` means only disable the migration?

yrong:
@jinzhu 
some refactor per your request, pls check if anything missing

jinzhu:
Thank you for your PR

birdycn:
I want to Add field tag to ignore migration,too

What should I do now

PR #4030, opened 2021-01-31
feat: add functional option for Config and add OpenWithOption func for gorm DB.

daheige:
1.add functional option for Config
2.add OpenWithOption func for gorm DB when you create db entry.
3.gofmt code

### DIFF
--- a/association.go
+++ b/association.go
@@ -24,10 +24,12 @@ func (db *DB) Association(column string) *Association {
 
 	if err := db.Statement.Parse(db.Statement.Model); err == nil {
 		db.Statement.Table = table
-		association.Relationship = db.Statement.Schema.Relationships.Relations[column]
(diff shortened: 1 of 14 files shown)


daheige:
I have been using gorm for more than 4 years, and I feel that the code style of gorm does need to be improved. For example, the instantiation of this Config uses control flow like if in many places, which is really compared to the authentic code of go. low. Here I add a new function to use the functional function design pattern to make some default settings for Config. Gorm is a low-level library. The code is written for humans, not machines. The overall code does not do some gofmt operations and code comments. For example, the exported functions and public variables have no reasonable comments. The third aspect is that there are a lot of repetitive codes in gorm's code base. For example, when setting data using reflection, there are a lot of repetitive codes. I think these can be done better. You can use switch case to merge or extract small ones. Functional functions can be used.

PR #4032, opened 2021-01-31
add functional option for Config and add OpenWithOption func for gorm DB.

daheige:
1.add functional option for Config
2.add OpenWithOption func for gorm DB
3. go fmt code

### DIFF
--- a/association.go
+++ b/association.go
@@ -24,10 +24,12 @@ func (db *DB) Association(column string) *Association {
 
 	if err := db.Statement.Parse(db.Statement.Model); err == nil {
 		db.Statement.Table = table
-		association.Relationship = db.Statement.Schema.Relationships.Relations[column]
+		association.Relationship = db.Statement.Schema.
+			Relationships.Relations[column]
 
 		if association.Relationship == nil {
-			association.Error = fmt.Errorf("%w: %v", ErrUnsupportedRelation, column)
+			association.Error = fmt.Errorf("%w: %v",
+				ErrUnsupportedRelation, column)
 		}
 
 		db.Statement.ReflectValue = reflect.ValueOf(db.Statement.Model)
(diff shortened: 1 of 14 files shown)


daheige:
I have been using gorm for more than 4 years, and I feel that the code style of gorm does need to be improved. For example, the instantiation of this Config uses control flow like if in many places, which is really compared to the authentic code of go. low. Here I add a new function to use the functional function design pattern to make some default settings for Config. Gorm is a low-level library. The code is written for humans, not machines. The overall code does not do some gofmt operations and code comments. For example, the exported functions and public variables have no reasonable comments. The third aspect is that there are a lot of repetitive codes in gorm's code base. For example, when setting data using reflection, there are a lot of repetitive codes. I think these can be done better. You can use switch case to merge or extract small ones. Functional functions can be used. 

For code inspection and style, please refer to this https://github.com/golangci/golangci-lint and then deal with the code in gorm accordingly.

jinzhu:
sorry, although there are some good changes, I am not the fan of 80 chars limitation, don't think it make users easier to read in modern monitors...

daheige:
Looking at the overall Gorm V2 design, it is worrying about the quality and style of the code. Some methods after encapsulation really make people feel that the quality of the code needs to be improved. In particular, some configurations have a lot of repetitive code and a lot of large sections of code. It is recommended to make reasonable adjustments. On the one hand, most of the comments in components are not written. Whether the code is written for people, not for machines, we should fully consider the readability. Another is that a lot of code is written at a low level, and some can be organized by using function design mode, which is more suitable for upgrading and maintenance.

daheige:
See this post about the "functional options" pattern:
http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

daheige:
code style,please see: https://github.com/uber-go/guide/blob/master/style.md. You can take a look at this code specification. Of course, many big companies in the industry also have standard standards. If you really don't know the code specification, you can also refer to Google's code specification style, so as to better write authentic code.

jinzhu:
> See this post about the "functional options" pattern: http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

Although I use the  "functional options" pattern sometime, don't think it matches GORM's case better.

Use struct here, because I want users clearly known all options that he could use during initialize, compare to the issues that `functional options` try to resolve, this is more important from my understanding.

>  On the one hand, most of the comments in components are not written

Time costing/return value balance, from my GORM v1 experience, even documented all public methods, it doesn't give much value back. and yeah, contributions welcome, but just like the guide said: "Do only one thing" in one pull request.

PR #4039, opened 2021-02-02
Add NewDB as a Shortcut Call

vzool:
- [x] Do only one thing
- [x] Tested

Restore old New from v1 to be available in v2  as a shortcut call which is using new way of NewDB

### User Case Description

Current usage is very long, I used all over the application points.
```golang
// Current
db.Session(&gorm.Session{NewDB: true})
// Proposed
db.NewDB()
```

### DIFF
--- a/gorm.go
+++ b/gorm.go
@@ -248,6 +248,11 @@ func (db *DB) Debug() (tx *DB) {
 	})
 }
 
+// NewDB clone a new db connection without search conditions
+func (db *DB) NewDB() *DB {
+	return db.Session(&Session{NewDB: true})
+}
+
 // Set store value with key into current db instance's context
 func (db *DB) Set(key string, value interface{}) *DB {
 	tx := db.getInstance()


jinzhu:
sorry, but not going to merge.

jinzhu:
in most cases, this `db.Session(&gorm.Session{NewDB: true})` in not necessary, if you use it a lot, maybe there are something wrong in your code.

vzool:
> in most cases, this `db.Session(&gorm.Session{NewDB: true})` in not necessary, if you use it a lot, maybe there are something wrong in your code.

In fact, I disagree. Because, if some use it inside HTTP handler with one database connection shared across many handlers like this:

```golang
// AccountIndex Handler
func AccountIndex(db *gorm.DB) gin.HandlerFunc {
	return func(ctx *gin.Context) {
                db = db.New() // Clear old query conditions from last call
                // ...
        }
}
```
Always they will need to call "**New**" because all HTTP handlers are shared one database connection.
Even, there are some interconnected logic needs "**New**" function to be called as well.
I hope that you may reconsider this PR again.
Thanks for this nice library <3

jinzhu:
//  Clear old query conditions from last call

this is not necessary, check out https://gorm.io/docs/method_chaining.html#Method-Chain-Safety-Goroutine-Safety for details

PR #4061, opened 2021-02-06
Preload fix, when condition has LIMIT clause use UNION sql not IN (...)

junstyle:
```
type User struct {
	ID int
	Books []Book
}

type Book struct {
	ID int
	UserID int
	Name string
}
```

我只想取每个User的前三条Book的记录，要怎么实现？好像生成的sql语句是这样的：
```
select * from book where userID in (1, 2, 3, 4, 5)
```

```
tx.Preload("Books", func(db *gorm.DB) *gorm.DB {
	return db.Select("id, userID, name").Order("book.id DESC")
})
```

加上Limit也不行：
```
tx.Preload("Books", func(db *gorm.DB) *gorm.DB {
	return db.Select("id, userID, name").Order("book.id DESC").Limit(3)
})
```
然而生成的SQL是类似这样的，明显不对，总共只能取3条，期望是每个User取3条。
```
select * from book where userID in (1, 2, 3, 4, 5) **limit** 3
```

这个PR，当含有Limit时，生成union语句，类似这样的：
```
(select * from book where userID=1 limit 3) UNIOIN (select * from book where userID=2 limit 3) UNION (select * from book where userID=3 limit 3) UNIOIN (select * from book where userID=4 limit 3) UNION (select * from book where userID=5 limit 3)
```

### DIFF
--- a/callbacks/preload.go
+++ b/callbacks/preload.go
@@ -2,6 +2,7 @@ package callbacks
 
 import (
 	"reflect"
+	"strings"
 
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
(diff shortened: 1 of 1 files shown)


jinzhu:
Sorry, not going to merge it, as it doesn't works like what I expected.

But I am good if you changed it to something like, and change the SQL to something you like by changing db's Statement.

image

PR #4063, opened 2021-02-06
optimization check for ParseCheckConstraints

daheige:
ParseCheckConstraints func
match English letters and midline
Here, the regularity of matching English characters and midline does not have to be a regular compilation operation every time. At the bottom of go reg, the logic here is also consumed. For details, see the implementation of compile and compile of the bottom source code of go reg.

### DIFF
--- a/schema/check.go
+++ b/schema/check.go
@@ -5,6 +5,11 @@ import (
 	"strings"
 )
 
+var (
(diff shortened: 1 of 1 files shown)


daheige:
This failure may be a problem with the unit test writing.

PR #4064, opened 2021-02-06
feat: Optimize loop assignment

daheige:
Optimize loop assignment for statement.go#441 map loop assignment.


daheige:
This failure may be a problem with the unit test writing.

jinzhu:
the changes won't work, the original czgz possible to be changed by the cloned one.

PR #4078, opened 2021-02-10
feat: define short function name easy to use

lockp111:
define short function name easy to use "Scopes"

#### before:
```golang
func WithUsername(name string) func(db *gorm.DB) *gorm.DB) error {
	return func(db *gorm.DB) *gorm.DB {
		return db.Where("username = ?", name)
	}
}

func DeleteUser(scope ...func(db *gorm.DB) *gorm.DB) error {
	return sysDB.
		Scopes(scope...).
		Delete(&model.User{}).
		Error
}
```

#### after:
```golang
func WithUsername(name string) gorm.Scope {
	return func(db *gorm.DB) *gorm.DB {
		return db.Where("username = ?", name)
	}
}

func DeleteUser(scope ...gorm.Scope) error {
	return sysDB.
		Scopes(scope...).
		Delete(&model.User{}).
		Error
}
```

### DIFF
--- a/chainable_api.go
+++ b/chainable_api.go
@@ -9,6 +9,9 @@ import (
 	"gorm.io/gorm/utils"
 )
 
+// Scope define short function name easy to use
+type Scope func(*DB) *DB
+
 // Model specify the model you would like to run db operations
 //    // update all users's name to `hello`
 //    db.Model(&User{}).Update("name", "hello")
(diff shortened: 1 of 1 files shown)


abelseniorly:
@lockp111 eventhough I like to define Scope function, it would confuse many with the older gorm v1 Scope... So I think this might not be necessary... @jinzhu whats your take?

lockp111:
> @lockp111 eventhough I like to define Scope function, it would confuse many with the older gorm v1 Scope... So I think this might not be necessary... @jinzhu whats your take?

Why not? It makes coder easier read the code.

jinzhu:
honestly, I don't think this is necessary, it doesn't make users easier to understand

PR #4096, opened 2021-02-17
change the method of initializing slice

sivchari:
- [x] Do only one thing
- [x] Tested

Hi, This is a small change, but it will send a PR.
There are places where the slice capacity is not specified.
I thought this would be better for Go's slice specification, since the capacity of the slice to be appended can be allocated in advance by the length.


sivchari:
sry, I'll resolve conflicts

PR #4113, opened 2021-02-22
Add ActivatedAt for scheduling future events.

ralic:
- [v] Do only one thing
- [v] Non breaking API changes
- [v] Tested

Add ActivatedAt for scheduling future events.

### User Case Description
Sometimes we hope to prepare partial or full information before it goes online. 
By default, gorm fills in a timestamp in the past. if you assigned a timestamp in the future, then it gets activated in the future .

### DIFF
--- a/model.go
+++ b/model.go
@@ -8,8 +8,10 @@ import "time"
 //      gorm.Model
 //    }
 type Model struct {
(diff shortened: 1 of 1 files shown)


jinzhu:
Better to create an own Model struct for this case.

Issue #4140, opened 2021-03-05
Add support for named arguments in Select


ratanphayade:
## GORM Playground Link
https://github.com/go-gorm/playground/pull/294

## Description

While using the `Where` condition we can use named arguments but the same is not true for `Select`.  This brings inconsistency in the usage of APIs. Currently `Select` only handle `?` as placeholders 
https://github.com/go-gorm/gorm/blob/a948c846071f7e4fd264c6a95a81a0ef04293a28/chainable_api.go#L95-L100

With named arguments, we can also allow the consumer to write queries like below
```go
           DB.Model(user).
		Select("sum( if( created_at >= @rangeOneStart AND created_at < @rangeOneEnd, 1, 0 )) as range_one, "+
			"sum( if( created_at >= @rangeTwoStart AND created_at < @rangeTwoEnd, 1, 0 )) as range_two",
		sql.Named("rangeOneStart", rangeOneStart),
		sql.Named("rangeOneEnd", rangeOneEnd),
		sql.Named("rangeTwoStart", rangeTwoStart),
		sql.Named("rangeTwoEnd", rangeTwoEnd)).
		Where("created_at >= @startTime", sql.Named("startTime", rangeTwoStart)).
		Where("created_at< @endTime", sql.Named("endTime", rangeTwoEnd)).
		Scan(&result).Error
```

Using named argument will help improve readability and help write complex queries

PR #4142, opened 2021-03-05
support named params in Select API 

ratanphayade:
- [x] Do only one thing
- [x] Tested

Adds support for named argument In Select API

### User Case Description

refer: issue https://github.com/go-gorm/gorm/issues/4140


daheige:
After reading your adjustments, it is true that for functional design, try to maintain the KISS single design principle. Give you a thumbs up, great job.

PR #4149, opened 2021-03-06
for Config.cacheStore store PreparedStmtDB key

daheige:
It is used to store the key stored in Config.cacheStore. It is recommended to use a constant. If you need to change, you can use var to modify.


daheige:

PR #4151, opened 2021-03-06
invalid db error and value and invalid value length error refactor.

daheige:
invalid db error
invalid value and invalid value length error


daheige:

PR #4185, opened 2021-03-14
Optimize code

daheige:
1.optimize MigrateColumn method for regexp
2.for FieldSchema.LookUpField or column Alias Delimiter

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -10,6 +10,11 @@ import (
 	"gorm.io/gorm/schema"
 )
 
+const (
+	// DoubleUnderscore for FieldSchema.LookUpField or column Alias Delimiter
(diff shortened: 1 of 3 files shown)


scan.go
```diff
@@ -10,6 +10,11 @@ import (
 	"gorm.io/gorm/schema"
 )
 
+const (
+	// DoubleUnderscore for FieldSchema.LookUpField or column Alias Delimiter
+	DoubleUnderscore = "__"
```
jinzhu, line 15: don't really want to export the `DoubleUnderscore` to public, which won't be confuse to end users.

daheige:
ok

PR #4191, opened 2021-03-16
Fix LIMIT clause not valid in DELETE or UPDATE SQL

hoojos:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested

Fix LIMIT clause not valid in DELETE And UPDATE  SQL.
```
    db.Limit(1).Delete(&Email{})
    // expected: DELETE FROM emails LIMIT 1;
    // actual: DELETE FROM emails;
    
    db.Model(User{}).Where("role = ?", "admin").Limit(1).Update({"age", 18})
    // expected: UPDATE users SET age=18 WHERE role = 'admin' LIMIT 1;
    // actual: UPDATE users SET age=18 WHERE role = 'admin';
```

### User Case Description
While I delete a lot of history records in a big MySQL's table, I want to use `DELETE xxx LIMIT 1000` rather than `DELETE xxx` to avoid `table lock`,  my code just like:
```
db.Where("id < ?" maxID).Limit(1000).Delete(&Model{})
```
but when I execute my script, whole Table was locked! according to check MySQL's transaction, I find actual SQL is `DELETE xxxx`, without `Limit 1000`. And then, I debugged gorm source code and found that gorm don't deal with `LIMIT` clause while building SQL in DELETE and UPDATE callbacks,

### DIFF
--- a/callbacks/delete.go
+++ b/callbacks/delete.go
@@ -135,7 +135,7 @@ func Delete(db *gorm.DB) {
 			}
 
 			db.Statement.AddClauseIfNotExists(clause.From{})
-			db.Statement.Build("DELETE", "FROM", "WHERE")
+			db.Statement.Build("DELETE", "FROM", "WHERE", "LIMIT")
 		}
 
 		if _, ok := db.Statement.Clauses["WHERE"]; !db.AllowGlobalUpdate && !ok && db.Error == nil {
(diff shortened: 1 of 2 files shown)


hoojos:
it seems to be not compatible with sqlite and sqlserver...

lifegit:
are you using PostgreSQL ? I also encountered this situation.

PR #4261, opened 2021-04-07
add method RollbackUnlessCommitted

labulaka521:
- [x] Do only one thing
- [x] Tested

if  transaction is not committed, this method `RollbackUnlessCommitted ` will rollback this transaction


jinzhu:
Not going to accept this PR, but suggest to use the `Transaction` method.

Thank you.

PR #4269, opened 2021-04-09
Feature/json

buger:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


jinzhu:
suggest to use it like https://github.com/go-gorm/datatypes/blob/master/json.go

buger:
Hi! Actually I accidentally raised this PR, I intended to raise it against my forked master :) 

My change have a bit different purpose, but I will try to clean it up and raise later.

Issue #4272, opened 2021-04-10
The design of ErrRecordNotFound in gorm.errors.go is unreasonable


daheige:
https://github.com/go-gorm/gorm/blob/d278ca49ef30f003c9624ae58d4d8726f728c1f7/errors.go#L11
ErrRecordNotFound = logger.ErrRecordNotFound It is not recommended to use this method to introduce in gorm.errors, but to put the errors together, and then gorm.ErrRecordNotFound can be used for logger/logger.go#145. It is wrong for you to deal with it this way. An errors.go also depends on the ErrRecordNotFound variable in logger/logger.go. If you want to migrate and maintain later, you have to introduce ErrRecordNotFound in the logger package, which does not conform to the package design principles. , Does not comply with the principle of least dependency, it is recommended to put it in gorm.errors.go.

daheige:
please keep the principle of least dependency

daheige:
https://github.com/daheige/gorm/commit/d4a1313643fc1f8a868730c8056f32a05ce0fb51

PR #4273, opened 2021-04-10
record not found back to the previous mode

daheige:
https://github.com/go-gorm/gorm/issues/4272

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -2,13 +2,11 @@ package gorm
 
 import (
 	"errors"
-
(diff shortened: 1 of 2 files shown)


jinzhu:
this https://github.com/go-gorm/gorm/blob/d278ca49ef30f003c9624ae58d4d8726f728c1f7/logger/logger.go#L51 won't works then.

jinzhu:
import cycle not allowed

PR #4298, opened 2021-04-17
There is no data to update when updating data, the update time should not change

yiranzai:
There is no data to update when updating data, the update time should not change

- [x] Do only one thing
- [x] Tested

When update data, if the map[string]interface{} has a length of 0 or the struct all value is zero. the update time should not change

### User Case Description

When update data, if the map[string]interface{} has a length of 0 or the struct all value is zero. the update time should not change


jinzhu:
closing this PR due to tests failed

Issue #4314, opened 2021-04-22
[Question] Fields Tags as a Placeholder


vzool:
How can I define Fields Tags as a placeholder so, later can be filled with fields in the query?

I think there is a way to do so, but I'm not aware of like this:
```golang
type City struct {
   Name string `json:"name,omitempty" gorm:"NOT NULL"`
   Ref          string `json:"ref,omitempty" gorm:"<-:false" sql:"-"`
}
```

At auto migration, a new field called ref is created as I think it should not because it's made for read-only

I guess I'm looking forward to a placeholder that can pass a value to struct while not creating it at auto migration.
I hope that I made my point clear.
Thnaks

vzool:
After I post this, I think if it does not exist. I have a suggestion for it like the following:
```golang
type City struct {
   Name string `json:"name,omitempty" gorm:"NOT NULL"`
   Ref          string `json:"ref,omitempty" gorm:"<->:false" sql:"-"`
}

```

vzool:
After some code review, I found the solution which is a field tag in read-mode and ignore migration.
```golang
type City struct {
   Name string `json:"name,omitempty" gorm:"NOT NULL"`
   Ref          string `json:"ref,omitempty" gorm:"<->" sql:"-"`
}
```

PR #4315, opened 2021-04-22
Add placeholder field tags feature

vzool:
- [x] Do only one thing
- [x] Tested

I have added a placeholder to the field tags which is for read-mode and ignore migration. It is very useful for auto field names generated by Raw query.

### User Case Description

#4314

### DIFF
--- a/schema/field.go
+++ b/schema/field.go
@@ -341,6 +341,15 @@ func (schema *Schema) ParseField(fieldStruct reflect.StructField) *Field {
 		}
 	}
 
+	// placeholder; readonly without migration
+
+	if _, ok := field.TagSettings["<->"]; ok {
+		field.Readable = true
+		field.Creatable = false
+		field.Updatable = false
+		field.IgnoreMigration = true
+	}
+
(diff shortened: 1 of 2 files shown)


jinzhu:
sorry, but not going to accept this feature. the change is not very obvious

vzool:
> sorry, but not going to accept this feature. the change is not very obvious

You had this feature [before](https://github.com/vzool/gorm/blob/69ebf1f4b99b368f5de5e7e2f94ef93e73820d85/schema/field.go#L314) which is not obvious, not documented, and not working!  😂

jinzhu:
Use it like `-:migration` or `-:all`

vzool:
First one `-:migration` not working at all, while this `-:all` just act like this `-`.
What I'm looking for is to read permission but ignore migration. And this is what this pull request is about.

jinzhu:
`-:migration` should actually works for `read permission but ignore migration`, if it doesn't work, could you create a playground PR?

vzool:
I already created a test for this [here](https://github.com/vzool/gorm/blob/69ebf1f4b99b368f5de5e7e2f94ef93e73820d85/schema/field_test.go#L238) and [here](https://github.com/vzool/gorm/blob/69ebf1f4b99b368f5de5e7e2f94ef93e73820d85/schema/field_test.go#L256) in the same PR. All tests passed.

Issue #44, opened 2014-01-03
Integration with Revel logger?


leebrooks0:
How would I best integrate the Revel logger (http://robfig.github.io/revel/manual/logging.html) with Gorm?

Would this work?

``` go
db.SetLogger(Revel.INFO)
```

PR #4330, opened 2021-04-27
Added missing white space

KarolosLykos:

Added some missing spaces.

### User Case Description

When using `clause.OnConflict`  with `DoUpdates` a white space was missing.

### DIFF
--- a/clause/on_conflict.go
+++ b/clause/on_conflict.go
@@ -33,14 +33,14 @@ func (onConflict OnConflict) Build(builder Builder) {
 	}
 
 	if onConflict.DoNothing {
-		builder.WriteString("DO NOTHING ")
+		builder.WriteString("DO NOTHING")
 	} else {
 		builder.WriteString("DO UPDATE SET ")
 		onConflict.DoUpdates.Build(builder)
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @KarolosLykos 

Thank you for your PR, I think maybe better to add a space before the `WHERE` in line #44

KarolosLykos:
I thought it would be better for consistency reasons. As all other checks adds a space at the end.

jinzhu:
> I thought it would be better for consistency reasons. As all other checks adds a space at the end.

Other places add a space at the end because there are always something after it, but not sure for this one.

KarolosLykos:
Ok I update so the space is added before the where clause.

clause/on_conflict.go
```diff
@@ -40,6 +40,7 @@ func (onConflict OnConflict) Build(builder Builder) {
 	}
 
 	if len(onConflict.Where.Exprs) > 0 {
+		builder.WriteByte(' ')
 		builder.WriteString("WHERE ")
```
jinzhu, line 44: change to? `builder.WriteString(" WHERE ")`

PR #4364, opened 2021-05-09
Optimizing the parse function in schema

chenquan:
- [x] Do only one thing
- [ ] Tested

Optimizing the `func Parse(dest interface{}, cacheStore *sync.Map, namer Namer) (*Schema, error)` function in schema.

### User Case Description
Store schema in advance to prevent other goroutines from generating schema repeatedly.

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -71,7 +71,7 @@ type Tabler interface {
 	TableName() string
 }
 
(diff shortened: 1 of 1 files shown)


chenquan:
@jinzhu Maybe it's worth your look?

Issue #4373, opened 2021-05-13
Updation question for custom time field


wk30:
Hey, guys. I am using the custom time field, found some problems with the model defination `autoUpdateTime`. Here is my code. 

The pkgs version I'm using.
golang version: 1.16.3
gorm.io/gorm 1.21.9
gorm.io/driver/mysql v1.0.6
gorm.io/driver/sqlite v1.1.4

[https://play.golang.org/p/jXqL446v8yF](https://play.golang.org/p/jXqL446v8yF)

What I really need is the custom time string not the timestamp.

![image](https://user-images.githubusercontent.com/18651103/118104126-50511e00-b40d-11eb-83cb-f27d66cfa68d.png)

The data `id = 1` is created first then updated and the data is not ok for the updated_at field.
The data `id = 2` is just created.

When using sqlite not found error, but the data in table is not correct.
When using mysql error occured, eg:
`Error 1292: Incorrect datetime value: '1620889489' for column 'updated_at' at row 1`

It's strange beacuse the creation is ok, while the updation is bad.

wk30:
I have another question, as my goal is to define a custom field, seems that it's not offically supported and I'm confused about the field (eg: `CuDateTime`) method with `Scan` and `Value`. The situation is that the creation always invoke `Scan` and `Value` method, while the updation invoke `Scan` and not always invoke `Value`. Is there anything I missed?

KiddoV:
I still have the same error.
The fields `autoCreateTime` always giving the correct timestamp format, but the fields with `autoUpdateTime` give me the timestamp as number instead.

PR #4379, opened 2021-05-14
Set default autoUpdateTime and autoCreatedTime value

wk30:
Even though I'm not sure about the original thought and this change is ok, #4375 could be fixed but still with the call problem about [`Scan` and `Value` methods](https://github.com/go-gorm/gorm/issues/4373#issuecomment-840967476)


jinzhu:
I don't get the change, can you add some tests to explain it?

wk30:
> I don't get the change, can you add some tests to explain it?

Sorry to link #4373, check please.

jinzhu:
Hi @wk30 

Please add some tests in the PR will fail and fix the test in this PR. Thank you.

Issue #4386, opened 2021-05-17
has_many stopped working if field names are identical


paraswaykole:
After updating to version v1.21.10, the models stopped working:

```
type Document struct {
	ID uint `gorm:"column:id;primarykey;"`
	OwnerID uint
	OwnerType string
	Users []*User `gorm:"foreignkey:OwnerID"`
}

type User struct {
	ID uint `gorm:"column:id;primarykey;"`
	Email string 
	Password *string `json:"-"`
	OwnerID *uint     
	Owner   *Document 
}
```

Error: invalid field found for struct models.Document's field Users, need to define a valid foreign key for relations or it need to implement the Valuer/Scanner interface

In v1.21.9 and earlier, this bug was not observed. I guess the last fix did something.

In the Document model, the OwnerID field has nothing to do with the OwnerID field from the User model. They just have the same name, which is causing the problem. Renaming the OwnerID field in the Document model resolves the issue.

_Originally posted by @yushulgin in https://github.com/go-gorm/gorm/issues/4343#issuecomment-842335955_

paraswaykole:
I have a fix ready for this. sending a PR soon.

paraswaykole:
@jinzhu Submitted a PR, please take a look :)

paraswaykole:
PR merged and issue fixed

PR #4343, opened 2021-05-01
fixed belongs_to & has_one reversed if field same

paraswaykole:
This PR is for the issue  #4316 (belongs_to and has_one reversed if field names are identical). 
It fixes the issue which was introduced in [`628a0ae`](https://github.com/go-gorm/gorm/commit/628a0ae707f230c67bca2e632fb302037c707705) commit.

- Added the correct conditions necessary to `reguessOrErr`.
- Added New Test case which handles the use-case that was failing.
- Added a utility function needed.

PR #4387, opened 2021-05-17
fixed has_many stopped working if field names are identical

paraswaykole:
This PR is for the issue #4386 (has_many stopped working if field names are identical).
It fixes the issue which was introduced in [`a222c75`](https://github.com/go-gorm/gorm/pull/4343/commits/a222c758bb1444d192d92c17be13721a8ede0725) commit.

- Corrected the condition.
- Added New Test case which handles the use-case that was failing.

### DIFF
--- a/schema/relationship.go
+++ b/schema/relationship.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/jinzhu/inflection"
 	"gorm.io/gorm/clause"
+	"gorm.io/gorm/utils"
 )
 
 // RelationshipType relationship type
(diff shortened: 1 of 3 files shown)


paraswaykole:
@jinzhu Please take a look at the PR

jinzhu:
test still pass w/o the change.

paraswaykole:
@jinzhu I added a new test `TestHasManySameForeignKey` which fails w/o the change. Just tested again. This is when there is has_many relationship with identical names as reported in #4386. I request you to please test again :)

PR #4402, opened 2021-05-21
FindInBatches using Offset instead of PrimaryKey iterations

JeffPHP:
- [x] Do only one thing
- [x] Tested

Actually find in Batches works using the primary key as an increment, so it assumes that the Batch is sorted with the PrimaryKey, which means that it will fail if the query contains any other Sort attribute.

I suggest to use the "offset" clause in order to iterate.

### User Case Description

The motivation is to get a more universal (and more simple function) working for every select, not only those sorted on PrimaryKey.

I probably miss out something, like Offset clause not being available on all SQL databases?

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -167,7 +167,7 @@ func (db *DB) Find(dest interface{}, conds ...interface{}) (tx *DB) {
 }
 
 // FindInBatches find records in batches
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @JeffPHP 

Using `Offset` might broken if you are making some changes to the data in batches.

PR #4495, opened 2021-07-03
slightly better callback warning

bdemirpolat:
- [x] Do only one thing
- [x] Tested

This PR will fix wrong callback warning.


bdemirpolat:
"Model %v don't match %vInterface, should be %v(*gorm.DB)" this warning may be misleading.
From GORM docs:
"The type of hook methods should be func(*gorm.DB) error"

PR #4538, opened 2021-07-16
Fix GORM returns ErrRecordNotFound when failed to find data with Find…

syanwei:
* [x] Do only one thing
* [x] Non breaking API changes
* [x] Tested

[ErrRecordNotFound description by official documentation](https://gorm.io/docs/error_handling.html#ErrRecordNotFound)

Fixed a bug GORM returns ErrRecordNotFound when failed to find data with *Find*, *Pluck*, *ScanRows*

The reality is not consistent with the official document description

### User Case Description

It is like this demo

```go
package main

import (
	"fmt"
	"gorm.io/driver/sqlite"
	"gorm.io/gorm"
	"log"
)

// DB ...
var DB *gorm.DB

func newDB(dsn string) *gorm.DB {
	db, err := gorm.Open(sqlite.Open(dsn), nil)
	if err != nil {
		log.Fatalln(err)
	}
	return db
}

func init() {
	var dns = fmt.Sprintf("file:%s?_synchronous=OFF", "test.db")

	DB = newDB(dns)
}

type car struct {
	ID    uint `gorm:"primarykey"`
	Color string
	Price int
}

func main() {
	err := DB.AutoMigrate(&car{})
	if err != nil {
		log.Fatalln(err)
	}
	type data struct {
		Cars       []*car `json:"cars" gorm:"-"`
		TotalPrice int    `json:"total_price"`
		Total      int    `json:"total"`
	}
	var d data
	tx := DB.Debug().Model(&car{})
	if err := tx.
		Select("count(1) as total, sum(price) as total_price").
		Take(&d).Error; err != nil {
		log.Fatalln(err)
	}
	// tx.Statement.RaiseErrorOnNotFound = false
	err = tx.Select("*").Where("cars.id>1").Find(&d.Cars).Error
	if err != nil {
        log.Fatalln("==========", err) // err: RecordNotFound
	}
}
```

**Stdout**

========== record not found

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -162,6 +162,7 @@ func (db *DB) Find(dest interface{}, conds ...interface{}) (tx *DB) {
 			tx.Statement.AddClause(clause.Where{Exprs: exprs})
 		}
 	}
+	tx.Statement.RaiseErrorOnNotFound = false
 	tx.Statement.Dest = dest
 	return tx.callbacks.Query().Execute(tx)
 }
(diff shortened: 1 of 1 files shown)


jinzhu:
https://gorm.io/docs/method_chaining.html

PR #4546, opened 2021-07-19
Support Raw func support use Limit method

xuejipeng:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested

Support Raw func support use Limit method
### User Case Description

When i do query like `t2 = tx.Raw("SELECT * FROM `pod_events`").Limit(1)`, It didn't prompt me any errors, I think it works fine，at that time when I deployed my service on the server it didn't work，this problem took me a lot of time to troubleshoot

I think supporting the Limit method on the Raw method will give users a better experience

### DIFF
--- a/callbacks/query.go
+++ b/callbacks/query.go
@@ -28,13 +28,28 @@ func Query(db *gorm.DB) {
 }
 
 func BuildQuerySQL(db *gorm.DB) {
+	var hasLimitClause bool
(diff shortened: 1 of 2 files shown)


daheige (approved):

jinzhu:
Hi @xuejipeng Raw should not work with Limit, or it might cause some issues.

PR #4618, opened 2021-08-16
Fix extra 'AND' when len(values) == 0 ON IN.NegationBuild()

secake:
- [x] Do only one thing
- [x] Tested

Three reasons for this PR:   
1. When `DB.Not(map[string]interface{}{"name": []string{}}).Where("age = ?", 10)`, Generate sql `SELECT * FROM ``users`` WHERE   AND ``age`` = 10;` , it is a bad sql.
2. In some database, like postgresql, when exec ` ``name`` NOT IN ('jinzhu')`, the result dose not contain the row where `name` == NULL, so when `is not` in an empty array,  to some extent, it can be considered that `name` IS NOT NULL. 
3. NegationBuild() should be the inverse of Build(), but when len(values)==0, the Build() return the row where `name`==NULL, so NegationBuild() should not return the row where `name`==NULL. 

### User Case Description

```go
DB.Not(map[string]interface{}{"name": []string{}}).Where("age = ?", 10).Find(&User{})`   
```

Thanks!

### DIFF
--- a/clause/expression.go
+++ b/clause/expression.go
@@ -210,11 +210,12 @@ func (in IN) Build(builder Builder) {
 }
 
 func (in IN) NegationBuild(builder Builder) {
+	builder.WriteQuoted(in.Column)
 	switch len(in.Values) {
 	case 0:
+		builder.WriteString(" IS NOT NULL")
 	case 1:
 		if _, ok := in.Values[0].([]interface{}); !ok {
-			builder.WriteQuoted(in.Column)
 			builder.WriteString(" <> ")
 			builder.AddVar(builder, in.Values[0])
 			break
@@ -222,7 +223,6 @@ func (in IN) NegationBuild(builder Builder) {
 
 		fallthrough
 	default:
-		builder.WriteQuoted(in.Column)
 		builder.WriteString(" NOT IN (")
 		builder.AddVar(builder, in.Values...)
 		builder.WriteByte(')')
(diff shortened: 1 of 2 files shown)


secake:
### Suggestion 
In other ORM, like django, when use `User.objects.filter(name__in=[])`, the result dose not contain the row where `name`==NULL, when use `User.objects.excludes(name__in=[])`, the result return all rows, also contains the row where `name`==NULL, as follows:
select id,name from users;
| id  |  name  |
| --  |  ----     |
| 1   |  jinzhu |
| 2   |  NULL  |
```py
User.objects.filter(name__in=[]).count()  // 0
User.objects.exclude(name__in=[]).count()` // 2
```

So I suggest the following implementation，if you think this suggestion works, I can implement it : 
```go
func (in IN) Build(builder Builder) {
	builder.WriteQuoted(in.Column)

	switch len(in.Values) {
	case 0:
		builder.WriteString(" 1=2")
	case 1:
		if _, ok := in.Values[0].([]interface{}); !ok {
			builder.WriteString(" = ")
			builder.AddVar(builder, in.Values[0])
			break
		}

		fallthrough
	default:
		builder.WriteString(" IN (")
		builder.AddVar(builder, in.Values...)
		builder.WriteByte(')')
	}
}

func (in IN) NegationBuild(builder Builder) {
	builder.WriteQuoted(in.Column)
	switch len(in.Values) {
	case 0:
		builder.WriteString(" 1=1")
	case 1:
		if _, ok := in.Values[0].([]interface{}); !ok {
			builder.WriteString(" <> ")
			builder.AddVar(builder, in.Values[0])
			break
		}

		fallthrough
	default:
		builder.WriteString(" NOT IN (")
		builder.AddVar(builder, in.Values...)
		builder.WriteByte(')')
	}
}
```

PR #4628, opened 2021-08-22
Fix onConflict with non-updatable in associations

zkqiang:
- [x] Do only one thing
- [x] Tested

If a column is non-updatable `gorm:"->"`, it will still conflict updating in associations.

So we need to add a judgment.

### User Case Description

```go
type User struct {
	gorm.Model
	Name      string
	CompanyID *int
	Company   Company
}

type Company struct {
	ID   int
	Name string
	ReadOnlyField string `gorm:"->"`
}

func TestUpdateBelongsTo(t *testing.T) {
	user := User{Name: "user-belongs-to-association"}
	user.Company = Company{Name: "company-belongs-to-association"}
	if err := DB.Create(&user).Error; err != nil {
		t.Fatalf("errors happened when create: %v", err)
	}
	user.Company.Name += "new"
	if err := DB.Debug().Session(&gorm.Session{FullSaveAssociations: true}).Save(&user).Error; err != nil {
		t.Fatalf("errors happened when update: %v", err)
	}
}

// Before fix:
// INSERT INTO `companies` (`name`,`id`) VALUES ("company-belongs-to-associationnew",1) ON CONFLICT (`id`) DO UPDATE SET `name`=`excluded`.`name`,`read_only`=`excluded`.`read_only`

// After fix:
// INSERT INTO `companies` (`name`,`id`) VALUES ("company-belongs-to-associationnew",1) ON CONFLICT (`id`) DO UPDATE SET `name`=`excluded`.`name`
```

### DIFF
--- a/callbacks/associations.go
+++ b/callbacks/associations.go
@@ -314,7 +314,7 @@ func onConflictOption(stmt *gorm.Statement, s *schema.Schema, selectColumns map[
 	if stmt.DB.FullSaveAssociations {
 		defaultUpdatingColumns = make([]string, 0, len(s.DBNames))
 		for _, dbName := range s.DBNames {
-			if v, ok := selectColumns[dbName]; (ok && !v) || (!ok && restricted) {
+			if v, ok := selectColumns[dbName]; (ok && !v) || (!ok && restricted) || !s.FieldsByDBName[dbName].Updatable {
 				continue
 			}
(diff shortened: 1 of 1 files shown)


jinzhu:
Are you using the latest version, I remember this should be fixed? or can you write some test cases?

PR #4671, opened 2021-09-03
support unique index and soft delete

Laisky:
- [x] Do only one thing
- [x] Tested

Add support to unique index & soft delete.

#### Current

When use unique index with soft-delete. deleted row still occupied unique index. 

For example:

| id | deleted_at | unique_col |
| :---: | :---: | :---: |
| 1 | 2021-09-03T00:00:00Z | 1 |

If you want insert an row with `unique_cil=1`, will be rejected. even if there is no undeleted row with value 1.

### With this patch

with this patch, when you add an new column named `deleted_flag`, 
deleted row will not occupied unique index.

* migrating: every unique index will be convert to composite index with `deleted_flag`
* creating: `deleted_flag` is default to 0
* deleting: `deleted_flat` will set to primary id

For example:

| id | deleted_at | deleted_flag | unique_col |
| :---: | :---: | :--: |:---: |
| 1 | 2021-09-03T00:00:00Z | 1 | 1 |

When you insert an new row with `unique_col = 1`, the insert row will be `(2, nil, 0, 1)`, it won't conflict with the deleted row.

### DIFF
--- a/clause/on_conflict.go
+++ b/clause/on_conflict.go
@@ -26,7 +26,7 @@ func (onConflict OnConflict) Build(builder Builder) {
 		}
 		builder.WriteString(`) `)
 	}
-	
+
 	if len(onConflict.TargetWhere.Exprs) > 0 {
 		builder.WriteString(" WHERE ")
 		onConflict.TargetWhere.Build(builder)
--- a/model.go
+++ b/model.go
@@ -14,9 +14,27 @@ type Model struct {
 	DeletedAt DeletedAt `gorm:"index"`
 }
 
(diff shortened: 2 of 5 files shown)


jinzhu:
Hello @Laisky 

Checkout https://github.com/go-gorm/soft_delete , not going to make any changes to the build-in soft delete feature.

Thank you.

Laisky:
> https://github.com/go-gorm/soft_delete

It seems two different cases.

`https://github.com/go-gorm/soft_delete` still do not support soft-delete with unique index.

PR #4674, opened 2021-09-05
feat: copy dest fields to model struct

riverchu:
- [x] Do only one thing
- [x] Tested

Fix the dislocation of struct fields when calling `Updates`


daheige (changes requested):
Still needs to be optimized

PR #4680, opened 2021-09-07
feat: adjust SetupJoinTable func if..else code

daheige:
Adjust the if...else logic of this function to make the context logically coherent and express the meaning simply and intuitively
@jinzhu


daheige:
@jinzhu I have adjusted it, please CR/MR again

PR #4762, opened 2021-10-11
[WIP] fix: automigrate error caused by indexes while using dynamic table name

### LINKED ISSUE #4752: Automigrate error caused by indexes while using dynamic table name
## GORM Playground Link

https://github.com/go-gorm/playground/pull/386

## Description

I am trying to migrate tables using dynamic table name. All these tables have different name but the same schema. I am using scopes to generate the dynamic table name.

The tables are created with the dynamic table name but the indexes are not created as it should.

Struct
type NewUser struct {
	gorm.Model
	Name  string `gorm:"index"`
	City  string `gorm:"index"`
	Table string `gorm:"-"`
}

tableNameFormat `newuser_%x`

tableName := []string{"a","b","c"}

table created successfully `newuser_a`, `newuser_b`,`newuser_c`

When using scopes to generate the dynamic table name, the indexes are formatted as `idx_new_users_name` for all the tables and thus error -> `Failed, got error: index idx_new_users_name already exists`

### Expected Output:

While doing migration, the tables with dynamic names should be created along with its indexes.

xwjdsh:
- [x] Do only one thing
- [x] Tested

fix #4752

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -687,10 +687,10 @@ func (m Migrator) ReorderModels(values []interface{}, autoAdd bool) (results []i
 		parsedSchemas                 = map[*schema.Schema]bool{}
 		valuesMap                     = map[string]Dependency{}
 		insertIntoOrderedList         func(name string)
-		parseDependence               func(value interface{}, addToList bool)
+		parseDependence               func(value interface{}, addToList, useSchemaTable bool)
 	)
 
-	parseDependence = func(value interface{}, addToList bool) {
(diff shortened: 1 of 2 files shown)


longlihale:
@xwjdsh   hello, you can look following code.
https://github.com/go-gorm/gorm/blob/418c60c83cf8472d883bb9ab8b9821444e7c8f0a/schema/index.go#L108

index's table name use schema.fields[i].field.Table instead  use statement.Table, so result https://github.com/go-gorm/gorm/issues/4752.  
in fact,schema.fields[i].field.Schema.Table  use reflect type name or Tabler interface  to set Table.

Issue #4372, opened 2021-05-12
Bug: self-referential has-one does not create foreign key


ezk84:
https://github.com/go-gorm/playground/pull/322

In the existing playground models the user foreign key (manager_id) gets created by the has-many Team relation.

When that is no longer present no foreign key is created for the self-referential has-one relation corresponding to the Manager member.

This is a test case which shows this bug with a separate model:

```golang
type Task struct {
	gorm.Model
	Content  string    `gorm:"not null; default:Empty Task" json:"content"`

	ParentTaskID *uint `gorm:"not null" json:"parent_task_id"`
	ParentTask   *Task `gorm:"foreignKey:ParentTaskID" json:"-"`
}
```
When this is AutoMigrated, parent_task_id isn't marked as a foreign key.

ChappIO:
I ran into the same issue with PostgreSQL and reported it before I found this ticket. I have closed my report and will keep track of this one.

hovinlu:
I also met this bug.

jinzhu:
Should be fixed, thank you for your report.

QUDUSKUNLE:
This has not been fixed..

alexanderhorner:
This hasn't been fixed, please reopen the issue.

PR #4801, opened 2021-10-24
Fix self-referential belongs to constraint

mgovilla:

- Added a condition in the parse constraint in order to differentiate between a duplicate constraint (caused by a belongs-to and a has-many relationship between two models) and a self-referential belongs-to.
- There are probably many ways to make this check (perhaps checking the type of the relation instead of the whole object), so I would appreciate any feedback!

### User Case Description

Addresses the concerns from #4372 (now able to create self-referential belongs-to constraints)

### DIFF
--- a/tests/migrate_test.go
+++ b/tests/migrate_test.go
@@ -54,6 +54,27 @@ func TestMigrate(t *testing.T) {
 	}
 }
 
+func TestAutoMigrateSelfReferential(t *testing.T) {
+	type MigratePerson struct {
+		ID        uint
+		Name      string
+		ManagerID *uint
+		Manager   *MigratePerson
+	}
+
+	if err := DB.Debug().Migrator().DropTable(&MigratePerson{}); err != nil {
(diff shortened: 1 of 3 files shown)


jinzhu:
seems like tests still pass w/o the change?

tests/migrate_test.go
```diff
@@ -54,6 +54,27 @@ func TestMigrate(t *testing.T) {
...
 
+func TestAutoMigrateSelfReferential(t *testing.T) {
+	type MigratePerson struct {
+		ID        uint
+		Name      string
+		ManagerID *uint
+		Manager   *MigratePerson
+	}
+
+	if err := DB.Debug().Migrator().DropTable(&MigratePerson{}); err != nil {
```
jinzhu, line 65: maybe the drop table error check is not necessary? it will broken when running the test first time? also please remove the `Debug()` method here.

mgovilla: make sense, I have removed that.

PR #4831, opened 2021-11-03
add connections leak test

longlihale:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


jinzhu:
in general I think the tests is not solid enough.... and might broken when running concurrent tests

PR #4833, opened 2021-11-03
preload null judgment

0fv:
when using count contain preload，will get panic message

### DIFF
--- a/callbacks/query.go
+++ b/callbacks/query.go
@@ -186,6 +186,9 @@ func BuildQuerySQL(db *gorm.DB) {
 }
 
 func Preload(db *gorm.DB) {
+	if db.Statement.Schema == nil {
+		return
+	}
 	if db.Error == nil && len(db.Statement.Preloads) > 0 {
 		preloadMap := map[string]map[string][]interface{}{}
 		for name := range db.Statement.Preloads {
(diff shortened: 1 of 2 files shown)


huacnlee:
Need a test for cover this.

0fv:
Added

jinzhu:
Hi @0fv 

In this case, I think we should return an error, the `Preload` doesn't works as expected.

0fv:
> Hi @0fv
> 
> In this case, I think we should return an error, the `Preload` doesn't works as expected.

In practice,something like page load,sometime need load total count first, when count equal 0,no need call Find function,and after count need add limit and offset get correct page.Here is  my example

```go
type Page struct {
	Page  int
	Limit int
	Total int64
}

func (p *Page) LoadPage(db *gorm.DB, data interface{}) error {
	err := db.Count(&p.Total).Error
	if err != nil {
		return err
	}
	if p.Total == 0 {
		return nil
	}
	offset := (p.Page - 1) * p.Limit
	return db.Offset(offset).Limit(p.Limit).Find(data).Error
}
```

daheige (approved):

jinzhu:
maybe consider another way to wrap it, ignore the error silently doesn't sound correct.

daheige:
@jinzhu,support what you said.



发自我的iPhone


From: Jinzhu ***@***.***&gt;
Date: Mon,Nov 29,2021 9:27 AM
Cc: heige ***@***.***&gt;, Comment ***@***.***&gt;
Subject: Re: [go-gorm/gorm] preload null judgment (PR #4833)





 
maybe consider another way to wrap it, ignore the error silently doesn't sound correct.
 
—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

PR #4841, opened 2021-11-06
feat(migrator,migrator/migrator.go,tests/migrate_test.go) : Get multiple data tables for migrator.

dino-ma:


migrator.go
```diff
@@ -33,6 +33,14 @@ type ViewOption struct {
 	Query       *DB
 }
 
+// ALLTables get all database tables
```
jinzhu, line 36: maybe just add a new method `GetTables() []string` should be enough?

dino-ma: Yet

PR #4852, opened 2021-11-13
fix(updateClauses): add FROM

kelvinfloresta:
- [x] Do only one thing
- [ ] Tested

Add support to Postgres to update when you need to join with another table to perform a where (or update a column with join table data)

Postgres needs a clause FROM to JOIN with another table during an UPDATE

#### Postgres docs:
https://www.postgresql.org/docs/current/sql-update.html
`A table expression allowing columns from other tables to appear in the WHERE condition and update expressions. This uses the same syntax as the FROM clause of a SELECT statement; for example, an alias for the table name can be specified. Do not repeat the target table as a from_item unless you intend a self-join (in which case it must appear with an alias in the from_item).`

#### Adding FROM manually
```go
db = db.Clauses(clause.From{
    Tables: []clause.Table{{Name: "my_join_table"}},
).Where("my_join_table.id = ?", input.value))
db.Statement.BuildClauses = append(db.Statement.BuildClauses, "UPDATE", "SET", "FROM", "WHERE")

db.Updates(myNewValues)
```

#### After this pull request
```go
db.Clauses(clause.From{
    Tables: []clause.Table{{Name: "my_join_table"}},
).Where("my_join_table.id = ?", input.value)).Updates(myNewValues)
```

### DIFF
--- a/callbacks/callbacks.go
+++ b/callbacks/callbacks.go
@@ -7,7 +7,7 @@ import (
 var (
 	createClauses = []string{"INSERT", "VALUES", "ON CONFLICT"}
 	queryClauses  = []string{"SELECT", "FROM", "WHERE", "GROUP BY", "ORDER BY", "LIMIT", "FOR"}
-	updateClauses = []string{"UPDATE", "SET", "WHERE"}
+	updateClauses = []string{"UPDATE", "SET", "FROM", "WHERE"}
 	deleteClauses = []string{"DELETE", "FROM", "WHERE"}
 )
 


kelvinfloresta:
Example:
```
 db.Clauses(clause.From{
    Tables: []clause.Table{{Name: "my_join_table"}},
).Where("my_join_table = ?", input.value)
```

daheige (approved):

jinzhu:
Hello @kelvinfloresta 

Maybe should change the postgres driver to add it?

PR #4870, opened 2021-11-23
feat: ColumnType interface

qqxhb:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


daheige:
@qqxhb For this, you should give the interface implementation part, 
and also need to do compatibility processing for the old version.

PR #4871, opened 2021-11-23
Add unused argument

jinzhu:
- [*] Do only one thing
- [*] Tested

### DIFF
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -10,4 +10,4 @@ jobs:
       - name: golangci-lint
         uses: reviewdog/action-golangci-lint@v2
         with:
-          golangci_lint_flags: '--enable-all --exclude-use-default=false'
+          golangci_lint_flags: '-E cyclop,unconvert,misspell,unparam,ineffassign,maligned,gocritic,prealloc,scopelint,scopelint,gosec'
--- a/.github/workflows/reviewdog.yml
+++ b/.github/workflows/reviewdog.yml
@@ -6,6 +6,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out code into the Go module directory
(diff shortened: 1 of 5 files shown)


spantaleev:
This appears to be a breaking change.

Code that used to work before:

```golang
entities := make([]*SomeModel, 0)

sql := "SELECT * FROM some_table WHERE 1"
bindParams := make(map[string]interface{})

if someCondition() {
	sql += " AND key = @key"
	bindParams["key"] = "something"
}

// Previously, this worked as is, regardless of what bindParams contained.
//
// Currently:
// If bindParams contains at least 1 value (or rather, if sql contains `@`), this will be OK.
// Likely because it uses NamedExpr, not Expr.
//
// If bindParams is an empty map[string]interface{}, we get this error:
// > ERRO[0005] sql: converting argument $1 type: unsupported type map[string]interface {}, a map
err = db.Raw(sql, bindParams).Find(&entities).Error
```

----

Workaround 1:

```golang
entities := make([]*SomeModel, 0)

sql := "SELECT * FROM some_table WHERE 1"
bindParams := make(map[string]interface{})

if someCondition() {
	sql += " AND key = @key"
	bindParams["key"] = "something"
}

if len(bindParams) > 0 {
	err = db.Raw(sql, bindParams).Find(&entities).Error
} else {
	err = db.Raw(sql).Find(&entities).Error
}
```

----

Workaround 2:

```golang
entities := make([]*SomeModel, 0)

sql := "SELECT * FROM some_table WHERE 1"
bindParams := make(map[string]interface{})

if someCondition() {
	sql += " AND key = @key"
	bindParams["key"] = "something"
}

args := make([]interface{}, 0)
if len(bindParams) > 0 {
	args = append(args, bindParams)
}

err = db.Raw(sql, args...).Find(&entities).Error
```

Issue #4877, opened 2021-11-27
let gorm can scan protobuf struct without add an ignore tag ("-") manually


frankee:
## Describe the feature

let gorm can scan protobuf struct without add an ignore tag ("-") manually.

## Motivation

In gRPC backend service, we usually use the protobuf to describe all the data model, but the data model which is generated by the protobuf, can't work properly in the gorm without change anything.

If gorm could work properly in that situation, that will help developers to reduce some effort.

## Related Issues

https://github.com/go-gorm/gorm/issues/3994
https://github.com/go-gorm/gorm/issues/3990

PR #4878, opened 2021-11-27
let gorm can scan protobuf struct without add an ignore tag ("-")

frankee:
let gorm can scan protobuf struct without add an ignore tag ("-")

link: https://github.com/go-gorm/gorm/issues/4877

- [V] Do only one thing
- [V] Non breaking API changes
- [V] Tested

implement the  [#4877](https://github.com/go-gorm/gorm/issues/4877)

### User Case Description

when gorm process the struct which is generated by the protobuf.


jinzhu:
Rely on the `json:"-"` tag doesn't sound correct.

PR #4887, opened 2021-11-30
feat(order): support call Order once to order by multi column

bestgopher:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested

```
var a []interface
db.Table("aaa").
		Order(clause.OrderByColumns{
			{Column: clause.Column{Name: "e"}, Desc: true},
			{Column: clause.Column{Name: "f"}, Desc: false},
		}).
		Find(&a)
```
sql: ```SELECT * FROM "aaa" ORDER BY "e" DESC,"f"```

### User Case Description
When I want to order by multi columns,I can call Order like this:
```
db.Table("aaa").
		Order(clause.OrderByColumn{Column: clause.Column{Name: "id"}, Desc: true}).
		Order(clause.OrderByColumn{Column: clause.Column{Name: "other column"}, Desc: false}).
		Find(&a)

// SELECT * FROM "aaa" ORDER BY "id" DESC,"other column"
```
This pr only implement a more convenient way which pass a `[]clause.OrderByColumn` to `Order` instead of calling `Order` multi time

```
db.Table("aaa").
		Order(
			clause.OrderByColumns{
				{Column: clause.Column{Name: "id"}, Desc: true},
				{Column: clause.Column{Name: "other column"}, Desc: false},
			},
		).
		Find(&a1)
```

### DIFF
--- a/chainable_api.go
+++ b/chainable_api.go
@@ -219,6 +219,12 @@ func (db *DB) Order(value interface{}) (tx *DB) {
 		tx.Statement.AddClause(clause.OrderBy{
 			Columns: []clause.OrderByColumn{v},
 		})
+
+	case clause.OrderByColumns:
+		tx.Statement.AddClause(clause.OrderBy{
+			Columns: v,
+		})
+
 	case string:
 		if v != "" {
 			tx.Statement.AddClause(clause.OrderBy{
--- a/clause/order_by.go
+++ b/clause/order_by.go
@@ -6,6 +6,8 @@ type OrderByColumn struct {
 	Reorder bool
 }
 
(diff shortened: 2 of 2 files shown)


jinzhu:
Would like to reject it as don't want to export another struct for a rarely used sugar API.

Sorry for that, thank you for your PR.

bestgopher:
Could add a new case banch like this:
```
case []clause.OrderByColumn:
		tx.Statement.AddClause(clause.OrderBy{
			Columns: v,
		})
```
I only want to add multi fields ordering stmt by call `Order` once.

PR #4888, opened 2021-12-01
fix type alias AutoMigrate bug（Add Test Case）

piyongcai:
- [x] Do only one thing
- [x] Tested

At some times, we need to define alias types for native data types, such as easy reuse for some common interfaces.
but, when you denfine alias types, the AutoMigrate will crash!

### User Case Description

eg: 
1. defing `IDer` interface.
2. define `int64` alias type as `ID`.
3. define a `Test` struct contian column `ID`.
4. call `AutoMigrate`, and gorm return error `invalid embedded struct for Test's field ID, should be struct, but got main.ID`

bellow is the full demo code.

```go
package main

import (
	"log"

	"gorm.io/driver/postgres"
	"gorm.io/gorm"
	"gorm.io/gorm/logger"
)

type IDer interface{ GetID() int64 }

// ID will add some method to implement some interface eg: GetID
type ID int64

func (z ID) GetID() int64 { return int64(z) }

type Test struct {
	ID
	Code string `gorm:"size:50"`
	Name string `gorm:"size:50"`
}

func main() {
	db, err := gorm.Open(postgres.New(postgres.Config{
		DSN:                  `dsn`,
		PreferSimpleProtocol: false,
	}), &gorm.Config{
		Logger:                 logger.Default.LogMode(logger.Info),
		SkipDefaultTransaction: true,
	})
	if err != nil {
		log.Fatal(err)
	}

	if err = db.AutoMigrate(&Test{}); err != nil {
		// invalid embedded struct for Test's field ID, should be struct, but got main.ID
		log.Fatal(err)
	}
}
```

### DIFF
--- a/callbacks/associations.go
+++ b/callbacks/associations.go
@@ -207,7 +207,7 @@ func SaveAfterAssociations(create bool) func(db *gorm.DB) {
 							}
 
 							cacheKey := utils.ToStringKey(relPrimaryValues)
-							if len(relPrimaryValues) == 0 || (len(relPrimaryValues) == len(rel.FieldSchema.PrimaryFields) && !identityMap[cacheKey]) {
+							if len(relPrimaryValues) != len(rel.FieldSchema.PrimaryFields) || !identityMap[cacheKey] {
 								identityMap[cacheKey] = true
 								if isPtr {
 									elems = reflect.Append(elems, elem)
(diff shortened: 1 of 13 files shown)


jinzhu:
Can you please create a test for this?

piyongcai:
> Can you please create a test for this?

Test case Added.
`gorm/schema/field_test.go`

piyongcai:
> Can you please create a test for this?

but I don't understand the error of `reviewdog / runner / golangci-lint (pull_request) `. 
and I don't know how to resolve it.

PR #4908, opened 2021-12-08
Remove redundant percision check for migrator

wushu037:
extra checking can cause "tag:percision" to lose meaning

- [√] Do only one thing
- [√] Non breaking API changes
- [√] Tested

remove redundant percision check for migrator

### User Case Description
if I want to change the percision

before
image

now
image


jinzhu:
in some case, it will keep altering columns. for your case, maybe just don't set column's type.

PR #4921, opened 2021-12-12
Ability to use pure-Go SQLITE driver

glebarez:
This PR is solely for triggering CI checks against non-standard, pure-Go SQLite driver for GORM.
The corresponding issue is: https://github.com/go-gorm/gorm/issues/4922

### DIFF
--- a/tests/go.mod
+++ b/tests/go.mod
@@ -10,7 +10,7 @@ require (
 	golang.org/x/crypto v0.0.0-20211209193657-4570a0811e8b // indirect
 	gorm.io/driver/mysql v1.2.1
 	gorm.io/driver/postgres v1.2.3
-	gorm.io/driver/sqlite v1.2.6
+	github.com/glebarez/sqlite v1.3.2
 	gorm.io/driver/sqlserver v1.2.1
 	gorm.io/gorm v1.22.4
 )
--- a/tests/tests_test.go
+++ b/tests/tests_test.go
@@ -9,7 +9,7 @@ import (
 
 	"gorm.io/driver/mysql"
 	"gorm.io/driver/postgres"
-	"gorm.io/driver/sqlite"
+	"github.com/glebarez/sqlite"
 	"gorm.io/driver/sqlserver"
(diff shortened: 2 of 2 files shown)


jinzhu:
Nice job, @glebarez have you ever compared its performance?

If the pure-go one's performance is better or similar, then I would like to replace the old one, or create another driver if it is much worse.

piyongcai:
> Nice job, @glebarez have you ever compared its performance?
> 
> If the pure-go one's performance is better or similar, then I would like to replace the old one, or create another driver if it is much worse.

[look....](https://gitlab.com/cznic/sqlite/-/issues/27)

glebarez:
Currently there are:
- simple insert / select benchmarks vs Mattn's implementation (https://gitlab.com/cznic/sqlite/-/tree/master/benchmark). This shows that pure-Go may be faster that Cgo in some cases
- https://gitlab.com/cznic/sqlite also has TPCH benchmark implemented for both pure-Go and CGo drivers. You may just run tpch.sh to see the results.

glebarez:
I've prepared a set of comprehensive benchmarks for this driver against mattn's:
https://github.com/glebarez/go-sqlite/tree/master/benchmark

Banyc (approved):
It works for me.

morremeyer:
@glebarez @jinzhu Are there plans to integrate @glebarez's work back into upstream or is this unlikely to happen because of the performance penalty?

glebarez:
> @glebarez @jinzhu Are there plans to integrate @glebarez's work back into upstream or is this unlikely to happen because of the performance penalty?

Hello. I don't think that's reasonable because:
- underlying database/sql drivers are different in implementation, and even connection string syntax varies
- GORM simply too mature to pull out such tricks, since it may break a lot of dependent code
- I'm happy with just pure-Go driver listed in docs, so people are free to choose which one suites best

I hope this claryfies things 🙂

morremeyer:
Yes, it does. Thank you for your work on this.

PR #4937, opened 2021-12-18
Fix: Where clauses with named arguments may cause generation of unintended queries

emregullu:
* [x]  Do only one thing
* [x]  Non breaking API changes
* [x]  Tested

During the building of Where clauses, the NamedExpr type is omitted while determining whether the expression should be wrapped in parentheses. Therefore, although it is required, Where clauses containing both named arguments and OR conditions are not wrapped, and this causes the generation of different queries than desired. This pull request tries to fix this issue.
### User Case Description

eg:

For the gorm statement below,

```go
db.Model(&User{}).Where("role = @role1 OR role = @role2",
map[string]interface{}{"role1": "admin", "role2": "super_admin"}).Find(&results)
```

the following query is generated.

~~~~sql
SELECT * FROM users WHERE role = 'admin' or role = 'super_admin' AND users.deleted_at IS NULL
~~~~

Due to the precedence of `AND` over `OR`, the query will be interpreted like
~~~~sql
`WHERE role = 'admin' or (role = 'super_admin' AND users.deleted_at IS NULL)` 
~~~~

instead of

~~~~sql
`WHERE (role = 'admin' or role = 'super_admin') AND users.deleted_at IS NULL` 
~~~~

which is actually intended, and it leads to getting softly deleted admin users without being aware, which may also be dangerous.

### DIFF
--- a/clause/where.go
+++ b/clause/where.go
@@ -60,6 +60,9 @@ func buildExprs(exprs []Expression, builder Builder, joinCond string) {
 			case Expr:
 				sql := strings.ToLower(v.SQL)
 				wrapInParentheses = strings.Contains(sql, "and") || strings.Contains(sql, "or")
+			case NamedExpr:
+				sql := strings.ToLower(v.SQL)
+				wrapInParentheses = strings.Contains(sql, "and") || strings.Contains(sql, "or")
 			}
 		}
 
--- a/tests/named_argument_test.go
+++ b/tests/named_argument_test.go
@@ -2,6 +2,7 @@ package tests_test
 
 import (
 	"database/sql"
+	"errors"
 	"testing"
 
 	"gorm.io/gorm"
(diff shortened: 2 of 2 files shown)


clause/where.go
```diff
@@ -60,6 +60,9 @@ func buildExprs(exprs []Expression, builder Builder, joinCond string) {
 			case Expr:
```
sryavuz, line 60: We can use case list instead of a separate case:
``` go
case Expr, NamedExpr:
```

emregullu: @sryavuz thanks for the review.

I tried this way first, but it didn't work. Because, for the cases with multiple types, the type of the variable won't be narrowed. So the type of `v` will be Expression interface, which invalidates accessing `SQL` field as `v.SQL`. To handle this, we can use additional type assertion or a `reflection` based approach like `reflect.ValueOf(v).FieldByName("SQL").String()`. However, I think it's better to leave it as is, despite the code duplication.

sryavuz (approved):
Thank you for the catch, LGTM 👍

PR #4959, opened 2021-12-27
require lint before tests

mmorel-35:
Use golangci/golangci-lint-action instead of reviewdog/action-golangci-lint as the second was not reporting any failures even if there was some.

Report code coverage with codecov/codecov-action
I have set some flags per dialect and go version

Several linters has been fixed, some disabled so the build can pass


daheige (commented):
I probably took a look at it. There are so many changes. Can it be divided into several features to submit?

jinzhu:
closing due to merge conflicts

PR #4964, opened 2021-12-28
improve the error handle in tests_test

liweitingwt:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


longlihale (commented):
thanks

daheige (approved):

PR #4980, opened 2022-01-05
fix: auto migration column order unpredictable

### LINKED ISSUE #4351: Deterministic execution of AutoMigrate
## Describe the feature

At the moment, the order of which columns are migrated during `AutoMigrate` is non-deterministic. 

I'm using [copyist](https://github.com/cockroachdb/copyist) to test our migrations, which when run in "recording" mode, intercepts all SQL operations and saves them to a file. Then you can run the tests in "playback" mode and if the queries run do not match those which were recorded, then the test fails. This is great when you want to run a full test suite and you don't want to have a database running locally.

I'm creating a feature request, rather than a bug, since I don't think this would have been an intended feature unless someone else has specifically requested it.

## Motivation

Tests are good, tests without having to stand up a full database instance are even better.

Essentially, this for loop on line 99 iterates in a non-deterministic way, meaning sometimes my migrations migrate columns in one order, then in another run in another order: https://github.com/go-gorm/gorm/blob/2aca96d1474967da11bac81a58db9c97bd7bdcac/migrator/migrator.go#L96-L102

My request is to build a slice of column names, sort them, and then iterate over that slice instead.

## Related Issues

Couldn't find any related issues.

halfcrazy:
- [X] Do only one thing
- [X] Tested

Fix #4351 make migration column order predictable.

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -97,11 +97,12 @@ func (m Migrator) AutoMigrate(values ...interface{}) error {
 			if err := m.RunWithValue(value, func(stmt *gorm.Statement) (errr error) {
 				columnTypes, _ := m.DB.Migrator().ColumnTypes(value)
 
-				for _, field := range stmt.Schema.FieldsByDBName {
+				for _, dbName := range stmt.Schema.DBNames {
+					field := stmt.Schema.FieldsByDBName[dbName]
 					var foundColumn gorm.ColumnType
 
 					for _, columnType := range columnTypes {
-						if columnType.Name() == field.DBName {
+						if columnType.Name() == dbName {
 							foundColumn = columnType
 							break
 						}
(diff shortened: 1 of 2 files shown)


tests/go.mod
```diff
@@ -16,3 +16,5 @@ require (
 )
 
 replace gorm.io/gorm => ../
+
+replace gorm.io/driver/mysql => github.com/halfcrazy/mysql v1.2.3-0.20220106042630-805c93c0d8d9
```
jinzhu, line 20: Can you remove those debug code, thank you.

halfcrazy: done

halfcrazy: ci get failed after switching origin MySQL driver. pls take a look at https://github.com/go-gorm/mysql/pull/60

jinzhu:
Hi @halfcrazy 

Can you format the source code with `gofmt` according to the bytedog suggestion? thank you.

halfcrazy:
> Hi @halfcrazy
> 
> Can you format the source code with `gofmt` according to the bytedog suggestion? thank you.

Done

PR #5030, opened 2022-01-27
add before find hook,we need split table with it

dino-ma:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


huacnlee:
You can use ConnPool replace for catching the SQL query before it execute to database.

See the [Sharding](https://github.com/go-gorm/sharding) plugin implementation.

https://github.com/go-gorm/sharding/blob/main/conn_pool.go#L58

jinzhu:
Closing this PR, the `BeforeFind` hooks must be used very carefully, as it usually relied on the query results in the database.

Suggest to use `Scopes` or `Callbacks` for the case.

Thank you for your PR.

PR #5074, opened 2022-02-14
Fixed the use of "or" to be " OR ", to account for words that contain…

sammyrnycreal:
Fixed the use of "or" to be " OR ", via a constant, to account for words that contain "or" or "and" (e.g., 'score', 'band') in a sql statement as the name of a DB field. While it doesn't create an error, it's not consistent with how the other " OR " and " AND " are defined and used.

Example Currently:
```
SELECT * FROM `users` WHERE (`users`.`id` <> ? AND `age` <= ?) OR `score` <= ? got SELECT * FROM `users` WHERE (`users`.`id` <> ? AND `age` <= ?) OR (`score` <= ?)
```

is generated but, the last field should not be enclosed in parentheses it should be:
```
SELECT * FROM `users` WHERE (`users`.`id` <> ? AND `age` <= ?) OR `score` <= ? got SELECT * FROM `users` WHERE (`users`.`id` <> ? AND `age` <= ?) OR `score` <= ?
```

- [X] Do only one thing
- [X] Tested

1. Adds a constant to use " OR " and " AND ", when looking for a 'or' or 'and' in a SQL statement. Also replaced `.ToLower` to `.ToUpper`, to make functionality consistent.
2. Added a few test cases that increased the code coverage to 79.7% of `where.go`

### User Case Description

When providing a `clause.Expr{}` and includes a word that contains `or` within it (like `score`), it should not enclose "\`score\` <= ?" in parentheses "(\`score\` <= ?)"

Here is one of the test case that were added to `where_test.go`
```
		{
			[]clause.Interface{clause.Select{}, clause.From{}, clause.Where{
				Exprs: []clause.Expression{clause.Not(clause.Eq{Column: clause.PrimaryColumn, Value: "1"}, clause.Gt{Column: "age", Value: 18}), clause.Or(clause.Expr{SQL: "`score` <= ?", Vars: []interface{}{100}, WithoutParentheses: false})},
			}},
			"SELECT * FROM `users` WHERE (`users`.`id` <> ? AND `age` <= ?) OR `score` <= ?",
			[]interface{}{"1", 18, 100},
		},

```

### DIFF
--- a/clause/where.go
+++ b/clause/where.go
@@ -4,6 +4,11 @@ import (
 	"strings"
 )
 
+const (
+	AndWithSpace = " AND "
+	OrWithSpace  = " OR "
+)
+
 // Where where clause
 type Where struct {
 	Exprs []Expression
@@ -26,7 +31,7 @@ func (where Where) Build(builder Builder) {
 		}
 	}
 
-	buildExprs(where.Exprs, builder, " AND ")
+	buildExprs(where.Exprs, builder, AndWithSpace)
 }
 
 func buildExprs(exprs []Expression, builder Builder, joinCond string) {
@@ -35,7 +40,7 @@ func buildExprs(exprs []Expression, builder Builder, joinCond string) {
 	for idx, expr := range exprs {
 		if idx > 0 {
 			if v, ok := expr.(OrConditions); ok && len(v.Exprs) == 1 {
-				builder.WriteString(" OR ")
+				builder.WriteString(OrWithSpace)
 			} else {
 				builder.WriteString(joinCond)
 			}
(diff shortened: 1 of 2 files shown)


sammyrnycreal:
@daheige Not sure if you will be able to review this PR? Let me know if it makes sense or if I need to make an adjustment.

daheige:
For this PR，LGTM！That's a good suggestion. I agree.


From: sammyrnycreal ***@***.***&gt;
Date: Tue,Feb 15,2022 5:40 AM
Cc: heige ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [go-gorm/gorm] Fixed the use of "or" to be " OR ", to account for words that contain… (PR #5074)





 
@daheige Not sure if you will be able to review this PR? Let me know if it makes sense or if I need to make an adjustment.
 
—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

clause/where.go
```diff
@@ -165,8 +170,8 @@ func (not NotConditions) Build(builder Builder) {
 			builder.WriteString("NOT ")
 			e, wrapInParentheses := c.(Expr)
 			if wrapInParentheses {
-				sql := strings.ToLower(e.SQL)
-				if wrapInParentheses = strings.Contains(sql, "and") || strings.Contains(sql, "or"); wrapInParentheses {
```
huacnlee, line 169: There is not need to change check with space.

Just use `strings.Contains(sql, "AND") || strings.Contains(sql, "OR")` is better

sammyrnycreal: Sorry, @huacnlee, but it seems that it is necessary, as GORM has it coded in this manner in multiple places. So I'm just wanting to make it consistent with the rest of the functional code.

References for:
" OR ":
https://github.com/go-gorm/gorm/blob/19ac396a22668e2cdbd77a262de84478787989d0/clause/where.go#L38
https://github.com/go-gorm/gorm/blob/19ac396a22668e2cdbd77a262de84478787989d0/clause/where.go#L134
https://github.com/go-gorm/gorm/blob/19ac396a22668e2cdbd77a262de84478787989d0/clause/where.go#L137

" AND ":
https://github.com/go-gorm/gorm/blob/19ac396a22668e2cdbd77a262de84478787989d0/clause/where.go#L29
https://github.com/go-gorm/gorm/blob/19ac396a22668e2cdbd77a262de84478787989d0/clause/where.go#L113
https://github.com/go-gorm/gorm/blob/19ac396a22668e2cdbd77a262de84478787989d0/clause/where.go#L116
https://github.com/go-gorm/gorm/blob/19ac396a22668e2cdbd77a262de84478787989d0/clause/where.go#L159

The reason that adding the spaces on both ends of the `and` and `or` as requirement is because, if one is scanning a string that contains "score" as a field, as in the test case, this logic will enclose the field in Parentheses. Which is unnecessary, and only occurs when the type is `clause.Expr`.

e.g.,
```
clause.And(clause.Expr{SQL: "`score` <= ?", Vars: []interface{}{100}, WithoutParentheses: false})
```

For background, my team had found this bug when we created our sqlMocks and had to write exception mocks when testing with a db field with the name of `category`. And yes, this is for selfish reasons, if you apply my fix to master, I can simplify my mocks and not have an odd `if field name contains 'or' or 'and' enclose it in Parentheses`.

Thanks in Advance for your understanding.

huacnlee: I see.

sammyrnycreal: @huacnlee @daheige can we proceed and have this merged into master for a future release?

jinzhu:
Thank you for your PR.

PR #5080, opened 2022-02-17
Provides dynamic parameter merge queries

gswy:
- [✅] Do only one thing
- [✅] Non breaking API changes
- [✅] Tested

Provides a new API, the `When` method, which requires a lot of judgment to find that in the dynamic filter conditions; the `When` method can make judgments and call to execute the query that meets the conditions.

### User Case Description
```golang
user := &User{}
db.When(true, func(query *gorm.DB) {
    query.Where("gender = 1")
}).First(user)
```

### DIFF
--- a/chainable_api.go
+++ b/chainable_api.go
@@ -150,6 +150,15 @@ func (db *DB) Where(query interface{}, args ...interface{}) (tx *DB) {
 	return
 }
 
+// When Condition established
+func (db *DB) When(expr bool, query func(tx *DB)) (tx *DB) {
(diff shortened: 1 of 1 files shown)


li-jin-gou:
use `scope` or `SubQuery` also can slove it.
- https://gorm.cn/docs/scopes.html
- https://gorm.cn/docs/advanced_query.html#SubQuery

gswy:
一些单体条件可以使用when，借鉴的laravel orm中的when

PR #5108, opened 2022-02-23
feat: support gob serialize

ag9920:
- [x] Do only one thing
- [x] Tested

Implement a GobSerializer in schema/serializer to support Gob encoding in the latest released serializer tag feature. 

### User Case Description

In the past, if we want to set the value of a struct into a certain field, we would marshal the struct to a string/bytes(normally with json/msgpack/gob). With the support for serializer, it becomes a relatively easy job. 

But now it only supports json and unixtime. Golang provide native support for gob, which has a better performance over json. Hopefully, this will help developers reduce the trouble on writing gob encoding and decoding.

### DIFF
--- a/schema/serializer.go
+++ b/schema/serializer.go
@@ -1,11 +1,12 @@
 package schema
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
 	"database/sql/driver"
+	"encoding/gob"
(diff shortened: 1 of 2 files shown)


jinzhu:
LGTM, thank you for your PR.

PR #5132, opened 2022-03-07
fix: soft delete for join

### LINKED ISSUE #4918: Joins() doesn't respect soft deletes
## GORM Playground Link

https://github.com/go-gorm/playground/pull/412

## Description

If you soft-delete a Has-One relationship, the deleted association won't be loaded if you use `Preload()`. If you use `Joins()`, though, it won't respect the soft deletion.

```
  var account Account{Number:"77"}
  DB.Create(&account)
  user := User{ Name: "jinzhu", Account: account}
  DB.Create(&user)
  
  DB.Delete(&account)
  
  // Works, no account in User
  DB.Preload("Account").First(&result, user.ID)
  
  // Doesn't work, account is hydrated even though it's been soft-deleted
  DB.Joins("Account").First(&result, user.ID)
```

a631807682:
- [x] Do only one thing
- [x] Tested

Soft delete should work for Joins.ON.Exprs.
close #4918

### DIFF
--- a/soft_delete.go
+++ b/soft_delete.go
@@ -84,11 +84,12 @@ func (sd SoftDeleteQueryClause) ModifyStatement(stmt *Statement) {
 	}
 
 	// Modify for Joins[i].ON exprs
-	if _, ok := stmt.Clauses["soft_delete_join_enabled"]; !ok && !stmt.Statement.Unscoped {
-		if c, ok := stmt.Clauses["FROM"]; ok && len(stmt.Joins) > 0 {
+	if c, ok := stmt.Clauses["FROM"]; ok && len(stmt.Joins) > 0 && sd.Field.Schema != nil {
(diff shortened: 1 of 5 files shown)


soft_delete.go
```diff
@@ -82,6 +82,24 @@ func (sd SoftDeleteQueryClause) ModifyStatement(stmt *Statement) {
 		}})
 		stmt.Clauses["soft_delete_enabled"] = clause.Clause{}
 	}
+
+	// Modify for Joins[i].ON exprs
+	if _, ok := stmt.Clauses["soft_delete_join_enabled"]; !ok && !stmt.Statement.Unscoped {
```
a631807682, line 87: It doesn't seem like a good way to update an join exprs here,  but it fits the function name and doesn't break the changes.
If break changes are allowed, i think we can create a new interface for custom `QueryClauses` to get right `Joins.ON.Exprs`

jinzhu:
Thank you for your PR.

PR #5135, opened 2022-03-07
fix: Omit 'Reference' and 'Reference.*' should be same

### LINKED ISSUE #5058: [mysql] Omit("SomeReference.*") does empty insert for references
## GORM Playground Link

https://github.com/go-gorm/playground/pull/435

## Description

Gorm v1.22.5

For objects with a "has one" or "belongs to" relationship, using the `Omit("ObjectName.*")` notation will cause an empty record to be inserted into the database. You see the raw sequel:

```
[1.648ms] [rows:1] INSERT INTO `users` VALUES() ON DUPLICATE KEY UPDATE `id`=`id`
[1.484ms] [rows:1] INSERT INTO `accounts` VALUES() ON DUPLICATE KEY UPDATE `user_id`=VALUES(`user_id`)
```

This is not great as a) you get empty rows dangling around and b) will fail if you have a non-null constraint or unique constraint.

The docs have the following:

```
For many2many associations, GORM will upsert the associations before creating the join table references, if you want to skip the upserting of associations, you could skip it
```

The docs say to use this notation to avoid upserting associations. And while this does specifically call out `many2many` associations, after looking existing issues reported, it has been suggested to use this notation for other relationship types as well.

Note that using `Omit("Reference")` will cause mysql to try to insert a null column into the db at the time of creation, which is not desired.

a631807682:
- [x] Do only one thing
- [x] Tested

```go
	DB.Omit("Account,Manager").Create(&user)	 // work
	DB.Omit("Account.*,Manager.*").Create(&user) // not work 
```
Seems `Omit("Reference")` work but `Omit("Reference.*")` doesn't, they should be the same.

close #5058

### DIFF
--- a/callbacks/associations.go
+++ b/callbacks/associations.go
@@ -348,6 +348,10 @@ func saveAssociations(db *gorm.DB, rel *schema.Relationship, values interface{},
 		refName        = rel.Name + "."
 	)
 
+	if _, ok := selectColumns[rel.Name+".*"]; ok {
+		return nil
+	}
+
 	for name, ok := range selectColumns {
 		columnName := ""
 		if strings.HasPrefix(name, refName) {
--- a/statement.go
+++ b/statement.go
@@ -674,6 +674,7 @@ func (stmt *Statement) SelectAndOmitColumns(requireCreate, requireUpdate bool) (
 
 	// omit columns
 	for _, omit := range stmt.Omits {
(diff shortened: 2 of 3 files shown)


jinzhu:
It has different meaning, refer https://gorm.io/docs/associations.html#Skip-Auto-Create-Update

image

PR #5140, opened 2022-03-08
fix: circular reference save

### LINKED ISSUE #4935: Circular reference in model causes app freeze
## GORM Playground Link
https://github.com/go-gorm/playground/pull/420

## Description
Hey there! I've just started using Gorm and finding it really great, but I quickly ran into the following issue:

I have two models that reference each other (A -> B and B -> A).

The database modelling supports this, however when my models are configured in this way, a save on the model ends up causing the app to freeze.

## Expected result
The model that is being saved is indeed saved, and the application does not freeze.

Thanks for your help!

a631807682:
- [x] Do only one thing
- [x] Tested

Circular reference in model causes app freeze, should stop `saveAssociations` loop if relation model has been created.
close #4935

### DIFF
--- a/callbacks/associations.go
+++ b/callbacks/associations.go
@@ -1,7 +1,6 @@
 package callbacks
 
 import (
-	"fmt"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


jinzhu:
if a user has many children, and only one of them have been saved, then it won't save other children?  looks like an issue?

a631807682:
> if a user has many children, and only one of them have been saved, then it won't save other children? looks like an issue?

I will create test for them.

a631807682:
I create a test for `LoadOrStoreVisitMap`.
If one of them haven't been saved, then all children will be save again.
I don't know if I should change `values` to exclude saved children,  because it looks like `values` is a reference and will be changed by `tx.Create(values)`
@jinzhu

Issue #5142, opened 2022-03-08
Gorm V2 cannot handle JOIN for tables with the same column name


Joker666:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/443

## Description

If two tables have the same column names like `id` and `name`, once we perform the JOIN query with a junction table in between, we see that the library does not populate the right value into the right struct. This happens due to how value is scanned back to the struct.

I have developed a fix for it inspired by Gorm V1. Will send a PR for it.

Joker666:
Fixed with https://github.com/go-gorm/gorm/pull/5143

Joker666:
This issue was fixed by https://github.com/go-gorm/gorm/pull/5143 but however, @jinzhu modified my commit to be more concise which resulted in a bug. That's why opening this issue again and going to send PR to fix the bug as well.

PR #5143, opened 2022-03-08
Added offset when scanning the result back to struct

Joker666:
- [x] Do only one thing
- [x] Tested

The solution is inspired by Gorm V1. To avoid always picking the same value from the `LookUpField` method we create a `fields` array and keep track of already picked values using an offset. In Gorm 1, this was the default way to scan results back.

close https://github.com/go-gorm/gorm/issues/5142

This fixes the demo PR for the playground - https://github.com/go-gorm/playground/pull/443

### User Case Description

The issue in question occurs because we are using the `LookUpField` method of the `schema` package. This method just looks for the `name` in the map and returns the first `field` it gets. So we get this scenario where two tables have the same column names and we do a `JOIN` query, it returns values that might be from a different table to a struct that should not have it.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -156,10 +156,12 @@ func Scan(rows *sql.Rows, db *DB, mode ScanMode) {
 		}
 	default:
 		var (
-			fields       = make([]*schema.Field, len(columns))
-			joinFields   [][2]*schema.Field
-			sch          = db.Statement.Schema
-			reflectValue = db.Statement.ReflectValue
+			fields             = make([]*schema.Field, len(columns))
+			selectFields       []*schema.Field
+			selectedColumnsMap = map[string]int{}
+			joinFields         [][2]*schema.Field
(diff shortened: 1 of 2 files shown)


Joker666:
@huacnlee please take a look

jinzhu:
Hello @Joker666 

Can you add some tests?

Joker666:
@jinzhu Added a test. But I am not sure why this check has failed though.

https://github.com/go-gorm/gorm/runs/5538082130?check_suite_focus=true

Haven't worked on anything for the `Coupon` model. And shows permission issue. Weird

Joker666:
Hello @jinzhu , could you tell me if there was any issue with the implementation?

Joker666:
Hi, @jinzhu I missed that. Thanks for merging. I just deleted my comment once I found out that it got merged and you commented.

So, I'm closing the issue as well.

PR #5155, opened 2022-03-14
fix when index name is "type", parseFieldIndexes will set index TYPE is "TYPE"

labulakalia:
- [x] Do only one thing
- [x] Tested

```
Name7        string `gorm:"index:type"``
```
call `AutoMigrate`,when index name is type, `parseFieldIndexes` will parse index TYPE is "TYPE", because `ParseTagSetting`'s params is `type`, it will return `map[TYPE]TYPE`,,so we should exclude index name in call `ParseTagSetting`'s params

error generate sql 
```
`CREATE INDEX `type` ON `xxxx`(`xxx`,`type`) USING TYPE`
```

### User Case Description
```
Error 1064: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'TYPE' at line 1", "elapsed": 0.000959069, "rows": 0, "sql": "CREATE INDEX `type` ON `xxxxxxxxx`(`xxxxx`,`type`) USING TYPE
```

https://github.com/go-gorm/playground/pull/446

### DIFF
--- a/schema/index.go
+++ b/schema/index.go
@@ -89,11 +89,12 @@ func parseFieldIndexes(field *Field) (indexes []Index) {
 			k := strings.TrimSpace(strings.ToUpper(v[0]))
 			if k == "INDEX" || k == "UNIQUEINDEX" {
 				var (
-					name      string
-					tag       = strings.Join(v[1:], ":")
-					idx       = strings.Index(tag, ",")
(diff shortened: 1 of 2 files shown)


labulakalia:
https://github.com/go-gorm/playground/runs/5532362884?check_suite_focus=true

Issue #5176, opened 2022-03-19
v1.23.3 ,there is bug that: [error] failed to parse default value `CURRENT_TIMESTAMP` for field xxxxx


iGoogle-ink:
![image](https://user-images.githubusercontent.com/17245594/159131406-8b656e5e-4fa1-4dd1-bf11-0d5a880538e6.png)

when   db.Create() , 
[error] failed to parse default value `CURRENT_TIMESTAMP` for field xxxxx

iGoogle-ink:
`		if field.HasDefaultValue && !skipParseDefaultValue && field.DataType == Time {
			if field.DefaultValueInterface, err = now.Parse(field.DefaultValue); err != nil {
				schema.err = fmt.Errorf("failed to parse default value `%v` for field %v", field.DefaultValue, field.Name)
			}
		}`

这个地方有可能出现Bug，当 default_value  写  `CURRENT_TIMESTAMP ` 时

PR #5177, opened 2022-03-19
fix bug that default:CURRENT_TIMESTAMP

iGoogle-ink:
fix the bug  #5176


jinzhu:
Fixed in https://github.com/go-gorm/gorm/commit/0097b39a77b9573d63f89c22f3cea0aae103a77f 

There are some other SQL datetime funcs like `now()` also need to check, so we just ignored all error.

jinzhu:
Thank you for your PR.

xuanskyer:
@jinzhu

PR #5193, opened 2022-03-27
fix: BeforeCreate, db.Statement.ReflectValue不是指针型的时候AfterCreate, Befo…

johnnylei:
fix: BeforeCreate, db.Statement.ReflectValue不是指针型的时候AfterCreate, BeforeUpdate, AfterUpdate会panic

比如下面这种
```
type OperationStep {
    ID int64
    State uint
}

func (o *OperationStep) AfterUpdate(tx *gorm.DB) error {
	return nil
}

db.Model(OperationStep{}).Where("id=?", operationStep.ID).Update("state", types.OperationStateSuccess).Error
```

### DIFF
--- a/callbacks/callbacks.go
+++ b/callbacks/callbacks.go
@@ -24,6 +24,10 @@ func RegisterDefaultCallbacks(db *gorm.DB, config *Config) {
 		return !db.SkipDefaultTransaction
 	}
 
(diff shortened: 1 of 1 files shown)


jinzhu:
`Callbacks` are global shared, and the `Match` func only checked when registering

PR #5201, opened 2022-03-29
refactor: provides error for finisher api updateable vaild

### LINKED ISSUE #5197: Provides friendlier errors for finisher api
## Describe the feature

Provide more friendly errors for those finisher api that need to modify the passed parameters.

## Motivation

It will panic if we pass not updatable param to finisher api like `Create/CreateInBatches/Save...`
```go
user := User{Name: "123"}
DB.Where("id=?", "1").Create(user)
```
It will execute sql but no error and dose not update param if we pass not updatable param to finisher api like `Scan/Find/...`
```go
var user User
DB.Model(User{}).Where("id=?", "1").Scan(user)
```
So I think we should check and throw more friendly errors at first.

a631807682:
- [x] Do only one thing
- [x] Tested

close #5197

This pr will verfiy finisher api which params will be updated by gorm. 
It will not verfiy other api like `Updates`, It will still panic when using hooks like other issues say.

### DIFF
--- a/tests/create_test.go
+++ b/tests/create_test.go
@@ -366,7 +366,7 @@ func TestCreateInvalidSlice(t *testing.T) {
 		nil,
 	}
 
-	if err := DB.Create(&users).Error; !errors.Is(err, gorm.ErrInvalidData) {
+	if err := DB.Create(&users).Error; !errors.Is(err, gorm.ErrInvalidValue) {
 		t.Errorf("should returns error invalid data when creating from slice that contains invalid data")
 	}
 }
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -20,6 +20,7 @@ func (db *DB) Create(value interface{}) (tx *DB) {
 	}
 
 	tx = db.getInstance()
+	tx.vaildUpdateableParam(value)
 	tx.Statement.Dest = value
 	return tx.callbacks.Create().Execute(tx)
 }
@@ -32,6 +33,7 @@ func (db *DB) CreateInBatches(value interface{}, batchSize int) (tx *DB) {
 	case reflect.Slice, reflect.Array:
 		var rowsAffected int64
 		tx = db.getInstance()
+		tx.vaildUpdateableParam(value)
 
 		callFc := func(tx *DB) error {
 			// the reflection length judgment of the optimized value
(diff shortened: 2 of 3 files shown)


tests/create_test.go
```diff
@@ -366,7 +366,7 @@ func TestCreateInvalidSlice(t *testing.T) {
 		nil,
 	}
 
-	if err := DB.Create(&users).Error; !errors.Is(err, gorm.ErrInvalidData) {
+	if err := DB.Create(&users).Error; !errors.Is(err, gorm.ErrInvalidValue) {
```
a631807682, line 369: This test error change form `ErrInvalidData` to `ErrInvalidValue`, seems not be a breaking change.

jinzhu:
Maybe we can remove all `CanAddr` check, and use recover in `(*processor).Execute` method to recover those panics and set the error

a631807682:
> Maybe we can remove all `CanAddr` check, and use recover in `(*processor).Execute` method to recover those panics and set the error

I think recover in `(*processor).Execute`  is dangerous.
1. there is no guarantee that all panic comes from this situation, this may make some bugs harder to spot.
2. `Execute` run callback function executed contains panic registered by the user that we have no control.

Or we can register a vaild callback for them.
1. vaild `CanAddr` for `createCallback` and `queryCallback`
2. vaild `CanAddr` if `schema` defined hooks for `deleteCallback` and `queryCallback`

jinzhu:
> > Maybe we can remove all `CanAddr` check, and use recover in `(*processor).Execute` method to recover those panics and set the error
> 
> I think recover in `(*processor).Execute` is dangerous.
> 
>     1. there is no guarantee that all panic comes from this situation, this may make some bugs harder to spot.
> 
>     2. `Execute` run callback function executed contains panic registered by the user that we have no control.

We can check the panic info match the error or not, and we can reraise other panics.

Just would like to avoid affect performance due to those rare cases.

a631807682:
Sounds like we need to check panic info if contained `CanAddr`, and we have a whitelist to check for internal panic not user's, also avoid `CanAddr` panic caused by dependent packages used in internal methods.
But this still makes it difficult to find errors that are hidden because there is no records, like `Scan/Find`

None of it looks particularly good, maybe we just need to receive only pointer parameter in the next incompatible version so we don't need to add any checks or recover panic.

Issue #5220, opened 2022-04-01
"hasMany" only insert one row


kaisawind:
related issues
#5126 gorm最新版Has Many只写入了关联数组的第一个 
#4796 pgsql: ON CONFLICT DO UPDATE command cannot affect row a second time
#4890 Regression in 1.22.4: Storing int PKs with value zero

I think it should be
```go
cacheKey := utils.ToStringKey(relPrimaryValues...) 
```
not 
https://github.com/go-gorm/gorm/blob/f7b52bb649ba803ec149a06fec9e9da7b311d36e/callbacks/associations.go#L209

and **when there's no primary keys**, we should append elems

```go
cacheKey := utils.ToStringKey(relPrimaryValues...)
if len(relPrimaryValues) == 0 || len(relPrimaryValues) != len(rel.FieldSchema.PrimaryFields) || !identityMap[cacheKey] {
        identityMap[cacheKey] = true
        if isPtr {
	        elems = reflect.Append(elems, elem)
        } else {
	        elems = reflect.Append(elems, elem.Addr())
        }
}
``` 
not
https://github.com/go-gorm/gorm/blob/f7b52bb649ba803ec149a06fec9e9da7b311d36e/callbacks/associations.go#L210

kaisawind:
https://github.com/go-gorm/gorm/pull/5221#issuecomment-1086824732

i give it up.

PR #5221, opened 2022-04-01
fix: "hasMany" only insert one row(#5220)

kaisawind:
- [X] Do only one thing
- [X] Tested

when there's no primary keys, we should append elems

### User Case Description

has many relationship without primary keys


jinzhu:
Hi @kaisawind 

Can you add some tests?

kaisawind:
> Hi @kaisawind
> 
> Can you add some tests?

I give it up.
 there has many errors when without primary key only unique key.
append will append old values.
replace will replace all, nothing left.

/close it

PR #5226, opened 2022-04-04
fix variable shadowing in finisher_api

ZhangShenao:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


jinzhu:
is it causing any issue? seems good to me?

PR #5227, opened 2022-04-04
Offset issue resolved for scanning results back into struct

Joker666:
- [x] Do only one thing
- [x] Tested

This PR fixes an accidentally introduced bug while merging PR https://github.com/go-gorm/gorm/pull/5143
The modified commit https://github.com/go-gorm/gorm/commit/f3e2da5ba359f0d672249fc52f54ae41c5a66d3a introduced an offset bug which resulted in `nil` for scanning results back into the struct.

It's my bad, I did not write extensive test cases which will catch this bug. Now I've updated test cases to account for all known cases that can happen.

close https://github.com/go-gorm/gorm/issues/5142

This fixes the demo PR for the playground - https://github.com/go-gorm/playground/pull/460

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -196,7 +196,7 @@ func Scan(rows Rows, db *DB, mode ScanMode) {
 				for idx, column := range columns {
 					if field := sch.LookUpField(column); field != nil && field.Readable {
 						if curIndex, ok := selectedColumnsMap[column]; ok {
-							for fieldIndex, selectField := range sch.Fields[curIndex:] {
+							for fieldIndex, selectField := range sch.Fields[curIndex+1:] {
 								if selectField.DBName == column && selectField.Readable {
 									selectedColumnsMap[column] = curIndex + fieldIndex + 1
 									fields[idx] = selectField
--- a/tests/scan_test.go
+++ b/tests/scan_test.go
@@ -184,11 +184,34 @@ func TestScanToEmbedded(t *testing.T) {
 		t.Errorf("Failed to run join query, got error: %v", err)
 	}
 
+	personMatched := false
+	addressMatched := false
+
 	for _, info := range personAddressInfoList {
-		if info.Person != nil {
-			if info.Person.ID == person1.ID && info.Person.Name != person1.Name {
+		if info.Person == nil {
+			t.Fatalf("Failed, expected not nil, got person nil")
+		}
+		if info.Address == nil {
+			t.Fatalf("Failed, expected not nil, got address nil")
+		}
+		if info.Person.ID == person1.ID {
+			personMatched = true
+			if info.Person.Name != person1.Name {
 				t.Errorf("Failed, expected %v, got %v", person1.Name, info.Person.Name)
 			}
 		}
+		if info.Address.ID == address1.ID {
+			addressMatched = true
+			if info.Address.Name != address1.Name {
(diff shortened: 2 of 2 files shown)


Joker666:
Hmm, the test `TestSmartMigrateColumn` is failing for `Postgres` dialect. I see the same issue for https://github.com/go-gorm/gorm/pull/5226

I think the issue is somewhere else

scan.go
```diff
@@ -196,7 +196,7 @@ func Scan(rows Rows, db *DB, mode ScanMode) {
 				for idx, column := range columns {
 					if field := sch.LookUpField(column); field != nil && field.Readable {
 						if curIndex, ok := selectedColumnsMap[column]; ok {
-							for fieldIndex, selectField := range sch.Fields[curIndex:] {
+							for fieldIndex, selectField := range sch.Fields[curIndex+1:] {
```
clarkmcc, line 199: @jinzhu Is there a reason this was merged when it failed to pass on Postgres? We're currently using this on CockroachDB (Postgres protocol compatible), after upgrading to 1.23.5 we started getting this error:
```
panic: runtime error: slice bounds out of range [11:10]
```

After downgrading to 1.23.4, the issue went away. We were doing a simple scan from a raw query
```go
tx.Raw("...").Scan(&rows).Error
```

If we need to put together a test case, I can take the time, however I'm curious why this was merged when it caused the Postgres tests to fail?

Joker666: @clarkmcc the tests for Postgres were failing for a different case, specifically `TestSmartMigrateColumn` which did not invoke any `scan` code from here. And the failure was intermittent. All the tests do pass with the code merged in with the latest test runs.

However, I can clearly see how the panic is occurring. The fix would be having a check that the index exists.

clarkmcc: I'm not sure if this could explain the panic, but after further investigation, we could only reproduce it on this query which performs an inner join on the same table as the original select, which means the resulting recordset had a `parent` and a `children` copy of each column.
```sql
SELECT * FROM entity parent
    INNER JOIN entity children ON children.genealogy @> ARRAY[parent.entity_key]
WHERE parent.entity_key = @entity_key;
```

Joker666: There is a PR ready for a duplicate issue fix on a `distinct` query. https://github.com/go-gorm/gorm/pull/5369

See if that resolves your issue. I don't think it will. But you can run a fork with the check that the index exists before running the `for` loop and see if that resolves your issue. Otherwise, would have to dig to see what's causing the panic and also need a test case for it.

Joker666: After thinking a bit more about the issue, I believe you are right. It can occur for inner join on the same table.

What this PR original sought to solve was, that if there are two tables with identical column names, before this fix, the fields would be returned with the same values across different tables while scanning the result back to the struct.

So this offset fix was picked right from GORM V1 where this used to work https://github.com/jinzhu/gorm/blob/master/scope.go#L491

But running inner join on the same table would cause it to panic since the number of fields are the same. So the fix would be to not offset when running join on the same table.

a631807682: It works because the defined schema is consistent with the database fields, when we use raw and scan, it is often inconsistent, which causes the array to go out of bounds.
I think we can skip this part because we can't determine whether we want to use the value of `children` or `parent` in this case.

PR #5234, opened 2022-04-09
test: test for postgrs serial column

a631807682:
- [x] Do only one thing
- [x] Tested

test for https://github.com/go-gorm/postgres/pull/98


Genora51:
Should that be `Serial` rather than `Sercial`? Likewise on https://github.com/go-gorm/postgres/pull/98.

a631807682:
@Genora51 thank you for review

PR #5241, opened 2022-04-10
Fix scanIntoStruct

delmorof:
- [x] Do only one thing
- [x] Tested

If there are more joins in the query, the scanIntoStruct function stops loading the data as soon as a null association is encountered. The test reproduces the case in question


jinzhu:
thank you for your PR.

PR #5255, opened 2022-04-14
fix: FindInBatches with offset limit

### LINKED ISSUE #5252: `FindInBatches` in an unexpected manner with `limit` or `offset` in original query
## GORM Playground Link

It's a reversed test, which means the CI SHOULD FAIL instead of success.

https://github.com/go-gorm/playground/pull/466

## Description

The use case is pretty simple. Consider we may have 1k user records, and we want to pre-filter 500 records by like, last login time, then process them in 5 per 100 records batch. The query use `FindInBatches` will be like:

```go
var singleBatch []User
dbInstance.Limit(500).Offset(100).Where("").FindInBatches(&singleBatch, 100, func(tx *gorm.DB, batch int) error {
  // Some operations over singleBatch
})
```

When using the `Limit` and `Offset` clauses, due to the implementation of `.FindInBatches`, it will cause different issues(Both are covered in the playground):
- `Limit`: The inner `Limit` by `.FindInBatches` will overwrite the originial query, which will fetch all records instead of the desired set.
- `Offset`: The parameter will be applied in each query, which will skip certain entries over the desired results.

By design, it leads to two solutions:

1. If it's the intended behavior, we may should add a *warning* section in the doc to tell the user not to use it with `limit` or `offset` clauses, or even refactor the API to avoid misusing.
2. If the pre-filtered query should be supported, we may need one or more queries overhead(s) to properly get the query sets since the `where` clause can be pretty complex and it's hard to handle them(e.g., a query will make the `id` or primary key field not continous).

a631807682:
- [x] Do only one thing
- [x] Tested

I think when the user uses `Offset` in `FindInBatches` the meaning is to skip the previous N records, and user uses `LIMIT` in `FindInBatches` the meaning is find total of N records.

close #5252

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -181,6 +181,21 @@ func (db *DB) FindInBatches(dest interface{}, batchSize int, fc func(tx *DB, bat
 		batch        int
 	)
 
+	// user specified offset or limit
+	var totalSize int
+	if c, ok := tx.Statement.Clauses["LIMIT"]; ok {
+		if limit, ok := c.Expression.(clause.Limit); ok {
+			totalSize = limit.Limit
+
+			if batchSize > totalSize {
+				batchSize = totalSize
+			}
+
+			// reset to offset to 0 in next batch
+			tx = tx.Offset(-1).Session(&Session{})
+		}
+	}
+
 	for {
 		result := queryDB.Limit(batchSize).Find(dest)
 		rowsAffected += result.RowsAffected
@@ -194,6 +209,16 @@ func (db *DB) FindInBatches(dest interface{}, batchSize int, fc func(tx *DB, bat
 
 		if tx.Error != nil || int(result.RowsAffected) < batchSize {
 			break
+		}
+
+		if totalSize > 0 {
+			if totalSize/batchSize == batch {
+				batchSize = totalSize % batchSize
+			}
+
+			if totalSize <= int(rowsAffected) {
+				break
+			}
 		}
 
 		// Optimize for-break
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -187,7 +187,7 @@ func (db *DB) FindInBatches(dest interface{}, batchSize int, fc func(tx *DB, bat
 		if limit, ok := c.Expression.(clause.Limit); ok {
 			totalSize = limit.Limit
 
-			if batchSize > totalSize {
+			if totalSize > 0 && batchSize > totalSize {
 				batchSize = totalSize
 			}
 
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -212,12 +212,11 @@ func (db *DB) FindInBatches(dest interface{}, batchSize int, fc func(tx *DB, bat
 		}
 
 		if totalSize > 0 {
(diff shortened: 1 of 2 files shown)


finisher_api.go
```diff
@@ -181,6 +181,21 @@ func (db *DB) FindInBatches(dest interface{}, batchSize int, fc func(tx *DB, bat
...
+	if c, ok := tx.Statement.Clauses["LIMIT"]; ok {
+		if limit, ok := c.Expression.(clause.Limit); ok {
+			totalSize = limit.Limit
+
+			if batchSize > totalSize {
+				batchSize = totalSize
+			}
+
+			// reset to offset to 0 in next batch
+			tx = tx.Offset(-1).Session(&Session{})
```
lxdlam, line 195: Should we extract this call to another check? The `offset` is also buggy when using w/o `limit`.

a631807682: > Should we extract this call to another check? The `offset` is also buggy when using w/o `limit`.

What does `offset is also buggy` mean ? `FindInBatches` query condition is greater than primary key and limit.

lxdlam: > > Should we extract this call to another check? The `offset` is also buggy when using w/o `limit`.
> 
> What does `offset is also buggy` mean ? `FindInBatches` query condition is greater than primary key and limit.

If we only using `offset`, which scenario will be like *select all users which uid is larger than 100*, then using `dbInstance.Offset(100).OrderBy("user_id").FindInBatches()`, it will skip 100 records in each of all subsequent queries.

a631807682: it will skip 100 records in first subsequent query.
here is reset offset to zero after first subsequent query that we can use `id > ? and limit = ?` to query.

lxdlam: > it will skip 100 records in first sub query. here is reset offset to zero after first sub query that we can use `id > ? and limit = ?` to query.

That's correct. But this call just been embedded in the `limit` check. I mean, we may need to extract it to another check that if we have a `Offset` clause and the argument is large than `0` so we can both support only using `Offset` or togethered with `Limit` clause.

a631807682: `Offest` and `Limit` use the same `clause.Limit`, so it support for only using `Offset` or togethered
https://github.com/go-gorm/gorm/blob/master/chainable_api.go#L245

a631807682: Only `Offset` cannot be used is caused by [here](https://github.com/go-gorm/gorm/pull/5255/commits/f9a86e9148fab7af0f932d8c1350683087d794a4#diff-aaa805d49ddced2f2281789039fef3bdb2fca44a51914d2c9b05ea3bd959776aR190), thank you.

lxdlam: > Only `Offset` cannot be used is caused by [here](https://github.com/go-gorm/gorm/pull/5255/commits/f9a86e9148fab7af0f932d8c1350683087d794a4#diff-aaa805d49ddced2f2281789039fef3bdb2fca44a51914d2c9b05ea3bd959776aR190), thank you.

Got it. You're nice :).

Issue #3973, opened 2021-01-20
Defining composite unique index in embedded struct


arcln:
## Your Question

```go
type Foo struct {
  IndexA int `gorm:"index:my_idx,unique"`
  IndexB int `gorm:"index:my_idx,unique"`
}

type Bar struct {
  Foo `gorm:"embedded"`
}
```

In this example, both structs can be used as models and have a unique composite index named "my_idx".
However, when running auto-migrate of both of them, I get the following error from postgres (and a similar one from sqlite) :

```
ERROR: relation "my_idx" already exists
[1.361ms] [rows:0] CREATE INDEX "my_idx" ON "bars" ("index")
```

Notice that the error only happens for the model that is auto-migrated last. I understand that I can't have two relations with the same name. How can I define a model with a composite unique index and use it both as is and as an embedded struct?

## The document you expected this should be explained

https://gorm.io/docs/indexes.html

## Expected answer

I believe this is expected and not handled by gorm because of the complexity and rarity of this use case.
Maybe you can suggest me a workaround? Thanks.

A "real" solution would be to prefix the index name with the table name. Not sure if this is complicated or will break anything.

jinzhu:
You can make it works by changing the default `NamingStrategy` and change the generated index name.

arcln:
Actually this didn't solve the problem, because of [schema/index.go:107](https://github.com/go-gorm/gorm/blob/35ebfe68740ef8d1ff3fde2037fbba34d802e287/schema/index.go#L107) not calling `IndexName()` if a name is specified for the index. From my understanding I _must_ supply a manual name in order to create a composite index.

Changing the behaviour of `IndexName()` would be a breaking change so I suppose it isn't really an option.

I could create a messy switch statement that handles every table and index cases to return fully generated indexes, but at this point it is easier to duplicate the indexing fields.

PR #5269, opened 2022-04-18
index: add composite id

photon3108:
- [x] Do only one thing
- [x] Tested

In the issue #3973, if a struct with any composite indexes, these indexes will be invalid when the struct is embedded more than once. Because a composite index **must** specify its name, embedding more than once results in the duplicated name in db. Also if the name of index is specified manually, `NamingStrategy` will not be applied. Take the example of #3973:
```go
type Bar0 struct {
  Foo
}
type Bar1 struct {
  Foo
}
```
Both `Bar0` and `Bar1` will try to create an index of the same name `my_idx`. In this pull request, it provides a new field of index tag which is named to `composite`. It means the id of composite index. All fields which have the same composite id of the struct are put together to the same index, just like the original rule. But the improvement is it lets the most derived/embedding struct generates the name of index by `NamingStrategy`. For example:
```go
type Foo struct {
  IndexA int `gorm:"index:,unique,composite:myname"`
  IndexB int `gorm:"index:,unique,composite:myname"`
}
```
If the table `Foo` is created, the name of composite index will be `idx_foo_myname`.
```go
type Bar0 struct {
  Foo
}
type Bar1 struct {
  Foo
}
```
Respectively, the name of composite index are `idx_bar0_myname` and `idx_bar1_myname`.

`composite` only works if not specify the name of index, so it keeps the original API from being broken.

### DIFF
--- a/schema/index.go
+++ b/schema/index.go
@@ -1,6 +1,7 @@
 package schema
 
 import (
+	"fmt"
 	"sort"
 	"strconv"
 	"strings"
@@ -31,7 +32,12 @@ func (schema *Schema) ParseIndexes() map[string]Index {
 
 	for _, field := range schema.Fields {
 		if field.TagSettings["INDEX"] != "" || field.TagSettings["UNIQUEINDEX"] != "" {
-			for _, index := range parseFieldIndexes(field) {
+			fieldIndexes, err := parseFieldIndexes(field)
+			if err != nil {
+				schema.err = err
+				break
+			}
+			for _, index := range fieldIndexes {
 				idx := indexes[index.Name]
 				idx.Name = index.Name
 				if idx.Class == "" {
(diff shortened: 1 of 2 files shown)


daheige (approved):

jinzhu:
Hello,

@photon3108 can you add a test to `index_test.go` to test the `composite` logic.

Thank you.

daheige:
To support this change, you still need to provide unit tests, which is more reliable.


From: Jinzhu ***@***.***&gt;
Date: Sun,Apr 24,2022 9:31 AM
Cc: heige ***@***.***&gt;, Comment ***@***.***&gt;
Subject: Re: [go-gorm/gorm] index: add composite id (PR #5269)





 
Hello,
 
@photon3108 can you add a test to index_test.go to test the composite logic.
 
Thank you.

Issue #5282, opened 2022-04-21
AutoMigrate fails if custom index is already added to the table (Duplicate column error)


glebarez:
## GORM Playground Link
https://github.com/go-gorm/playground/pull/469

## Description
AutoMigrate fails if custom index is already added to the table (Duplicate column error)

glebarez:
This is due to SQLite driver does not expect index name in ```CREATE INDEX``` to be written without any quotes.
Fix PR opened: https://github.com/go-gorm/sqlite/pull/90

LouisSayers:
Not just a Sqlite issue, I just got this in postgres as well with the latest gorm and postgres driver versions

I'm getting errors that look like:
`relation "idx_signup_requests_hash_code" already exists (SQLSTATE 42P07)`

With Versions:
gorm.io/gorm v1.23.4
gorm.io/driver/postgres v1.3.4

Have had issues with this for a number of weeks now and having to do some hacky stuff to get migrations working...

khalilsarwari:
I'm using the latest versions:

gorm.io/datatypes v1.0.7
gorm.io/driver/postgres v1.3.8
gorm.io/gorm v1.23.7

And am still running into the same issue:

`relation "idx_users_user_name" already exists (SQLSTATE 42P07)`

khalilsarwari:
I switched back to the following versions and the issue went away:

gorm.io/datatypes v1.0.2
gorm.io/driver/postgres v1.1.1
gorm.io/gorm v1.21.15

a631807682:
@khalilsarwari  It is work for me in lastest version, please provide it in https://github.com/go-gorm/playground

khalilsarwari:
@a631807682 You are right, it is also working for me even when I try 

gorm.io/datatypes v1.0.7
gorm.io/driver/postgres v1.3.8
gorm.io/gorm v1.23.7

I'm not too sure why it was giving me the same error, perhaps I was not updating the versions correctly.

PaulSonOfLars:
I can reproduce this on postgres, using versions:

```
gorm.io/driver/postgres v1.3.8 (previously 1.1.0)
gorm.io/gorm v1.23.8 (previously 1.21.13)
```

### TL;DR:
The trick is to create a table with both an index AND a constraint on the same column. This causes the automigration to fail.

### Steps:

My production database originally didn't use gorm, so has some manually created constraints (backed by indexes, because postgres)
Namely, the output of `\d+ <tablename>` shows the following indexes:
[...]

PR #5283, opened 2022-04-21
fix missing error-check in AutoMigrate

glebarez:
- [x] Do only one thing
- [x] Tested

Error check was missing in AutoMigrate.
Due to this miss, errors of underlying drivers might not be delivered, resulting in misleading error texts.
Example of such behavior is showed in https://github.com/go-gorm/gorm/issues/5282, which is ```duplicate column name: id``` but real error was in SQLite driver, namely ```invalid ddl```

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -99,7 +99,10 @@ func (m Migrator) AutoMigrate(values ...interface{}) error {
 			}
 		} else {
 			if err := m.RunWithValue(value, func(stmt *gorm.Statement) (errr error) {
-				columnTypes, _ := m.DB.Migrator().ColumnTypes(value)
+				columnTypes, err := m.DB.Migrator().ColumnTypes(value)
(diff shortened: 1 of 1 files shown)


migrator/migrator.go
```diff
@@ -99,7 +99,10 @@ func (m Migrator) AutoMigrate(values ...interface{}) error {
 			}
 		} else {
 			if err := m.RunWithValue(value, func(stmt *gorm.Statement) (errr error) {
-				columnTypes, _ := m.DB.Migrator().ColumnTypes(value)
+				columnTypes, err := m.DB.Migrator().ColumnTypes(value)
```
sammyrnycreal, line 102: this `err` is shadowing the previous one, you can remove the `:`

glebarez: It is the local error inside function literal scope, which returns error as well. Looks good to me.

PR #5289, opened 2022-04-23
Add new schema initials from the Go lint package

behroozkamali:
- [x] Do only one thing
- [x] Tested

There are new common initials added to the Go lint package, which are not included in the schema common initials. To be consistent, they are added to the schema package.

### User Case Description

This PR ensures table names that include initials are consistent with the initial accepted by the Go lint package.


jinzhu:
Hi @behroozkamali 

Sorry but we are not going to accept new common initialisms as it will break existing applications, but maybe we can add a method for `NamingStrategy` to allow to register new common initialisms, can you create a new PR for that?

Thank you.

PR #5315, opened 2022-05-02
Closes #5259 - Adds new getter to DB for getting a transactions context

### LINKED ISSUE #5259: add a function to get DB.Statement.Context easily
## Describe the feature
Add a method to get context from `DB` object easily, like this:

```
func (db *DB) Context() context.Context {
     return db.Statement.Context
}
```

## Motivation

When I use transaction，I want to get context what pass in by `WithContext` method，I can only use `tx.Statement.Context`.
```
err := db.WithContext(context.WithValue(ctx, "aaa", "bbbb")).Transaction(func(tx *gorm.DB) error {
       aaa := tx.Statement.Context.Value("aaaa").(string)
})
```

instead:
```
err := db.WithContext(context.WithValue(ctx, "aaa", "bbbb")).Transaction(func(tx *gorm.DB) error {
       aaa := tx.Context().Value("aaaa").(string)
})
```

I think it is more easily.

## Related Issues

NO

townsymush:
- [x] Do only one thing
- [x] Tested

Adds a new method to the DB struct for getting the statement context

### User Case Description

Feature Request
https://github.com/go-gorm/gorm/issues/5259

Question:
Do I need to update Gorm.io with this feature?

### DIFF
--- a/gorm.go
+++ b/gorm.go
@@ -298,6 +298,11 @@ func (db *DB) WithContext(ctx context.Context) *DB {
 	return db.Session(&Session{Context: ctx})
 }
 
+// Context returns the db.Statement.Context
+func (db *DB) Context() context.Context {
+	return db.Statement.Context
+}
+
 // Debug start debug mode
 func (db *DB) Debug() (tx *DB) {
 	return db.Session(&Session{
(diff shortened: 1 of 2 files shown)


jinzhu:
Hi @townsymush 

Thank you for your PR, but sorry, but not going to add an API for that.

For people want to use context, he better to know the statement, we don't want to create a shortcut to hide it.

PR #5325, opened 2022-05-08
fix: tag defalut NULL and default not set should be same

### LINKED ISSUE #5324: Inconsistent default value check between MySQL and MariaDB
Consider the following schema:

```go
type User {
	ID string `gorm:"primary_key"`
	Name string `gorm:"unique"`
}
```

Note that no explicit default value was set for a user's `name` column.

Moreover, consider the following part of the auto migration functionality:

https://github.com/go-gorm/gorm/blob/19b8d37ae8155667d76021e4ca3314bb571756be/migrator/migrator.go#L451-L456

`field.DefaultValue` is always `""`, however, `columnType.DefaultValue()` returns `"NULL"` on MariaDB and `""` on MySQL, as a consequence of which an `ALTER TABLE` query will be issued on MariaDB although nothing had actually changed. In combination with https://github.com/go-gorm/mysql/issues/74 and the fact that only 64 keys per table are allowed on MariaDB, this can cause a program to crash over time. 

A similar issue was already brought up in https://github.com/go-gorm/gorm/issues/5188, I think.

a631807682:
- [x] Do only one thing
- [x] Tested

The following tags settings should be the same
```go
type UniqueTest struct {
		ID   string `gorm:"primary_key"`
		Name string `gorm:"unique"`
}

type UniqueTest2 struct {
		ID   string `gorm:"primary_key"`
		Name string `gorm:"unique;default:NULL"`
}
```
The first case will create duplicate index in `mariadb` just like the second case in other db.

close #5324

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -11,6 +11,7 @@ import (
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 	"gorm.io/gorm/schema"
+	"gorm.io/gorm/utils"
 )
 
 var (
@@ -448,7 +449,8 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 	}
 
 	// check default value
-	if v, ok := columnType.DefaultValue(); ok && v != field.DefaultValue {
+	if v, ok := columnType.DefaultValue(); ok &&
+		utils.CheckColumnDefaultNull(v) != utils.CheckColumnDefaultNull(field.DefaultValue) {
 		// not primary key
 		if !field.PrimaryKey {
 			alterColumn = true
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -449,11 +449,13 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 	}
 
 	// check default value
-	if v, ok := columnType.DefaultValue(); ok &&
-		utils.CheckColumnDefaultNull(v) != utils.CheckColumnDefaultNull(field.DefaultValue) {
-		// not primary key
-		if !field.PrimaryKey {
-			alterColumn = true
+	if v, ok := columnType.DefaultValue(); ok && v != field.DefaultValue {
+		// not all equal null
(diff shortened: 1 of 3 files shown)


a631807682:
It seems that a test driver is needed one by one, because there is currently no interface to get all the indexes of the schema.

migrator/migrator.go
```diff
@@ -449,9 +450,12 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 
 	// check default value
 	if v, ok := columnType.DefaultValue(); ok && v != field.DefaultValue {
-		// not primary key
-		if !field.PrimaryKey {
-			alterColumn = true
+		// not all equal null
+		if !(utils.CheckColumnDefaultNull(v) && utils.CheckColumnDefaultNull(field.DefaultValue)) {
```
a631807682, line 454: We can also check `MariaDB` version in mysql driver to replace string "NULL" instead of using `tils.CheckColumnDefaultNull(v)`directly, but I don't think it is necessary.

a631807682:
There seems to be a problem, so it is impossible to distinguish between NULL and empty string

jinzhu:
Maybe we can skip `ALTER` if field's `DefaultValueInterface` equals nil and the default value from database is `NULL`?

a631807682:
> Maybe we can skip `ALTER` if field's `DefaultValueInterface` equals nil and the default value from database is `NULL`?

Yes, I did it originally, but here is a problem, in MariaDB, `NULL` and `Empty String` will return `NULL` and `''`, but in mysql both return ` `, so in mysql we can't distinguish default value from database is `NULL` or `Empty String` that will skips `ALTER`

a631807682:
Maybe we can specify `UNIQUE` only field's unique is true and database is false in mysql driver like [postgres](https://github.com/go-gorm/postgres/blob/master/migrator.go#L289) ?

a631807682:
There are 2 problems here
1. `column_default` returned by different db versions is inconsistent
2. `gorm:"unique;default:NULL"` and `gorm:"unique;` are inconsistent in handling

I will create 2 PRs to fix them.

PR #5338, opened 2022-05-13
fix: trx in hooks clone stmt

### LINKED ISSUE #5336: Beginning transaction when statement not empty can cause issues
## GORM Playground Link
https://github.com/go-gorm/playground/pull/477

## Description
Within a Hook I would like to do some deletion within a transaction.
```go
type User struct {
	gorm.Model
}

type Account struct {
	CreatedAt time.Time
	UpdatedAt time.Time
	DeletedAt soft_delete.DeletedAt `gorm:"index"`
	UserID    sql.NullInt64
}

func (user *User) BeforeDelete(tx *gorm.DB) error {
	account := Account{UserID: sql.NullInt64{Int64: int64(user.ID), Valid: true}}
	DB.Find(&account)

	return tx.Transaction(func(tx2 *gorm.DB) error {
		return tx2.Delete(&account).Error
	})
}
```

I would expect this code to work, however it panics.

The cause of the issue seem to be that the transaction create a session with NewDB = false instead of true in finisher_api.go:593

In my example, this cause the code to panic because in the transaction, the statement refer to the table `user` when deleting the `account`.
So it uses the `StatementModifier` of `user` instead of the one of `account`.
As `account` uses `soft_delete.DeletedAt` and `user` use `gorm.DeletedAt`, gorm tries to set a time in `account`'s `DeletedAt` which is a uint and panic.

a631807682:
- [x] Do only one thing
- [x] Tested

close #5336

### DIFF
--- a/tests/transaction_test.go
+++ b/tests/transaction_test.go
@@ -3,9 +3,10 @@ package tests_test
 import (
 	"context"
 	"errors"
+	"testing"
+
 	"gorm.io/gorm"
 	. "gorm.io/gorm/utils/tests"
-	"testing"
 )
 
 func TestTransaction(t *testing.T) {
--- a/tests/transaction_test.go
+++ b/tests/transaction_test.go
@@ -3,10 +3,9 @@ package tests_test
 import (
 	"context"
 	"errors"
-	"testing"
-
 	"gorm.io/gorm"
 	. "gorm.io/gorm/utils/tests"
+	"testing"
 )
 
 func TestTransaction(t *testing.T) {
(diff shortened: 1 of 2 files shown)


tests/transaction_test.go
```diff
@@ -3,10 +3,9 @@ package tests_test
 import (
 	"context"
 	"errors"
-	"testing"
-
 	"gorm.io/gorm"
 	. "gorm.io/gorm/utils/tests"
+	"testing"
```
jinzhu, line 8: Hello @a631807682 

Seems like this code is not formatted with `gofmt`?

a631807682: @jinzhu 
It's formatted by my gofmt, it is formatted like this after I run `gofmt -s -w ./tests`.
I will try to understand why it happens.
```shell
> go version
go version go1.18 darwin/arm64
```

```shell
> gofmt -e -d ./tests
diff -u tests/transaction_test.go.orig tests/transaction_test.go
--- tests/transaction_test.go.orig      2022-05-17 12:50:51.000000000 +0800
+++ tests/transaction_test.go   2022-05-17 12:50:51.000000000 +0800
@@ -3,9 +3,9 @@
 import (
        "context"
        "errors"
-       "testing"
        "gorm.io/gorm"
        . "gorm.io/gorm/utils/tests"
+       "testing"
 )
 
 func TestTransaction(t *testing.T) {
```

a631807682: refer to https://github.com/mvdan/gofumpt/issues/228

PR #5341, opened 2022-05-16
add UniqueIndex attribute to Field strucrt(The check error is not caused by my code)

piyongcai:
- [x] Do only one thing
- [x] Tested

add UniqueIndex attribute to Field strucrt

in some case, 
1. defind a struct field and set uniqueIndex tag ony. and call AutoMigrate
2. in some case, you must call AutoMigrage more than once. (eg, restart your app or change struct), the db return  unique = true, but field struct return unique = false.
3. it will be call AlterColumn. in some time(eg: partition table), it will be failure.

in schema/field.go
1.Add UniqueIndex field to Field's struct。
4.Resolve UniqueIndex value from StructField.

in migrator/migrator.go -> MigrateColumn
1.check unique, Check for the uniqueIndex and unique simultaneously

### User Case Description

```go
type Order struct{
   ID int                     `gorm:"uniqueIndex:uk@order"`
   BizTime time.Time `gorm:"uniqueIndex:uk@order"`  // it's partition field.
}

// db init
// init PartitionTable 
//..... first
if err := db.AutoMigrate(&Order{}); err != nil{
   painc(err)
}

//... db.AutoMigrate Again. 
if err := db.AutoMigrate(&Order{}); err != nil{
   painc(err)  // it will be painc.  becouse alter Partition Field BizTime
}


```

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -440,7 +440,7 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 	}
 
 	// check unique
-	if unique, ok := columnType.Unique(); ok && unique != field.Unique {
+	if unique, ok := columnType.Unique(); ok && !(unique == field.Unique || unique == field.UniqueIndex) {
 		// not primary key
 		if !field.PrimaryKey {
 			alterColumn = true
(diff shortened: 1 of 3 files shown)


jinzhu:
Hi @piyongcai 

Seems CI failed for this PR.

PR #5359, opened 2022-05-21
fix: migrate column default value

a631807682:
- [x] Do only one thing
- [x] Tested

relate to https://github.com/go-gorm/gorm/pull/5325 and https://github.com/go-gorm/mysql/pull/75

mariadb test will pass after https://github.com/go-gorm/mysql/pull/75 merged

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -448,10 +448,20 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 	}
 
 	// check default value
-	if v, ok := columnType.DefaultValue(); ok && v != field.DefaultValue {
-		// not primary key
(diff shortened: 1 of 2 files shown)


sc-ms:
Hey, I saw your fix but in my case I have for
`dv, dvNotNull := columnType.DefaultValue()`
these values:
dv: "NULL"
dvNotNull: true

So it has `alterColumn = true` but in migrator it doesn't change the column.

a631807682:
@sc-ms can you provide an example?

sc-ms:
I found the reason why it didn't work for me.
I noticed it while creating an example.
I had still used gorm.io/driver/mysql in v1.3.3.
But with v1.3.4 it works correctly.

Please excuse me for the confusion.

PR #5367, opened 2022-05-23
fixed FirstOrCreate not handled error when table is not exists

ophum:
- [x] Do only one thing
- [x] Tested

In FirstOrCreate, `result.Error` of `queryTx.Find` is not nil, but not assign to tx.Error.
I fixed `result.Error` assign to `tx.Error`.

https://github.com/ophum/gorm/blob/8edc4825256aa0dcc0f74bd3e92b7fc59ba09241/finisher_api.go#L319

### User Case Description

when using a table that does not exist.

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -354,6 +354,8 @@ func (db *DB) FirstOrCreate(dest interface{}, conds ...interface{}) (tx *DB) {
 		} else {
 			tx.Error = result.Error
 		}
+	} else {
+		tx.Error = result.Error
 	}
 	return tx
 }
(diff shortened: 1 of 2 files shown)


a631807682 (approved):

finisher_api.go
```diff
@@ -354,6 +354,8 @@ func (db *DB) FirstOrCreate(dest interface{}, conds ...interface{}) (tx *DB) {
 		} else {
 			tx.Error = result.Error
```
a631807682, line 355: Seems this part is useless

ophum: Thank you for your review :bow:

changed with this commit.
[delete useless part](https://github.com/go-gorm/gorm/pull/5367/commits/231b98ae7afeaedb5235a4753e47f66dd8c5e04b)

PR #5452, opened 2022-06-23
add autoIncrement implement for #5266

wclssdn:
add autoIncrement implement

close #5266

PR #5453, opened 2022-06-23
Fix default value

wclssdn:
- [ x ] Do only one thing
- [ x ] Non breaking API changes
- [ x ] Tested

Fix a bug: when migrate a model to table, the default string value do not have `"` wear up.


wclssdn:
this pr require #5452 merged.

PR #5455, opened 2022-06-24
feat: use callback to handle transaction

sunfuze:
- [x] Do only one thing
- [x] Tested

Make transaction have before and after hooks, so plugin can hack before
or after transaction.

### User Case Description

We hava a gorm plugin to handle database sharding. We save SaaS tenant info into context, then plugin can get the tenant in before hooks and determine which database shard to select. For now, `Transaction` api is not invoked through callbacks, we have to wrap gorm.ConnPool  as gorm.ConnPoolBeginner to select database.

### DIFF
--- a/callbacks.go
+++ b/callbacks.go
@@ -2,6 +2,7 @@ package gorm
 
 import (
 	"context"
+	"database/sql"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


huacnlee:
Need test

PR #5473, opened 2022-06-30
fix: association many2many duplicate elem

### LINKED ISSUE #5007: Many2many error ON CONFLICT DO UPDATE postgresql
## GORM Playground Link

https://github.com/go-gorm/playground/pull/426

## Many2many error on conflict
I use many2many models
```
type Application struct {
	ID        string `gorm:"primarykey"`
	Name      string
	Type      string
	Resources []Resource `json:"resources,omitempty" gorm:"many2many:application_resources;" faker:"-"`
}

type Resource struct {
	ID       string `gorm:"primarykey"`
	Name     string
	Type     string
	Packages []Package `json:"packages,omitempty" gorm:"many2many:resource_packages;" faker:"-"`
}

type Package struct {
	ID   string `gorm:"primarykey"`
	Name string
	Type string
}
```
In my case there is a single application with 2 resources that **associate to the same package**.
The application update flow with `FullSaveAssociations: true` failed due to
```
ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time (SQLSTATE 21000)
[2.167ms] [rows:0] INSERT INTO "packages" ("id","name","type") VALUES ('1','name','type'),('1','name','type') ON CONFLICT ("id") DO UPDATE SET "name"="excluded"."name","type"="excluded"."type"
```

Instead of inserting only a single `package` GORM tries to insert the same duplicate `package - ('1','name','type')`  in a single INSERT command

I would expect that the save command would not failed and that only single `package` will be inserted.

Maybe there is a way to remove duplicates values **before** creating the INSERT statement for `Associations`

a631807682:
- [x] Do only one thing
- [x] Tested

close #5007

### DIFF
--- a/callbacks/associations.go
+++ b/callbacks/associations.go
@@ -253,6 +253,7 @@ func SaveAfterAssociations(create bool) func(db *gorm.DB) {
 					fieldType = reflect.PtrTo(fieldType)
 				}
 				elems := reflect.MakeSlice(reflect.SliceOf(fieldType), 0, 10)
+				distinctElems := reflect.MakeSlice(reflect.SliceOf(fieldType), 0, 10)
 				joins := reflect.MakeSlice(reflect.SliceOf(reflect.PtrTo(rel.JoinTable.ModelType)), 0, 10)
 				objs := []reflect.Value{}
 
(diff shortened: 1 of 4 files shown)


allidoisace:
@jinzhu would this also solve duplicates for 3 composite keys?

(id, version, time)?

jinzhu:
> @jinzhu would this also solve duplicates for 3 composite keys?
> 
> PRIMARY KEY(id, version, time)?

If you defined primary key tag for those three fields, it should works.

PR #5477, opened 2022-07-01
fix: return id which have type string after created

nohattee:
- [x] Do only one thing
- [x] Tested

Returning the ID which has uuid.UUID or String type after created

### User Case Description

I have a sql script like the below and run on Postgres DB:
```sql
CREATE TABLE "users" (
    "id" UUID NOT NULL DEFAULT gen_random_uuid(),
    "email" TEXT NOT NULL,
    "password" TEXT,
    "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL,
    "deleted_at" TIMESTAMP(3),
    CONSTRAINT "users_pkey" primary KEY ("id")
);
```
In the golang code, I created a struct like this:
```go
type User struct {
	ID        uuid.UUID      `json:"id"`
	Email     string         `json:"email"`
	Password  string         `json:"password"`
	CreatedAt time.Time      `json:"created_at"`
	UpdatedAt time.Time      `json:"updated_at"`
	DeletedAt gorm.DeletedAt `json:"deleted_at"`
}
```

I use this method to create user:
```go
user := &User{
         Email:    newUser.Email,
         Password: newUser.Password,
}
result := db.Select("Email", "Password").Create(user)
fmt.Println(user)
```
The ID was not return after created, so that the user.ID has zero value (not expected).

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -239,6 +239,14 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 				field.HasDefaultValue = true
 				field.AutoIncrement = true
 			}
+		case String:
+			if _, ok := field.TagSettings["PRIMARYKEY"]; !ok {
+				if !field.HasDefaultValue || field.DefaultValueInterface != nil {
+					schema.FieldsWithDefaultDBValue = append(schema.FieldsWithDefaultDBValue, field)
+				}
+
+				field.HasDefaultValue = true
+			}
 		}
 	}
 
--- a/tests/postgres_test.go
+++ b/tests/postgres_test.go
@@ -9,6 +9,56 @@ import (
 	"gorm.io/gorm"
 )
 
+func TestPostgresReturningIDWhichHasStringType(t *testing.T) {
+	if DB.Dialector.Name() != "postgres" {
+		t.Skip()
+	}
+
+	type Yasuo struct {
+		ID        string `gorm:"default:gen_random_uuid()"`
+		Name      string
+		CreatedAt time.Time `gorm:"type:TIMESTAMP WITHOUT TIME ZONE"`
+		UpdatedAt time.Time `gorm:"type:TIMESTAMP WITHOUT TIME ZONE;default:current_timestamp"`
+	}
+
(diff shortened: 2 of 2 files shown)


jinzhu:
Hi @nohattee 

Can you add some tests?

nohattee:
> ```sql
>  gen_random_uuid()
> ```

Hi @jinzhu 
For adding test cases, I need to create a table that has a DEFAULT gen_random_uuid() value like this:
```sql
CREATE TABLE "users" (
    "id" UUID NOT NULL DEFAULT gen_random_uuid(),
    "email" TEXT NOT NULL,
    "password" TEXT,
    "created_at" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
    "updated_at" TIMESTAMP(3) NOT NULL,
    "deleted_at" TIMESTAMP(3),
    CONSTRAINT "users_pkey" primary KEY ("id")
);
```
Do you know how to create a thing like that? Please show me and I will add more test cases

jinzhu:
> 

Hi , you can define a struct with default tag for example:

https://gorm.io/docs/create.html#Default-Values

Also refer `tests/postgres_test.go`

nohattee:
hi @jinzhu , I just added some test cases. Please help me double-check it.

a631807682:
Hi @nohattee 
Can the difference be reflected in the test? Can a test case fail before the change and succeed after the change?
It seems to me that it can be handled here https://github.com/go-gorm/gorm/pull/5477/files#diff-eafe62eae191e88faad4c223c4de9af268c175ee68695872e175b814d7c5baa5L227

nohattee-manabie:
Hi @a631807682 
This case is kind of hard to reproduce. I need to generate a table with the SQL script and the struct like `User Case Description
`. But it seems there is no way to do that in `gorm`

a631807682:
> Hi @a631807682 This case is kind of hard to reproduce. I need to generate a table with the SQL script and the struct like `User Case Description `. But it seems there is no way to do that in `gorm`

Do you mean there is a difference between ddl and model? you can use `Table(..).AutoMigrate()` to create table, and use other `model` to create records.
refer to https://github.com/go-gorm/gorm/blob/master/tests/migrate_test.go#L627

PR #5492, opened 2022-07-06
use callbacks.Begin as a hook to handle begin operation

sunfuze:
make transaction support hooks

- [x] Do only one thing
- [x] Tested

Use callbacks.Begin as hook to handle real transaction operation, make transaction support before/after hooks.

### User Case Description

- Database sharding: shard database by info in context
- Implement distributed transaction

### DIFF
--- a/callbacks.go
+++ b/callbacks.go
@@ -79,19 +79,10 @@ func (cs *callbacks) Transaction() *processor {
 }
 
 func (p *processor) Begin(tx *DB, opt *sql.TxOptions) *DB {
-	var err error
+	tx.InstanceSet("gorm:transaction_options", opt)
 
-	switch beginner := tx.Statement.ConnPool.(type) {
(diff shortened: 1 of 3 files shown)


jinzhu:
notes to myself:

The Transaction hooks API will be behaves like:

```go
tx := db.Scopes(OnTransaction{
       OnCommit: func(tx *gorm.TransactionInterface) {
               // xxx
               tx.Commit()
               // xxx
       },
       OnRollback: func(tx *gorm.TransactionInterface) {
               // xxx
               tx.Rollback()
               // xxx
       },
       OnBegin: func(db *gorm.DBInterface) {
               // xxx
               tx.Begin()
               // xxx
       },
}).Begin()

tx.Commit()
```

EnginnerDog:
I think I just saw a unicorn in this PR.

Issue #5495, opened 2022-07-08
FindInBatches fail when using composite primary key


fanlvG:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/491

## FindInBatches reports "primary key required" due to this bug.
FindInBatches works depends on sorting primaryKey.
when model has more primaryKeys, FindInBatches should use first primaryKeys as sort condition.
reference: https://gorm.io/docs/write_plugins.html

while in schema/schema.go:217. set `PrioritizedPrimaryField` has bug.
```
if schema.PrioritizedPrimaryField == nil && len(schema.PrimaryFields) == 1 {
	schema.PrioritizedPrimaryField = schema.PrimaryFields[0]
}
```

in gorm.io docs. `PrioritizedPrimaryField` described as:

>  // Prioritized primary key field: field with DB name `id` or the first defined primary key

in "Write Plugins" chapter. 

please check this out.

correct code should be like:
```
if schema.PrioritizedPrimaryField == nil && len(schema.PrimaryFields) >= 1 {
	schema.PrioritizedPrimaryField = schema.PrimaryFields[0]
}
```

a631807682:
In the composite primary key case, the primary key is not only the sorting keyword, but also the starting condition of the next query. We cannot use prioritized primary key to implement batch query in this case.
Maybe we need to use subqueries for this case.

fanlvG:
oh I see. If in previous query we couldn't select all records with same primaryKey1, In next query we can't use less condition to select the rest records with primaryKey1 value. So maybe FindInBatches() use ORDER BY OFFSET LIMIT will be better? 
That can handle most batch query situation. I think.

a631807682:
@fanlvG If using offset directly will have performance issues with large data volumes, I will try to fix it later.

tuhaolam:
It seems that the bug has not been fixed, I reproduces this bug on latest version v1.25.8.

kawhicurry:
same issue on v1.25.12

PR #5497, opened 2022-07-08
fix(FindInBatches): support composite primarykey

a631807682:
- [x] Do only one thing
- [x] Tested

### User Case Description

close https://github.com/go-gorm/gorm/issues/5495

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -173,7 +173,7 @@ func (db *DB) Find(dest interface{}, conds ...interface{}) (tx *DB) {
 // FindInBatches find records in batches
 func (db *DB) FindInBatches(dest interface{}, batchSize int, fc func(tx *DB, batch int) error) *DB {
 	var (
-		tx = db.Order(clause.OrderByColumn{
+		tx = db.Session(&Session{}).Order(clause.OrderByColumn{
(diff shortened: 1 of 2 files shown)


finisher_api.go
```diff
@@ -173,7 +173,7 @@ func (db *DB) Find(dest interface{}, conds ...interface{}) (tx *DB) {
 // FindInBatches find records in batches
 func (db *DB) FindInBatches(dest interface{}, batchSize int, fc func(tx *DB, batch int) error) *DB {
 	var (
-		tx = db.Order(clause.OrderByColumn{
+		tx = db.Session(&Session{}).Order(clause.OrderByColumn{
```
a631807682, line 176: test failed caused by here, modify it can be successful
```go
tx           = db.Session(&Session{})
```

a631807682:
I thought records were stored in primary key order, but using the first primary key ordering in the postgrel composite primary key makes the order inconsistent, which makes me not sure if there is a performance problem with the way sub-query, I need to look at the postgrel related information to decide whether to complete this PR

finisher_api.go
a631807682: https://www.postgresql.org/docs/current/queries-limit.html
 > it is an inherent consequence of the fact that SQL does not promise to deliver the results of a query in any particular order unless ORDER BY is used to constrain the order.

PR #5499, opened 2022-07-09
test(MigrateColumn): mock alter column to improve field compare

a631807682:
- [x] Do only one thing
- [x] Tested

A way for improve compare field in different drivers.
We can find the following problems
postgres: 
```log
smallint -> int2
integer -> int4 
bigint -> int8
decimal ->  numeric
```
mysql:
```log
bool -> tinyint
```
sqlserver:
```go
value, ok = columnType.DefaultValue()
// ok alway true, means default value not null
```


a631807682:
@jinzhu Yes, this PR should not be merged unless the issues that occurred in all drivers are fixed. Now its purpose is to find the problem in driver MigrateColum.

jinzhu (approved):

Issue #3553, opened 2020-09-29
Strange memory consumption, that is slowly increasing


petrjahoda:
## GORM Playground Link
https://github.com/go-gorm/playground/pull/161

## Description

I am using Gorm a lot (v1, now v2), but I spot something, that attracted my attention. My services are very slowly, but steadily, consuming more and more memory. I dig deeper and removed code one thing after another. At the end I was left just with one function, that repeatedly downloads data from database.
So after that I minimize that function to not download anything, just open and close the connection.
Then I made a simple program, that only opens the connection and closes.
See attached gif that show this behavior and two screenshots that show the memory increase after a while.
The memory is increasing not constantly, looks like at logarithmic scale, but is increasing still.

Maybe I am doing something wrong, but I did not found anything wrong with the code.
And yes, I need to open and closes connections as I need. I cannot opens - in my real program - one connection at the beginning and then reuse it everywhere.

![animated](https://user-images.githubusercontent.com/45359845/94548579-3a616500-0251-11eb-814d-b6250ddbac92.gif)

<img width="1530" alt="first" src="https://user-images.githubusercontent.com/45359845/94548628-4b11db00-0251-11eb-9719-86c9d107ae2f.png">

<img width="1537" alt="second" src="https://user-images.githubusercontent.com/45359845/94548639-4f3df880-0251-11eb-8881-c293710f85f8.png">

jinzhu:
Hello,

I have tested it `pprof` and used `GODEBUG=gctrace=1` to test the program, seems no leak happened, so this should not a GORM issue.

Thank you.

jinzhu:
also tested with `sqlDB, err := sql.Open("mysql", dbDSN)`, got same issue, seems like golang doesn't release the memory to OS or the OS just haven't take the memory from the go application yet.

petrjahoda:
meanwhile I found out that, when manually calling garbage collector with runtime.GC() before the end of the loop, the memory DOES NOT increases that fast.
It increases still... looks like... but at a very very very low speed.

I tested it for about 5 hours and it looks like it hit the limit above which it does not increase. Maybe I let it run for a weekend to confirm this.
But from this observation, the "problem" may be with garbage collector, maybe not collecting properly those database connections, or maybe collecting them at somewhat slow or strange speed (collecting them at slower rate that new are made - but that does not make sense to me).

Also I made a different test: as I am using postgres, I tested this postgres driver: https://github.com/jackc/pgx.
Similar results.
[...]

PR #5508, opened 2022-07-13
(WIP)fix: use preparestmt in trasaction will use new conn

li-jin-gou:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested

- fix: use preparestmt in trasaction will use new conn

### User Case Description
```
func TestGORM(t *testing.T) {
	user := User{Name: "jinzhu"}
	if err := DB.Transaction(func(tx *gorm.DB) error {
		tx.Session(&gorm.Session{PrepareStmt: true}).Create(&user)
		return errors.New("test")
	}); err != nil {
		t.Error(err)
	}
	var result User
	if err := DB.First(&result, user.ID).Error; err != nil {
		t.Errorf("Failed, got error: %v", err)
	}
	t.Log(result.ID)
}
```
it will create user

### DIFF
--- a/gorm.go
+++ b/gorm.go
@@ -249,7 +249,7 @@ func (db *DB) Session(config *Session) *DB {
 		if v, ok := db.cacheStore.Load(preparedStmtDBKey); ok {
 			preparedStmt := v.(*PreparedStmtDB)
 			tx.Statement.ConnPool = &PreparedStmtDB{
-				ConnPool: db.Config.ConnPool,
+				ConnPool: db.Statement.ConnPool,
 				Mux:      preparedStmt.Mux,
 				Stmts:    preparedStmt.Stmts,
 			}


lollipopjin:
LGTM

pablodz:
Maybe this causes memory leak due new connections?

https://github.com/go-gorm/gorm/issues/3553

li-jin-gou:
> Maybe this causes memory leak due new connections?
> 
> #3553

@pablodz does not lead to memory leaks, but does not guarantee that transactions.

gorm.go
```diff
@@ -249,7 +249,7 @@ func (db *DB) Session(config *Session) *DB {
 		if v, ok := db.cacheStore.Load(preparedStmtDBKey); ok {
 			preparedStmt := v.(*PreparedStmtDB)
 			tx.Statement.ConnPool = &PreparedStmtDB{
-				ConnPool: db.Config.ConnPool,
+				ConnPool: db.Statement.ConnPool,
```
jinzhu, line 252: if in transaction, should change the tx.Statement.ConnPool to

```go
PreparedStmtTX{
	Tx: db.Statement.ConnPool,
	PreparedStmtDB: PreparedStmtDB{
		ConnPool: db.Config.ConnPool,
		Mux:      preparedStmt.Mux,
		Stmts:    preparedStmt.Stmts,
	}
}
```

PR #5554, opened 2022-07-29
Update Delete Godoc to describe soft delete behaviour

bruc3mackenzi3:
- [x] Do only one thing
- [x] Tested

gorm's Delete() method performs either a soft or hard delete, depending on whether the passed data model includes a `deletedAt` field.  This is detailed [here in the docs](https://gorm.io/docs/delete.html#Soft-Delete) and this PR adds this to the Godoc comment.

I've also cleaned up the grammar to make the language of the comment consistent.

### User Case Description

As a new user to gorm I found myself referencing the external docs at gorm.io to learn the behaviour of this method.  Adding key details to inline documentation improves the user experience.

If there is an appetite for contribution to documentation I would like to submit more PRs like this across the codebase.

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -388,7 +388,9 @@ func (db *DB) UpdateColumns(values interface{}) (tx *DB) {
 	return tx.callbacks.Update().Execute(tx)
 }
 
-// Delete delete value match given conditions, if the value has primary key, then will including the primary key as condition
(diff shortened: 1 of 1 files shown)


bruc3mackenzi3:
@jinzhu is it possible to get a review on this PR?  I'm interested in making further contributions to GORM's documentation.  If there's a desire for these sorts of improvements I will follow this PR up with more like it.  Cheers!

Issue #4872, opened 2021-11-23
Allow for inner or right join in Join preloading


williamfinn:
## Describe the feature

Would like to be able to specify a Join type / direction when using Joins Preloading. 

## Motivation

Currently the documentation says that an "inner" join is carried out but in fact a left join is carried out. I think that it would be an excellent feature to have other join types allowed for join preloading for example: right joining between multiple entities will only return a row if all of the right joins are satisfied.

## Code example
```
type Card struct {
	Id           int          `json:"id,omitempty" gorm:"column:id"`
	Amount       float64      `json:"amount,omitempty" gorm:"column:amount"`
	Currency     string       `json:"currency,omitempty" gorm:"column:currency"`
	GiftCard     Giftcard     `json:"giftCard,omitempty" gorm:"foreignKey:CardId"`
	CardConsumer CardConsumer `json:"consumer,omitempty" gorm:"foreignKey:CardId"`
}

// note that the entity name is specified "GiftCard" and "CardConsumer" to facilitate join preloading
r.db.Joins("right join GiftCard").Joins("right join CardConsumer").Find(&card)
```

mnussbaum:
This is indeed a discrepancy between the docs and the code, and the ability to do inner joins would be very helpful

shencw:
@jinzhu  大佬 这个问题还挺有帮助的  想要使用关联关系的时候 right join 怎么操作呀

williamfinn:
On second thought, doesn't make much sense for right join preloading (as you could have no parent object to preload on). If you want to simulate inner or right joins currently I suggest just doing this in where conditions (CardConsumer.id != null for example).

tbistr:
Why does gorm replace "JOIN" with "LEFT JOIN"?
Just "JOIN" means "INNER JOIN" in MySQL and PostgressSQL at least.

I had the same problem.

## minimam sample
```go
type A struct {
	ID uint
	B  B
}

type B struct {
	ID   uint
	AID  uint
	Data string
}

db.Create(&A{B: B{Data: "data1"},})
db.Create(&A{B: B{Data: "data2"},})
db.Create(&A{B: B{Data: "data3"},})
As := []A{}
db.Debug().Joins("B", db.Where(&B{Data: "data3"})).Find(&As)
fmt.Println(As)
```

This code execute SQL like bellow.

```SQL
[...]

PR #5583, opened 2022-08-06
feat: support inner join

a631807682:
- [x] Do only one thing
- [x] Tested

related to https://github.com/go-gorm/gorm/issues/4872#issuecomment-1207201423

I also don't think right join is necessary, so only inner join is supported.


williamfinn:
@a631807682 I think a good alternative of going forward with this is to incorporate it somehow into the current Join function. this is where I wish we had kwargs in go 😄. Thoughts ?

abdoelrofiq (approved):

daheige (approved):

ammyblabla (approved):

jinzhu (approved):

PR #5603, opened 2022-08-13
Replace `ioutil.Discard` with `io.Discard`

zaneli:
- [x] Do only one thing
- [ ] Tested

`"io/ioutil"` has been deprecated since Go 1.16.
https://tip.golang.org/doc/go1.16#ioutil

### User Case Description

N/A

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module gorm.io/gorm
 
-go 1.14
+go 1.16
 
 require (
 	github.com/jinzhu/inflection v1.0.0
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module gorm.io/gorm
 
-go 1.14
+go 1.16
```
zaneli, line 3: Tentatively updated according to min version on CI.
https://github.com/zaneli/gorm/blob/4768b87f15cdfff699fb449cedb276d185ce8763/.github/workflows/tests.yml#L19

huacnlee:
Keep one PR just do one thing.

PR #5613, opened 2022-08-15
autoMigrate支持动态生成表

Caisin:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


jinzhu:
Sorry, but not going to support dynamic table name by design, maybe you can use `Scopes` for dynamic table support.

PR #5634, opened 2022-08-23
support scan assign slice cap

demoManito:
- [x] Do only one thing
- [x] Tested

support scan assign slice cap，instead of the default cap of 20

### User Case Description
for example:
```go
users := make([]User, 0, 2)
db.Where(...).Find(&users).Error
```
the initial make cap is 2, instead of 20


demoManito:
image
this error should not be caused by my modification

demoManito:
> image
> 
> this error should not be caused by my modification
> repaired PR #5620 ?

CI success

PR #5667, opened 2022-09-01
Migrator allow to change from NOT NULL to DEFAULT NULL

### LINKED ISSUE #5643: Automigration NOT NULL to DEFAULT NULL not working
## GORM Playground Link

https://github.com/go-gorm/playground/pull/512

## Description

It's not possible to change not nullable field to nullable.

```
type Test struct {
	gorm.Model
	SomeStr string `gorm:"type:VARCHAR(15);NOT NULL"`
}
```

After `AutoMigrate` is run, `some_str` field on DB is not nullable, then changing struct to this:
```
type Test struct {
	gorm.Model
	SomeStr string `gorm:"type:VARCHAR(15);default:NULL"`
}
```

and running `AutoMigrate` again, field remains not nullable.
I believe problem is here: https://github.com/go-gorm/gorm/blob/master/migrator/migrator.go#L439
both `nullable` (actual state on DB) and `field.NotNull` (new value) are `false`, meaning it's only possible to change from nullable to not nullable. I don't know if it's done on purpose.

pjarmalavicius:
- [x] Do only one thing
- [x] Tested

Fixes #5643

### User Case Description

Described in #5643

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -437,8 +437,8 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 
 	// check nullable
 	if nullable, ok := columnType.Nullable(); ok && nullable == field.NotNull {
-		// not primary key & database is nullable
-		if !field.PrimaryKey && nullable {
+		// not primary key
+		if !field.PrimaryKey {
 			alterColumn = true
 		}
 	}


a631807682:
Can you add some tests for it? We need to confirm that subsequent changes do not break this behavior.

jinzhu:
tests broken

PR #5682, opened 2022-09-08
Optimize: utils.Contains applicable to more types

demoManito:
- [x]  Do only one thing
- [x] Tested

I think the utils methods under utils should face more types than [] string

### User Case Description
refer to test cases

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -69,9 +69,14 @@ func ToStringKey(values ...interface{}) string {
 	return strings.Join(results, "_")
 }
 
-func Contains(elems []string, elem string) bool {
-	for _, e := range elems {
-		if elem == e {
+func Contains(elems interface{}, elem interface{}) bool {
+	reflectValue := reflect.ValueOf(elems)
+	if reflectValue.Kind() != reflect.Slice && reflectValue.Kind() != reflect.Array {
+		return false
+	}
+
+	for i := 0; i < reflectValue.Len(); i++ {
+		if reflectValue.Index(i).Interface() == elem {
 			return true
 		}
 	}
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -29,3 +29,23 @@ func TestToStringKey(t *testing.T) {
 		}
 	}
 }
+
+func TestContains(t *testing.T) {
+	tests := []struct {
+		elems  interface{}
+		elem   interface{}
+		target bool
+	}{
+		{elems: []string{"INSERT", "VALUES", "ON CONFLICT", "RETURNING"}, elem: "RETURNING", target: true},
+		{elems: []string{"INSERT", "VALUES", "ON CONFLICT"}, elem: "RETURNING", target: false},
+		{elems: []int{1, 2, 3}, elem: 1, target: true},
+		{elems: []interface{}{1, 2.0, "3"}, elem: 2.0, target: true},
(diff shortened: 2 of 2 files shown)


a631807682:
I don't think it makes sense, and reflect will have performance issues. If this method name is causing you to misunderstand, I think it can be changed to `ContainsString`

demoManito:
> I don't think it makes sense, and reflect will have performance issues. If this method name is causing you to misunderstand, I think it can be changed to `ContainsString`

I don't think so. The methods of the tool class should be more widely used. If they are not the methods of the tool class, there is no problem. Moreover, I have benchmarked the places I use, and the performance difference is not obvious

a631807682:
```go
var ss = []string{"1", "2", "3", "4", "5"}
var s = "5"

func BenchmarkContainsString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		ContainsString(ss, s)
	}
}

func BenchmarkContainsReflect(b *testing.B) {
	for i := 0; i < b.N; i++ {
		ContainsReflect(ss, s)
	}
}

func ContainsString(elems []string, elem string) bool {
	for _, e := range elems {
		if elem == e {
			return true
		}
	}
	return false
}

func ContainsReflect(elems interface{}, elem interface{}) bool {
	reflectValue := reflect.ValueOf(elems)
	if reflectValue.Kind() != reflect.Slice && reflectValue.Kind() != reflect.Array {
		return false
	}

	for i := 0; i < reflectValue.Len(); i++ {
		if reflectValue.Index(i).Interface() == elem {
			return true
		}
	}
	return false
}

```
go version go1.19 darwin/arm64
```
BenchmarkContainsString-8       91242894                12.43 ns/op
BenchmarkContainsReflect-8       6928802               170.1 ns/op
```

jinzhu:
Not going to merge this PR, as seems this is not a necessary change for GORM.

demoManito:
> ```go
> var ss = []string{"1", "2", "3", "4", "5"}
> var s = "5"
> 
> func BenchmarkContainsString(b *testing.B) {
> 	for i := 0; i < b.N; i++ {
> 		ContainsString(ss, s)
> 	}
> }
> 
> func BenchmarkContainsReflect(b *testing.B) {
> 	for i := 0; i < b.N; i++ {
> 		ContainsReflect(ss, s)
> 	}
> }
> 
> func ContainsString(elems []string, elem string) bool {
> 	for _, e := range elems {
> 		if elem == e {
> 			return true
> 		}
> 	}
> 	return false
> }
> 
> func ContainsReflect(elems interface{}, elem interface{}) bool {
> 	reflectValue := reflect.ValueOf(elems)
> 	if reflectValue.Kind() != reflect.Slice && reflectValue.Kind() != reflect.Array {
> 		return false
> 	}
> 
> 	for i := 0; i < reflectValue.Len(); i++ {
> 		if reflectValue.Index(i).Interface() == elem {
> 			return true
> 		}
> 	}
> 	return false
> }
> ```
> 
> go version go1.19 darwin/arm64
> 
> ```
> BenchmarkContainsString-8       91242894                12.43 ns/op
> BenchmarkContainsReflect-8       6928802               170.1 ns/op
> ```

image
for real implementation, the changes here will not affect performance

demoManito:
> Member

OK, thank you for your reply

PR #5699, opened 2022-09-16
fix: update omit

qqxhb:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


upliu:
can someone explain why update omit here ?

PR #5716, opened 2022-09-26
Modify gorm.io ref

brucewangviki:
Update gorm.io reference


brucewangviki:
My bad, opened a wrong PR

PR #5718, opened 2022-09-26
WIP Support clause tostring

demoManito:
- [x] Do only one thing
- [ ] Tested

want: `select * from user where start_time > 11 AND ((age > 10 AND age < 20) OR (age > 20 AND age < 21))`
At present, it is impossible to encapsulate correct SQL statements through ORM. I get the query SQL through expr, which has solved this problem


jinzhu:
Hi @demoManito 

Thank you for your PR, can you let me know the reason the expose this API? which doesn't make much sense for me.

PR #5719, opened 2022-09-26
style fmt import order

demoManito:
- [x] Do only one thing
- [x] Tested


jinzhu:
Hi @demoManito 

Thank you for your PR, but I don't think this is necessary or even a better change.

PR #5721, opened 2022-09-26
fix: primaryFields are overwritten

demoManito:
- [x] Do only one thing
- [x] Tested

### DIFF
--- a/schema/relationship.go
+++ b/schema/relationship.go
@@ -403,33 +403,30 @@ func (schema *Schema) guessRelation(relation *Relationship, field *Field, cgl gu
 	case guessBelongs:
 		primarySchema, foreignSchema = relation.FieldSchema, schema
 	case guessEmbeddedBelongs:
-		if field.OwnerSchema != nil {
-			primarySchema, foreignSchema = relation.FieldSchema, field.OwnerSchema
-		} else {
+		if field.OwnerSchema == nil {
 			reguessOrErr()
 			return
 		}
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @demoManito 

can you add some tests that might reproduce the panic?

demoManito:
> Hi @demoManito
> 
> can you add some tests that might reproduce the panic?

Hello @jinzhu 

The else block `primaryFields` covers the outer `primaryFields`, which is always empty, and may `index out of` panic:
https://github.com/go-gorm/gorm/blob/0b7113b618584edd76d74e7a73eecc2a28a4d17a/schema/relationship.go#L432

Other modifications just expand `if else`

schema/relationship.go
```diff
@@ -403,33 +403,30 @@ func (schema *Schema) guessRelation(relation *Relationship, field *Field, cgl gu
...
-			} else {
+			f := foreignSchema.LookUpField(foreignKey)
+			if f == nil {
 				reguessOrErr()
 				return
 			}
+			foreignFields = append(foreignFields, f)
 		}
 	} else {
-		var primaryFields []*Field
```
demoManito, line 432: This repeat and covered with the outer layer

demoManito: See line 373

demoManito: This causes 469 lines to start, and the primaryKeys are always empty

Issue #5735, opened 2022-10-03
A limit of zero returns the wrong results


robhafner:
GORM Playground Link

https://github.com/go-gorm/playground/pull/522

## Description

Specifying a limit of 0 results in all rows being returned instead of none due to https://github.com/go-gorm/gorm/blob/master/clause/limit.go#L18. 

For example:
gdb.Table(<table>).Limit(0).Find(&testDataRecords)

A pull request to fix this issue will be coming shortly.

robhafner:
Here is the link to the pull request:

https://github.com/go-gorm/gorm/pull/5736

robhafner:
Adding playground pull request which illustrates the issue. 

https://github.com/go-gorm/playground/pull/522

a631807682:
What is the purpose of this usage? If the user doesn't need the data, it shouldn't execute the sql.

robhafner:
Our Rest API supports returning additional pieces of information such as the number of resources in the collection. Clients retrieve this information by specifying a limit of zero when making a REST call to the service.  We accomplish this by executing the query with a limit of zero followed by another query which retrieves the count. 

For us to fix this without a change in gorm's behavior would require touching almost every service we deliver. Which is in the hundreds. Given this I'm hoping you might consider supporting the original behavior given it is valid to send a query to the database with a limit of zero. Also, as far as I can tell the change in behavior was not mentioned in the gorm v2 release notes and so it is unclear if the behavior change was intended or was just an oversight.

robhafner:
Thanks for merging these PRs in! Can you share when the next release of these projects will be released so I can plan accordingly?

robhafner:
I've upgraded our gorm dependencies to the following versions expecting this issue to be resolved now. 

gorm.io/datatypes v1.0.7
gorm.io/driver/postgres v1.4.4
gorm.io/driver/sqlite v1.4.2
gorm.io/gorm v1.24.0

However, our unit tests which run against sqlite are still showing that the limit clause is not being added to the generated sql statement for a value of zero. Stepping through the debugger it appears the following line is missing the '=" sign in the condition and should look like this instead: "if lmt >= 0 || limit.Offset > 0 {"

https://github.com/go-gorm/sqlite/blob/1007940dc92e18c71bce8712a9931f050bef9c23/sqlite.go#L104


I believe the following change might of meant to include the ">"  when the logic was refactored. 
[...]

PR #5736, opened 2022-10-03
fix: limit=0 results (#5735)

robhafner:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested

### DIFF
--- a/clause/limit.go
+++ b/clause/limit.go
@@ -4,7 +4,7 @@ import "strconv"
 
 // Limit limit clause
 type Limit struct {
-	Limit  int
+	Limit  *int
 	Offset int
 }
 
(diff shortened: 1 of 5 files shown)


robhafner:
Here is the link for the pull request to fix the sqlite module.

https://github.com/go-gorm/sqlite/pull/114

robhafner:
Here is the link for the pull request to fix the sqlserver module.

https://github.com/go-gorm/sqlserver/pull/70

robhafner:
Please let me know if there is anything else that you need from me to move this pull request forward as it is preventing some of our services from upgrading to gorm 2. 

Also, given that these fixes reside in different modules it's unclear to move as to how to make the build pass.

System-Glitch:
This PR contains breaking changes.

clause/limit.go
```diff
@@ -4,7 +4,7 @@ import "strconv"
 
 // Limit limit clause
 type Limit struct {
-	Limit  int
```
System-Glitch, line 7: This is a breaking change.

frankli0324: then why is this released? I don't see the point where this change is necessary for the fix...

clause/limit_test.go
```diff
@@ -8,18 +8,26 @@ import (
 )
 
 func TestLimit(t *testing.T) {
+	limit0 := 0
```
frankli0324, line 11: test cases should not be changed unless for a reason that's good enough, this is not good enough.

li-jin-gou: +1

PR #5745, opened 2022-10-08
Rm %v string at non Errorf log

RemcoAn:
- [X] Do only one thing
- [X] Tested

Removes the %v string at a non errof log.


jinzhu:
The `%v` is required to display the error message correctly.

PR #5762, opened 2022-10-14
add test: shouldn't execute ALTER COLUMN ... TYPE smallint

viatoriche:
- [x] Do only one thing
- [x] Tested

Add new test which should be failed for now because gorm still executes `ALTER TABLE "migrate_ints" ALTER COLUMN "int8" TYPE smallint USING "int8"::smallint` during `AutoMigrate` every time when such field already has such type (smallint).
It was tested on the newest version of gorm/postgres.

### User Case Description

It is not good idea to execute such DDL on the huge databases. Let's fix it. I also will try to find a root cause, but I am not sure that I can do it quickly...

### DIFF
--- a/tests/migrate_test.go
+++ b/tests/migrate_test.go
@@ -1,6 +1,7 @@
 package tests_test
 
 import (
+	"context"
 	"fmt"
 	"math/rand"
 	"reflect"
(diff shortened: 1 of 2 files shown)


viatoriche:
[Test example](https://github.com/go-gorm/gorm/actions/runs/3248701280/jobs/5330229748#step:6:143)

viatoriche:
There is a workaround:
USE `int16` instead of `int8`

jinzhu:
Hi @viatoriche 

Thank you for your report, fixed the issue in the latest commit.

PR #5771, opened 2022-10-14
fix: block and arguments is not equal

knkp1209:
image
image

clickhouse: when table field is Array(string)


jinzhu:
Hi @knkp1209 

Can you write a test case for the clickhouse driver?

jinzhu:
Closing this PR for now, please open a PR with test cases for clickhouse to reproduce the issue.

Thank you.

PR #5777, opened 2022-10-17
Ability to use generic functions

LeandroFranciscato:
- [x] Do only one thing
- [x] Tested

It creates the ability for the user to use a generic function in order to retrieve date fields without time.

### User Case Description

We have a report application that runs complexes queries, this application has the ability to connect into different databases.
When trying to use some functions, for the time being, we have to write a proprietary function to handle date formats.
This feature would solve this specific problem, but it can grow, as more specific functions became necessary.

### DIFF
--- a/clause/function.go
+++ b/clause/function.go
@@ -1,0 +1,42 @@
+package clause
+
+import (
+	"fmt"
+	"log"
+)
+
+type Database string
+
+var (
+	Postgres Database = "postgres"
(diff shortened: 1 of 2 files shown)


jinzhu:
Hi @LeandroFranciscato 

I think this PR should not be merged into GORM, maybe it should exists as a util package?

Thank you.

LeandroFranciscato:
I see, thank you for answering @jinzhu.

PR #5778, opened 2022-10-18
test: invalid cache plan with prepare stmt

a631807682:
- [x] Do only one thing
- [x] Tested

related to https://github.com/go-gorm/postgres/pull/137

### DIFF
--- a/tests/migrate_test.go
+++ b/tests/migrate_test.go
@@ -850,12 +850,47 @@ func findColumnType(dest interface{}, columnName string) (
 	return
 }
 
-func TestInvalidCachedPlan(t *testing.T) {
+func TestInvalidCachedPlanSimpleProtocol(t *testing.T) {
 	if DB.Dialector.Name() != "postgres" {
 		return
 	}
 
 	db, err := gorm.Open(postgres.Open(postgresDSN), &gorm.Config{})
+	if err != nil {
+		t.Errorf("Open err:%v", err)
+	}
+
+	type Object1 struct{}
+	type Object2 struct {
+		Field1 string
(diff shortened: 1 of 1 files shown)


jinzhu:
Hello @a631807682 

I think we might need to provide a `Reset` func for `*PreparedStmtDB`, after migrated columns we need to reset prepared statements.

a631807682:
@jinzhu  Reset is unnecessary, it will create prepared stmt for the same query. 
https://github.com/go-gorm/postgres/pull/137/files#diff-a5ca21254ced8196f81a4c4e926af32db03cb0ea78a1f853835300f6dda3f0c6R552
However, we can provide a method in PreparedStmtDB to remove a prepared stmt or support skipping prepare when the current query is executed

jinzhu:
> @jinzhu Reset is unnecessary, it will create prepared stmt for the same query. https://github.com/go-gorm/postgres/pull/137/files#diff-a5ca21254ced8196f81a4c4e926af32db03cb0ea78a1f853835300f6dda3f0c6R552 However, we can provide a method in PreparedStmtDB to remove a prepared stmt or support skipping prepare when the current query is executed

The `Reset` func from my opinion can just remove all cached statements, replace `.Stmts` with a empty map.

Only remove the `select limit 1` SQL seems can't completely avoid the issue.

a631807682:
@jinzhu  I created a pr for it https://github.com/go-gorm/gorm/pull/5782
The current failure is because there is only one place in postgres driver that queries the actual table data, to completely avoid the issue, we need to perform `Reset` after the table changed.

PR #5782, opened 2022-10-19
feat(PreparedStmtDB): support reset

a631807682:
- [x] Do only one thing
- [x] Tested


prepare_stmt.go
```diff
@@ -44,6 +44,12 @@ func (db *PreparedStmtDB) Close() {
 	}
 }
 
+func (db *PreparedStmtDB) Reset() {
+	db.Close()
```
jinzhu, line 48: Maybe we need to range `db.Stmts` and close all prepared statement, `Close` will only close prepared statements in this session.

PR #5790, opened 2022-10-20
fix(Joins): args with select and omit

### LINKED ISSUE #5781: Joins preloading ignores select and omit clauses
## GORM Playground Link

https://github.com/go-gorm/playground/pull/530

## Description

Despite what the docs says, it is not possible to chose the selected columns when using join preloading.

Docs quote:
```
db.Joins("Account", DB.Select("id").Where("user_id = users.id AND name = ?", "someName").Model(&Account{}))
```

I would expect only the columns mentionned in `Select()` to be selected, but all the columns of the relation are selected instead. (Same thing with `Omit()`, which has no effect at all).

This is because `Joins()` adds all columns without checking the Omit or Select statement in query.go [here](https://github.com/go-gorm/gorm/blob/3f20a543fad5f57016ef7a6c342536b0fcce6016/callbacks/query.go#L120).

a631807682:
- [x] Do only one thing
- [x] Tested

close #5781

### DIFF
--- a/callbacks/query.go
+++ b/callbacks/query.go
@@ -117,12 +117,20 @@ func BuildQuerySQL(db *gorm.DB) {
 				} else if relation, ok := db.Statement.Schema.Relationships.Relations[join.Name]; ok {
 					tableAliasName := relation.Name
 
+					columnStmt := gorm.Statement{
+						Table: tableAliasName, DB: db, Schema: relation.FieldSchema,
+						Selects: join.Selects, Omits: join.Omits,
+					}
+
+					selectColumns, restricted := columnStmt.SelectAndOmitColumns(false, false)
 					for _, s := range relation.FieldSchema.DBNames {
-						clauseSelect.Columns = append(clauseSelect.Columns, clause.Column{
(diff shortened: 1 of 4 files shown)


System-Glitch (approved):

gaoxunyang (commented):
not completely work
when you select more than one segment，like select("id,name")
it doesn't work

gaoxunyang:
> > not completely work when you select more than one segment，like select("id,name") it doesn't work
> 
> `.Select("ID", "name")`

ohh, thank you, not same as directly select
i think this way is more convenient...
image

gaoxunyang:
> > > > not completely work when you select more than one segment，like select("id,name") it doesn't work
> > > 
> > > 
> > > `.Select("ID", "name")`
> > 
> > 
> > ohh, thank you, not same as directly select i think this way is more convenient... image
> 
> We already have a lot of support for select, and supporting it won't bring much benefit. https://github.com/go-gorm/gorm/blob/master/tests/query_test.go#L764

ok, thanks for your answer.

PR #5810, opened 2022-10-25
feat(Scopes): make Schema and Table accessible from scopes

System-Glitch:
- [x] Do only one thing
- [x] Tested

This PR executes scopes after parsing the model so `Statement.Schema` and `Statement.Table` are accessible from them. 

### User Case Description

This would be useful in order to implement scopes that are re-usable across multiple models.

An example would be a scope that sorts by ID. The following would not work if there is a join or an ambiguity with the table name:
```go
func sortByID(db *gorm.DB) *gorm.DB) {
    return db.Sort("id")
}
```

Thanks to the change suggested in this PR, we would be able to do the following:
```go
func sortByID(db *gorm.DB) *gorm.DB) {
    return db.Sort(fmt.Sprintf("%s.id", db.Statement.Table))
}
```

This would also open a lot of possibilities for automatic query generation based on a schema.

### DIFF
--- a/callbacks.go
+++ b/callbacks.go
@@ -73,15 +73,6 @@ func (cs *callbacks) Raw() *processor {
 }
 
 func (p *processor) Execute(db *DB) *DB {
-	// call scopes
-	for len(db.Statement.scopes) > 0 {
-		scopes := db.Statement.scopes
-		db.Statement.scopes = nil
-		for _, scope := range scopes {
-			db = scope(db)
-		}
-	}
-
 	var (
 		curTime           = time.Now()
 		stmt              = db.Statement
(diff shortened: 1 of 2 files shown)


jinzhu:
Hi @System-Glitch 

Thank you for your PR, but it will remove the possibility to set `Model` in the `Scopes`, which I think it would be more common.

For your case, maybe you could refer this to implement it. https://github.com/go-gorm/gorm/blob/9d82aa56734999bb28e0c4d60fba69ae7cde66d5/statement.go#L402

Thank you.

System-Glitch:
Hello @jinzhu ! I made a change so we can still set the model from scopes. Can you please take another look at it?

PR #5836, opened 2022-11-04
fix logger path bug

kvii:
- [x] Do only one thing
- [x] Tested

To fix logger path bug.

### User Case Description

before:

``` shell
➜  app go run .
2022/11/04 21:10:29 /Users/kvii/go/pkg/mod/gorm.io/gorm@v1.24.1/callbacks.go:134
[0.004ms] [rows:0] SELECT 1
```

after:

``` shell
➜  app go run .
2022/11/04 20:58:48 /Users/kvii/workspace/qs/gormt/app/main.go:14
[0.005ms] [rows:0] SELECT 1
```

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -23,11 +23,11 @@ func sourceDir(file string) string {
 	dir := filepath.Dir(file)
 	dir = filepath.Dir(dir)
 
-	parent := filepath.Dir(dir)
-	if filepath.Base(parent) == "gorm.io" {
-		return parent
+	s := filepath.Dir(dir)
+	if filepath.Base(s) != "gorm.io" {
+		s = dir
 	}
-	return dir
+	return s + string(filepath.Separator)
 }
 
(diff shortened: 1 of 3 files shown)


utils/utils.go
```diff
@@ -16,7 +16,18 @@ var gormSourceDir string
...
+
+func sourceDir(file string) string {
+	dir := filepath.Dir(file)
+	dir = filepath.Dir(dir)
+
+	parent := filepath.Dir(dir)
+	if filepath.Base(parent) == "gorm.io" {
+		return parent
+	}
+	return dir
```
jinzhu, line 30: the returned source dir is better to include`filepath.Separator` at the end?

kvii: 别着急合并啊。有 bug！

kvii: I've created a new PR#5844 to solve it.

PR #5837, opened 2022-11-04
feat(Scopes): make Schema and Table accessible from scopes

System-Glitch:
[Original PR Ref](https://github.com/go-gorm/gorm/pull/5810)

- [x] Do only one thing
- [x] Tested

I'm open to feedback if you think there is a better way to do it.

This PR executes scopes after parsing the model so `Statement.Schema` and `Statement.Table` are accessible from them. 

This PR defers from the first one by the fact it still allows the model to be set from scopes.

### User Case Description

This would be useful in order to implement scopes that are re-usable across multiple models.

An example would be a scope that sorts by ID. The following would not work if there is a join or an ambiguity with the table name:
```go
func sortByID(db *gorm.DB) *gorm.DB) {
    return db.Sort("id")
}
```

Thanks to the change suggested in this PR, we would be able to do the following:
```go
func sortByID(db *gorm.DB) *gorm.DB) {
    return db.Sort(fmt.Sprintf("%s.id", db.Statement.Table))
}
```

This would also open a lot of possibilities for automatic query generation based on a schema.

### DIFF
--- a/callbacks.go
+++ b/callbacks.go
@@ -73,15 +73,6 @@ func (cs *callbacks) Raw() *processor {
 }
 
 func (p *processor) Execute(db *DB) *DB {
-	// call scopes
-	for len(db.Statement.scopes) > 0 {
-		scopes := db.Statement.scopes
-		db.Statement.scopes = nil
-		for _, scope := range scopes {
-			db = scope(db)
-		}
-	}
-
 	var (
 		curTime           = time.Now()
 		stmt              = db.Statement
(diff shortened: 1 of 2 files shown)


jinzhu:
Hi, how about implementing your requirements like this? GORM provides some placeholders that would be filled according to the current model, like `table name`, `primary key`

e.g:

```go
func sortByID(db *gorm.DB) *gorm.DB) {
    return db.Order(clause.OrderByColumn{
		Column: clause.Column{Table: clause.CurrentTable, Name: clause.PrimaryKey},
	})
}
```

System-Glitch:
Let me take a more advanced example to illustrate the benefits of this. I am working on a library that takes HTTP query parameters and turn them into an SQL query for easy and dynamic filtering, sorting, etc. I need the model information to generate a query that won't result in a DB error. I also use the model information to automatically generate joins when the user wants to filter the records based on a relation column.

Currently I am using [a workaround](https://github.com/go-goyave/filter/blob/1402f6e3e9332f209e011ea1071dffac3e561e10/settings.go#L60) to get the schema. With this change, the library wouldn't have to duplicate the schema parsing.

Here is a link to the library if you want more details: https://github.com/go-goyave/filter

However, I didn't know`clause.CurrentTable` worked to that extent. It looks useful and covers a lot of cases.

System-Glitch:
Closing since I can do without for now. If this ever becomes a necessity, I think the authors will find a better implementation.

Issue #5840, opened 2022-11-04
Use of a table alias does not work with soft delete


robhafner:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/541

## Description

Our REST API supports an expression filter parameter which is parsed and turned into generated sql code using table aliases. This works fine as long as the model doesn't support soft deletes. If the model supports soft delete the query fails due to the generated is null clause for the soft delete column being generated with as "[table-name]".[soft-delete-column]".

robhafner:
Linking to PR with fix:
https://github.com/go-gorm/gorm/pull/5841

yaofeng-wang:
Hi @robhafner, may I ask if you are still working on this issue? If not, I would like to work on it.

robhafner:
Hi @yaofeng-wang, feel free to work on it as I have not had time lately to take another look at it. Thanks!

PR #5841, opened 2022-11-04
fix: remove table name alias from soft delete column (#5840)

robhafner:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested

Remove use of the table name in the is null clause generated for soft delete model.s

### User Case Description

See: https://github.com/go-gorm/gorm/issues/5840

### DIFF
--- a/soft_delete.go
+++ b/soft_delete.go
@@ -78,7 +78,7 @@ func (sd SoftDeleteQueryClause) ModifyStatement(stmt *Statement) {
 		}
 
 		stmt.AddClause(clause.Where{Exprs: []clause.Expression{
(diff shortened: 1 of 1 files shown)


jinzhu:
Tests failed for this PR.

robhafner:
Thanks for taking a look Jinzhu. After reviewing the failures it seems this PR will not work due to queries which perform joins and contain a soft delete column of the same name. Those queries require the "table-name" alias currently.  Have you ever considered adding a TableAlias(string) function to *gorm.DB? For example:

From:
DB.Table("soft_delete_users sds").Where("sds.id = ?", 1).Find(&results).Error;

To:
DB.Table("soft_delete_users").TableAlias("sds").Where("sds.id = ?", 1).Find(&results).Error;

Maybe something like that would allow the table alias to be customized in the soft delete behavior.

PR #5388, opened 2022-05-30
enhancement: Avoid calling reflect.New() when passing in slice of values to `Scan()`

Bexanderthebex:
### What did this pull request do?

Tries to optimize `Scan()` method by recycling a struct when a user passes a slice of values to prevent calling `reflect.New()` every time regardless of the type.

### User Case Description

When passing a slice of values, the run time is the same as passing in a slice of pointers. Users should not be penalized for such.

Issue #5575, opened 2022-08-04
Regression of Scan when loading pointer object in a struct


tab1293:
## GORM Playground Link
https://github.com/go-gorm/playground/pull/504

## Description
There seems to have been a regression from `v1.23.5` to `v1.23.6` where loading objects with a pointer struct embed will result in that pointer struct being duplicated across all objects loaded. The commit that caused the regression seems to be this: https://github.com/go-gorm/gorm/commit/d01de7232b46987e239ef19a89d9ab192f453894. There are already two comments on this commit mentioning this behavior. 

Can we expect a fix to change this behavior back to not duplicating embed structs across object loads?

a631807682:
related to https://github.com/go-gorm/gorm/issues/5431

ianeal:
I've also observed this issue which is blocking us from upgrading.

ianeal:
@jinzhu Any update on this one?  I tried with gorm v1.23.9 and the problem still exists.  I can put together a sample program if that would help.

tab1293:
@jinzhu any thoughts around the behavior change introduced with this update?

miraclesu:
@jinzhu @a631807682 reproduction code

```
type User struct {
	*Embedded
}

type Embedded struct {
	ID   int
	Name string
}

func main() {
	db.AutoMigrate(&User{})
	db.Create([]User{
		{&Embedded{
			ID:   1,
			Name: "name1",
		}},
		{&Embedded{
			ID:   2,
			Name: "name2",
		}},
	})
	users := make([]User, 0, 2)
	db.Find(&users)
	// github.com/davecgh/go-spew/spew
	spew.Dump(users)
}
```

output
```
([]main.User) (len=2 cap=2) {
 (main.User) {
  Embedded: (*main.Embedded)(0xc0014a15f0)({
   ID: (int) 2,
   Name: (string) (len=5) "name2"
  })
 },
 (main.User) {
  Embedded: (*main.Embedded)(0xc0014a15f0)({
   ID: (int) 2,
   Name: (string) (len=5) "name2"
  })
 }
}
```

tsipo:
@jinzhu any updates on a fix for this one (or a plan when to fix it)? We are still holding back to `v1.23.5` (and the corresponding drivers), and this issue is blocking us from moving forward with new releases. Thanks!

yyoshiki41:
Duplicated issues: https://github.com/go-gorm/gorm/issues/5838

PR #5901, opened 2022-12-05
Add test case for embedded value selects

emcfarlane:
Adds a testcase for embedded fields when selecting multiple row values. Current behaviour returns the last row value for the embbed ptr to all the structs.

Adds a testcase currently. Looking into the issue.

```
--- FAIL: TestEmbeddedStruct (0.01s)
    embedded_struct_test.go:92: 0 &{ George }
    embedded_struct_test.go:94: expected author Edward got George
    embedded_struct_test.go:92: 1 &{ George }
```

### User Case Description

Fix embedded struct ptrs. Reverted the optimisation for reusing structs here: https://github.com/go-gorm/gorm/pull/5388

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -65,7 +65,6 @@ func (db *DB) scanIntoStruct(rows Rows, reflectValue reflect.Value, values []int
 
 	db.RowsAffected++
 	db.AddError(rows.Scan(values...))
-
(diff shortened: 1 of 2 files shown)


emcfarlane:
https://github.com/go-gorm/gorm/issues/5575

jinzhu (approved):

PR #5920, opened 2022-12-12
fix: save reflect err

xtpv:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested


finisher_api.go
```diff
@@ -75,7 +75,12 @@ func (db *DB) Save(value interface{}) (tx *DB) {
 
 	reflectValue := reflect.Indirect(reflect.ValueOf(value))
 	for reflectValue.Kind() == reflect.Ptr || reflectValue.Kind() == reflect.Interface {
-		reflectValue = reflect.Indirect(reflectValue)
```
a631807682, line 78: `reflect.Indirect` returns itself if it is not a pointer, just remove `reflectValue.Kind() == reflect.Interface` .
Please add unit tests to describe it.

jinzhu:
I don't get what got fixed by this PR, @xtpv  can you add some tests?

PR #5928, opened 2022-12-17
Add a private scanner

jiangchangren:
Add a private scanner interface.
We want to have an efficient scanner in sometime.

### User Case Description
For example:
```
func (r *Rows) Scan(values []interface{}, columns []string) {
     for idx, column := range columns {
        if values[idx] == nil {
	        continue
        }
        var value interface{}
        if v, ok := values[idx].(*interface{}); ok {
	        value = *v
        }
        if value == nil {
	        continue
        }
        switch it := value.(type) {
		case []byte:
                 case string:
                 case int64:
                 case float64:
                 case time.Time:
         }
     }
}
```

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -465,7 +465,7 @@ func (db *DB) Count(count *int64) (tx *DB) {
 	tx.Statement.Dest = count
 	tx = tx.callbacks.Query().Execute(tx)
 
-	if _, ok := db.Statement.Clauses["GROUP BY"]; ok || tx.RowsAffected != 1 {
+	if tx.RowsAffected != 1 {
 		*count = tx.RowsAffected
 	}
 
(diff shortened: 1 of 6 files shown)


jinzhu:
Hi @jiangchangren 

Can you describe more about your use case? Why not use the `rows` interface? which should be even more effective

We should be careful to add another interface if it is not that necessary.

jiangchangren:
> Hi @jiangchangren
> 
> Can you describe more about your use case? Why not use the `rows` interface? which should be even more effective
> 
> We should be careful to add another interface if it is not that necessary.

I get 50000 rows from db and I don't need structs or map.Because json(encode/decode) spend many time, more than 100ms.
So I just read the value from db and write it into bytes.Buffer.I join a json string and return to client.
I just first time to use gorm. Maybe you can help me another way.

jinzhu:
> > Hi @jiangchangren
> > Can you describe more about your use case? Why not use the `rows` interface? which should be even more effective
> > We should be careful to add another interface if it is not that necessary.
> 
> I get 50000 rows from db and I don't need structs or map.Because json(encode/decode) spend many time, more than 100ms. So I just read the value from db and write it into bytes.Buffer.I join a json string and return to client. I just first time to use gorm. Maybe you can help me another way.

I think it's better to use the `*sql.Rows` to scan the value, it should have better performance.

Thank you for your PR.

PR #5940, opened 2022-12-24
fix: support implicit table alias

### LINKED ISSUE #5840: Use of a table alias does not work with soft delete

yaofeng-wang:
- [X] Do only one thing
- [X] Tested

Resolves #5840.

### DIFF
--- a/chainable_api.go
+++ b/chainable_api.go
@@ -55,7 +55,7 @@ func (db *DB) Clauses(conds ...clause.Expression) (tx *DB) {
 	return
 }
 
-var tableRegexp = regexp.MustCompile(`(?i).+? AS (\w+)\s*(?:$|,)`)
+var tableRegexp = regexp.MustCompile(`(?i).+?(?: AS )?(\w+)\s*(?:$|,)`)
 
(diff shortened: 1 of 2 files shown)


chainable_api.go
```diff
@@ -55,7 +55,7 @@ func (db *DB) Clauses(conds ...clause.Expression) (tx *DB) {
 	return
 }
 
-var tableRegexp = regexp.MustCompile(`(?i).+? AS (\w+)\s*(?:$|,)`)
+var tableRegexp = regexp.MustCompile(`(?i).+?(?: AS )?(\w+)\s*(?:$|,)`)
```
jinzhu, line 58: the regexp is too general, will break in some cases, for example:

```go
db.Table("select * from users where name = 1") 
```

Issue #5957, opened 2023-01-04
tag `uniqueIndex` always cause a field migration


darkelf21cn:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/1

## Description

code to repo
```golang
func main() {
	mc := mysql.Config{
		DSN:               "root:123456@tcp(localhost:3306)/test?charset=utf8&parseTime=True",
		DefaultStringSize: 128, // default length of string type field
	}
	db, _ := gorm.Open(mysql.New(mc))
	db.Debug().AutoMigrate(&TestTable{})
}

type TestTable struct {
	C1 string `gorm:"uniqueIndex"`
	C2 int    `gorm:"uniqueIndex"`
}
```

logs
```log
2023/01/04 16:53:33 c:/GoPath/src/test/main.go:14
[52.801ms] [rows:0] ALTER TABLE `test_tables` MODIFY COLUMN `c1` varchar(128)

2023/01/04 16:53:33 c:/GoPath/src/test/main.go:14
[9.402ms] [rows:0] ALTER TABLE `test_tables` MODIFY COLUMN `c2` bigint
```

PR #5958, opened 2023-01-04
fix(migrator) Tag `uniqueIndex` always causes a field migration #5957

darkelf21cn:
- [x] Do only one thing
- [x] Tested

Fix 5957


darkelf21cn:
Has many associations are not correctly created after this patch. Will do more research and close this PR for now.

Issue #5752, opened 2022-10-11
Field should be unique but is not, causing unnecessary ALTER TABLE executions


siggy:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/525

## Description

Defining a model field with the `uniqueIndex` tag:
```go
type UniqueTest struct {
	UniqueIndex string `gorm:"size:191;uniqueIndex"`
}
```

`AutoMigrate` will always execute `ALTER TABLE`, because it does not recognize the field as unique, due to failing uniqueness check:
https://github.com/go-gorm/gorm/blob/2c56954cb12dd33fc8f1875a735091d61daff702/migrator/migrator.go#L464-L470

... presumably because the field does not have a unique tag:
https://github.com/go-gorm/gorm/blob/2c56954cb12dd33fc8f1875a735091d61daff702/schema/field.go#L116

Defining the model this way fixes the issue:
```go
type UniqueTest struct {
	UniqueIndex string `gorm:"size:191;unique;uniqueIndex"`
}
```

This is reproducible with `sqlite`, `mysql`, and `postgres`:
```bash
GORM_DIALECT=sqlite GORM_ENABLE_CACHE=true ./test.sh
GORM_DIALECT=mysql GORM_ENABLE_CACHE=true ./test.sh
GORM_DIALECT=postgres GORM_ENABLE_CACHE=true ./test.sh
```

rwrz:
So true!
I discovered why and found a way to fix it.

They have changed how the "unique" field is defined (this line):
```
Unique:                 utils.CheckTruth(tagSetting["UNIQUE"]), 
```

So, intead of using `index:,unique` OR `uniqueIndex`, change to `unique`, like this:

```golang
type UniqueTest struct {
	UniqueIndex string `gorm:"size:191;unique"`
}
```

It will fix the problem.

rwrz:
Issues related: #5556 , #5341 and probably more.
@jinzhu researching more... seems there is a discrepancy between what migrators drivers considers as a *Unique ColumnField* and GORM's Fields (from the reflection).

GORM considers unique fields if it has this TAG: `unique` (could not find anything on the DOCs about it).
Drivers considers a unique field if it has a `unique constraint`.

GORM considers `uniqueIndex` or `index:,unique` as an index to be created, not as a FIELD property.

[...]

PR #5982, opened 2023-01-13
fix: issue #5752 Field should be unique but is not

chirino:
causing unnecessary ALTER TABLE executions

- [X] Do only one thing
- [X] Tested

On cockroachdb, auto migrating a table with a field like:

    PublicKey string      `gorm:"uniqueIndex" json:"public_key"`

Will create the table and then create the unique index, and then do an alter to create the unique index again which causes an error.  This change fixes that.

### User Case Description

auto migrating a table with a field with a uniqueIndex tag.


jinzhu:
Thank you for your PR, Going to close this PR for now as tests failed.

PR #5990, opened 2023-01-18
fix update panic if has hook

### LINKED ISSUE #5989: if model is not ptr when update record, be panic
## GORM Playground Link

https://github.com/go-gorm/playground/pull/557

Patrick0308:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested

fix #5989  

By this way that if model is  not ptr and the hook is for ptr, the hook will be ignore and it will be not panic.

``` golang
DB.Model(User{}).Where("name = ?", "test").Update("name", "test1")

// will not be executed
func (*User) BeforeUpdate(*gorm.DB) error {
	return nil
}

// be executed
func (User) BeforeUpdate(*gorm.DB) error {
	return nil
}
```

### DIFF
--- a/callbacks/callmethod.go
+++ b/callbacks/callmethod.go
@@ -13,11 +13,18 @@ func callMethod(db *gorm.DB, fc func(value interface{}, tx *gorm.DB) bool) {
 		case reflect.Slice, reflect.Array:
 			db.Statement.CurDestIndex = 0
 			for i := 0; i < db.Statement.ReflectValue.Len(); i++ {
-				fc(reflect.Indirect(db.Statement.ReflectValue.Index(i)).Addr().Interface(), tx)
+				value := reflect.Indirect(db.Statement.ReflectValue.Index(i))
+				if value.CanAddr() {
+					fc(value.Addr().Interface(), tx)
+				} else {
+					fc(value.Interface(), tx)
+				}
 				db.Statement.CurDestIndex++
 			}
 		case reflect.Struct:
-			fc(db.Statement.ReflectValue.Addr().Interface(), tx)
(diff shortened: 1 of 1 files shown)


a631807682:
@Patrick0308 Can you add some tests?

Patrick0308:
@a631807682 
```
DB.Model(User{}).Where("name = ?", "test").Update("name", "test1")

// be executed
func (u User) BeforeUpdate(*gorm.DB) error {
        fmt.Println(u.Name) // println empty
	return nil
}
```
Is it this phenomenon expected.

a631807682:
@Patrick0308 We need to use unit test to make sure this changes don't break by other commit

Patrick0308:
@a631807682 I get it. I will add unit tests. 
But there's just one thing I don't understand. Hooks be executed too before this pr is committed when model is not ptr .  Cause of the model object is empty, hooks be executed seems meaningless.

black-06:
> @a631807682 I get it. I will add unit tests. But there's just one thing I don't understand. Hooks be executed too before this pr is committed when model is not ptr . Cause of the model object is empty, hooks be executed seems meaningless.

for example, modify some gobal vars ?

I also opened a pull request #6037 but the hook will still be executed when the model is not ptr.

a631807682:
> @a631807682 I get it. I will add unit tests. But there's just one thing I don't understand. Hooks be executed too before this pr is committed when model is not ptr . Cause of the model object is empty, hooks be executed seems meaningless.

It's necessary to execut for not addressable model, maybe we can raise an error.

jinzhu:
This PR should be resolved by this one? https://github.com/go-gorm/gorm/pull/6037/files

Issue #5975, opened 2023-01-12
logger.Default should use os.Stderr instead of os.Stdout


hidu:
## Description
logs should be written to stderr not stdout.

But now:
https://github.com/go-gorm/gorm/blob/baf1afa1fcb45b69a7c64c3fb82da7a0dd32bcfc/logger/logger.go#L75

Expect:
```go
Default = New(log.New(os.Stderr, "\r\n", log.LstdFlags), Config{
```

PR #5991, opened 2023-01-18
fix: logger.Default use os.Stderr instead of os.Stdout #5975

hidu:
- [] Do only one thing
- [] Non breaking API changes
- [] Tested

logger.Default use os.Stderr instead of os.Stdout


jinzhu:
Sorry, but we are not going to change the default logger, but you can customize it for yourself.

Issue #6002, opened 2023-01-26
Add error variable to check if an error is "Unique Constraint Violation" type


Henelik:
## Describe the feature
Add an error var in `errors.go` for unique constraint violations, so GORM users can check for that error type across all supported DBs and drivers.

## Motivation
Apps that support multiple databases require complex workarounds or extra queries to return a correct error to API consumers when a unique key constraint is violated.  GORM users should be able to check for this error type regardless of DB or driver used, without cumbersome code or additional imports.

## Related Issues
Closed with only partly-working workarounds: https://github.com/go-gorm/gorm/issues/4037

saeidee:
Closing the issue since this feature has been add already.
https://github.com/go-gorm/gorm/pull/6004

Issue #5651, opened 2022-08-28
Please add "gorm.ErrUniqueViolation" support


silvioprog:
## Describe the feature

A new error type like `gorm.ErrUniqueViolation` would allow handling errors related to unique violation.

## Motivation

Since GORM allows handling common DB errors, it would be nice to support handling errors related to unique violation too.

## Related Issues

1. https://github.com/go-gorm/gorm/issues/4037
2. https://github.com/go-gorm/gorm/issues/4135

Also this at StackOverflow:

[GORM create record that might already exist](https://stackoverflow.com/questions/57749720/gorm-create-record-that-might-already-exist)

silvioprog:
Edit: it would be useful even if declared in the driver (sqlite, postgres, mysql etc.).

sleepreading:
This error type is widely used, so we all hope gorm will consider adding this feature instead of closing it, again!

artlevitan:
Use this function to check for errors in SQLite 3

```
func gormErrUniqueViolation(err error) bool {
	e := errors.New("UNIQUE constraint failed")
	return !errors.Is(err, e)
}
```

Example after the INSERT request:
```
// ...
err = db.Create(&data).Error
	if err != nil { // Check SQL Errors
		**if gormErrUniqueViolation(err) { // UNIQUE constraint failed
			return
		}**
		// ...
	}
// ...
```

saeidee:
Closing the issue since this feature has been add already.
https://github.com/go-gorm/gorm/pull/6004

Issue #4135, opened 2021-03-04
POSTGRES How to check if error type is DUPLICATE KEY


khanakia:
@jinzhu 

OLD Ticket: https://github.com/go-gorm/gorm/issues/4037

I did not find any solution for postgres yet. I searched everywhere but still there is no way i found to check the duplicate key for postgres. @jinzhu Can you provide the PSQL example like you did for the Mysql in old ticket. Thanks

## Your Question

I have **unique index key** and when i try to insert duplicate record i get this error `ERROR: duplicate key value violates unique constraint "idx_itemid_partid" (SQLSTATE 23505)` but i want to show different message to the frontend based on ERROR TYPE but in gorm i do not see any Duplicate error type please check screenshot. How do i use the validation for duplicate error type
```
  if err != nil && errors.Is(err, gorm.ErrDuplicateKey) { // how do i do this ?
    fmt.Println("Duplicate key found.")
  }
```

<img width="787" alt="Screenshot 2021-02-01 at 4 20 21 PM" src="https://user-images.githubusercontent.com/2174170/106448898-631c9400-64a9-11eb-84ee-4de2790f6893.png">

## The document you expected this should be explained
N/A

## Expected answer

Should have error type duplicate

spiside:
This [stackoverflow answer](https://stackoverflow.com/a/37560627) should provide some details on how errors are commonly caught using Postgres error codes. For your problem specifically, I think you can do the following:

```go
const (
    UniqueViolationErr = pq.ErrorCode("23505")
)

if err != nil && errors.Is(err, UniqueViolationErr) {
    fmt.Println("Duplicate key found.")
}
```

Here's [a list of all the SQLSTATE](https://www.postgresql.org/docs/current/errcodes-appendix.html) error codes that are used in Postgres. You'll see that `23505` is unique violation error.

khanakia:
@spiside 
Your code is not working and giving error please check. 
FYI: I tried creating some custom errors already in the past with Error Interface those also did not work.

So can you please give some solid solution to these issue. Thanks

<img width="920" alt="Screenshot 2021-03-04 at 5 49 10 PM" src="https://user-images.githubusercontent.com/2174170/109963159-17604300-7d12-11eb-84fd-5a0709f150b9.png">

spiside:
I was hoping `Is()` might do the heavy lifting here but this is how I've defined a function that checks postgres error codes:

```
const (
    UniqueViolationErr = pq.ErrorCode("23505")
[...]

Issue #4037, opened 2021-02-01
How to check if error type is DUPLICATE KEY


khanakia:
## Your Question

I have **unique index key** and when i try to insert duplicate record i get this error `ERROR: duplicate key value violates unique constraint "idx_itemid_partid" (SQLSTATE 23505)` but i want to show different message to the frontend based on ERROR TYPE but in gorm i do not see any Duplicate error type please check screenshot. How do i use the validation for duplicate error type
```
  if err != nil && errors.Is(err, gorm.ErrDuplicateKey) { // how do i do this ?
    fmt.Println("Duplicate key found.")
  }
```

<img width="787" alt="Screenshot 2021-02-01 at 4 20 21 PM" src="https://user-images.githubusercontent.com/2174170/106448898-631c9400-64a9-11eb-84ee-4de2790f6893.png">

## The document you expected this should be explained
N/A

## Expected answer

Should have error type duplicate

jinzhu:
use mysql as example, check it like:

```go
var mysqlErr mysql.MySQLError
if errors.As(err, &mysqlErr) && mysqlErr.Number == 1062 {
}
```

FedorLap2006:
I think better would be easier to have such common error in predefined errors emitted by gorm.

khanakia:
@jinzhu 
How do i do that for postgres it's not working

```
var psqlErr postgers.PostgreError
if errors.As(err, &psqlErr) && psqlErr.Number == 1062 {
}
```

khanakia:
@jinzhu 
Any update on this ? I have check everywhere there is no solution i found for this.

GiancarlosIO:
👀

khanakia:
@GiancarlosIO What ??

jinagamvasubabu:
Currently i am able to solve this issue with a hack (even i expect `gorm` should solve it in a better way)

```go
import "github.com/jackc/pgconn"
....
....
if strings.Contains(err.(*pgconn.PgError).Message, "duplicate key value violates unique constraint") {}
```

bitbreakr:
@khanakia 

Hope it helps: 

`golang
if results :=  db.Create(&account); results.Error != nil {
			if pgError := results.Error.(*pgconn.PgError); errors.Is(results.Error, pgError) {
				var httpStatus int
				switch pgError.Code {
				case "23505":
					httpStatus = 409
				}
				
				c.Status(httpStatus)
				
				return nil
			}

[...]

PR #6004, opened 2023-01-29
feat: Unique Constraint Violation error translator for different drivers

saeidee:
- [x] Do only one thing
- [x] Tested

It will enable developers to have a better error response when there is a unique key constraint violation detected in different SQL engines.

Addresses these issues: https://github.com/go-gorm/gorm/issues/4037, https://github.com/go-gorm/gorm/issues/4135, https://github.com/go-gorm/gorm/issues/5651,  https://github.com/go-gorm/gorm/issues/6002. 

There is no new import or any dependency on any driver, the implementation works by marshaling the returned error from any query and comparing it to SQL engine's default error message by looking after the error code.

```go
err = DB.Create(&City{Name: "Madrid"}).Error
if errors.Is(err, gorm.ErrDuplicatedKey) {
    // some logic here
}
```

### User Case Description

When there is a unique key constraint violation error it will return `gorm.ErrDuplicatedKey`. Basically, it will look if there is a translation available for a specific type of error, if there was one it will return that translation, otherwise, it will return the given untranslated error.


a631807682:
That's nice, but we'd prefer it to be a [plugin](https://gorm.io/docs/write_plugins.html#Plugin) , or implemented per [drivers](https://github.com/orgs/go-gorm/repositories?q=driver&type=all&language=&sort=) rather than in gorm.

jinzhu:
Maybe we can add `TranslateError` support to dialector.

If the dialector implements `TranslateError` support, will use it to translate the error when `AddError` for db.

saeidee:
> Maybe we can add `TranslateError` support to dialector.
> 
> If the dialector implements `TranslateError` support, will use it to translate the error when `AddError` for db.

This is a good idea, I will update the PR & also create a new PR for one of the drivers.

a631807682 (approved):

gorm.go
```diff
@@ -347,6 +347,10 @@ func (db *DB) Callback() *callbacks {
 
 // AddError add error to db
 func (db *DB) AddError(err error) error {
+	if errTranslator, ok := db.Dialector.(ErrorTranslator); ok {
```
a631807682, line 350: We don't need to convert when `err == nil`, are you interested in creating an other PR for it?

saeidee: Yes sure

ZhouXing19:
Hello, will this new `ErrDuplicatedKey` shows up in the next patch (i.e. 1.24.7)?  Thanks!

guillermo-st:
> Hello, will this new `ErrDuplicatedKey` appear in the next patch (i.e. 1.24.7)? Thanks!

I would like to know too! And while I'm at it, how should I go about updating my project dependencies to use the new gorm release that has this error included?

saeidee:
> > Hello, will this new `ErrDuplicatedKey` appear in the next patch (i.e. 1.24.7)? Thanks!
> 
> I would like to know too! And while I'm at it, how should I go about updating my project dependencies to use the new gorm release that has this error included?

Documentation has been added https://github.com/go-gorm/gorm.io/pull/664

PR #6031, opened 2023-02-06
fix:throw model value required error

### LINKED ISSUE #6030: Feature Request: Model validation
## Describe the feature

run the code below
```
package main

import (
	"gorm.io/driver/sqlite"
	"gorm.io/gorm"
)

type Product struct {
	id    int
	code  string
	price uint
}

func main() {
	db, err := gorm.Open(sqlite.Open("test.db"), &gorm.Config{})
	if err != nil {
	}
	db.AutoMigrate(&Product{})
	var product Product
	db.First(&product, 1)                
}
```

we got an error 
```
2023/02/06 12:49:57 /内容削除/main.go:21 near "=": syntax error
[0.010ms] [rows:0] SELECT * FROM `products` WHERE `products`. = 1 ORDER BY `products`. LIMIT 1
```

we can see that GORM generated a sql query without column name because there is no Primary Key Definition in `Product` struct.

I hope we can get a warning when there is no public field or no primary key.

## Motivation

UX

a631807682:
- [x] Do only one thing
- [x] Tested

close #6030

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -121,7 +121,7 @@ func (stmt *Statement) QuoteTo(writer clause.Writer, field interface{}) {
 			} else if len(stmt.Schema.DBNames) > 0 {
 				write(v.Raw, stmt.Schema.DBNames[0])
 			} else {
-				stmt.DB.AddError(ErrModelValueRequired)
+				stmt.DB.AddError(ErrModelValueRequired) //nolint:typecheck
 			}
 		} else {
 			write(v.Raw, v.Name)
--- a/statement.go
+++ b/statement.go
@@ -121,7 +121,7 @@ func (stmt *Statement) QuoteTo(writer clause.Writer, field interface{}) {
 			} else if len(stmt.Schema.DBNames) > 0 {
 				write(v.Raw, stmt.Schema.DBNames[0])
 			} else {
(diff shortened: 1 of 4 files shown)


statement.go
```diff
@@ -120,6 +120,8 @@ func (stmt *Statement) QuoteTo(writer clause.Writer, field interface{}) {
 				write(v.Raw, stmt.Schema.PrioritizedPrimaryField.DBName)
 			} else if len(stmt.Schema.DBNames) > 0 {
 				write(v.Raw, stmt.Schema.DBNames[0])
+			} else {
+				stmt.DB.AddError(ErrModelValueRequired) //nolint:typecheck,errcheck
```
ExerciseBook, line 124: Why we don't use `ErrPrimaryKeyRequired`.

I am not very familiar with gorm. Maybe I asked a stupid question.

I am not very sure what the meaning `ErrPrimaryKeyRequired` and `ErrModelValueRequired` are.

a631807682: @ExerciseBook This has nothing to do with the primary key, it's because there aren't any accessible fields in the model.
The use of `ErrModelValueRequired` here is indeed misleading. It prompts the user to use the model, but the problem here is that the user uses the wrong model.

jinzhu (approved):

PR #6037, opened 2023-02-07
fix: update panic if model is not ptr

### LINKED ISSUE #5989: if model is not ptr when update record, be panic

black-06:
- [X] Do only one thing
- [X] Tested

fix #5989.
The same pull request is #5990.

### User Case Description

`DB.Model(Product5{}).Where("id", p.ID).Update("name", "update_name")`


callbacks/callmethod.go
```diff
@@ -17,6 +17,9 @@ func callMethod(db *gorm.DB, fc func(value interface{}, tx *gorm.DB) bool) {
 				db.Statement.CurDestIndex++
 			}
 		case reflect.Struct:
+			if !db.Statement.ReflectValue.CanAddr() {
+				db.Statement.ReflectValue = reflect.New(db.Statement.ReflectValue.Type()).Elem()
```
jinzhu, line 21: I think maybe better to raise an error?

a631807682, line 21: Modifying ReflectValue may produce unexpected results, there is no need to modify it here.
Actually I don't think it's necessary to implement it for not addressable model.

callbacks/callmethod.go
```diff
@@ -17,6 +17,9 @@ func callMethod(db *gorm.DB, fc func(value interface{}, tx *gorm.DB) bool) {
 				db.Statement.CurDestIndex++
```
a631807682, line 17: This problem also exists when the kind is `reflect.Array`

a631807682 (changes requested):

callbacks/callmethod.go
black-06: Do you mean that param of `Model()` method must be ptr?

black-06: just like #5990?
so we close this and merge 5990

a631807682: #5990 still has some issues, not re-reviewed since it hasn't been updated recently.

black-06: Updated

schema/utils.go
```diff
@@ -134,7 +134,11 @@ func GetIdentityFieldValuesMap(ctx context.Context, reflectValue reflect.Value,
 			elem := reflectValue.Index(i)
 			elemKey := elem.Interface()
 			if elem.Kind() != reflect.Ptr {
-				elemKey = elem.Addr().Interface()
+				if elem.CanAddr() {
+					elemKey = elem.Addr().Interface()
+				} else {
+					elemKey = elem.Interface()
```
a631807682, line 140: is it necessary?

a631807682 (approved):

callbacks/callmethod.go
```diff
@@ -13,11 +13,15 @@ func callMethod(db *gorm.DB, fc func(value interface{}, tx *gorm.DB) bool) {
 		case reflect.Slice, reflect.Array:
 			db.Statement.CurDestIndex = 0
 			for i := 0; i < db.Statement.ReflectValue.Len(); i++ {
-				fc(reflect.Indirect(db.Statement.ReflectValue.Index(i)).Addr().Interface(), tx)
+				if value := reflect.Indirect(db.Statement.ReflectValue.Index(i)); value.CanAddr() {
+					fc(value.Addr().Interface(), tx)
+				}
```
jinzhu, line 18: Should we raise an error if the value is not addressable? it should be ok to return an error in those cases would panic before.

black-06: `err = DB.Model(Product5{}).Where("id", p.ID).Update("name", "update_name_2").Error`

If so, this usage will raise an error, I'm not sure if it's good, it's up to you.

black-06: > `err = DB.Model(Product5{}).Where("id", p.ID).Update("name", "update_name_2").Error`
> 
> If so, this usage will raise an error, I'm not sure if it's good, it's up to you.

I thought I sent this message yesterday, but in fact it was pending.... 😨

callbacks/callmethod.go
```diff
@@ -15,12 +15,16 @@ func callMethod(db *gorm.DB, fc func(value interface{}, tx *gorm.DB) bool) {
 			for i := 0; i < db.Statement.ReflectValue.Len(); i++ {
 				if value := reflect.Indirect(db.Statement.ReflectValue.Index(i)); value.CanAddr() {
 					fc(value.Addr().Interface(), tx)
+				} else {
+					db.AddError(gorm.ErrInvalidValue)
```
a631807682, line 19: add `return` after `AddError`

black-06: done

Issue #5801, opened 2022-10-22
Race when concurrently appending m2m association when preloaded model


Mevaser:
## GORM Playground Link
https://github.com/go-gorm/playground/pull/534

## Description
I tried appending to a m2m association inside a concurrent goroutines and for some reason it appended only some of the data.
While creating a minimal code example to test this i narrowed it down to this: 
- When the model is fetched from the db(previously created) & the associations are preloaded(Preload(clause.Associations)), it fails.
- If any of this changes the test passes(note the comments in the code)

a631807682:
https://github.com/go-gorm/playground/pull/534/files#diff-79ce3229c13921b79b1175dcb211336aaf84dfd8edcf19c07698934d4fe70e5eR47

This error seems to be a race caused by `userCopy.Languages` even though each `userCopy` is different.
I haven't found the bug yet, but this way it can be avoided.
```diff
+ uuserCopy.Languages = nil
err := DB.Model(&userCopy).Association("Languages").Append(&languageCopy)
```

PR #6044, opened 2023-02-09
fix: association concurrently appending

black-06:
- [X] Do only one thing
- [X] Tested

Fix bug when association concurrently appending.

`Association` in each goroutine will modify the user.Languages by `reflect.Append`, which method grows the slice.
If it grows successfully,  languages will be a new slice, that's ok.
But If cap is enough, these goroutines will write the same slice, that's wrong.

If we only call `DB.FirstOrCreate(&user)`, user.Languages len is 0 and cap is 0, that's ok.
if we only call `DB.Preload("Languages").FirstOrCreate(&user)`, still len 0 and cap 0.
But we call `DB.FirstOrCreate(&user)` first, then `DB.Preload("Languages").FirstOrCreate(&user)`, user.Languages is len 0 and cap 10, this error will occur #5801

### User Case Description

```go
DB.FirstOrCreate(&user)
DB.Preload("Languages").FirstOrCreate(&user)

var wg sync.WaitGroup
for i := 0; i < count; i++ {
wg.Add(1)
go func(user User, language Language) {
	err := DB.Model(&user).Association("Languages").Append(&language)
	AssertEqual(t, err, nil)

	wg.Done()
}(user, languages[i])
}
wg.Wait()
```

### DIFF
--- a/tests/associations_many2many_test.go
+++ b/tests/associations_many2many_test.go
@@ -356,24 +356,29 @@ func TestDuplicateMany2ManyAssociation(t *testing.T) {
 }
 
 func TestConcurrentMany2ManyAssociation(t *testing.T) {
+	db, err := OpenTestConnection()
+	if err != nil {
+		t.Fatalf("open test connection failed, err: %+v", err)
+	}
+
 	var count = 3
 
 	var languages []Language
 	for i := 0; i < count; i++ {
 		language := Language{Code: fmt.Sprintf("consurrent %d", i)}
-		DB.Create(&language)
+		db.Create(&language)
 		languages = append(languages, language)
 	}
 
 	user := User{}
-	DB.Create(&user)
-	DB.Preload("Languages").FirstOrCreate(&user)
+	db.Create(&user)
+	db.Preload("Languages").FirstOrCreate(&user)
 
 	var wg sync.WaitGroup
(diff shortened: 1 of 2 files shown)


black-06:
When I added a new test, the old test failed. Did I miss anything?

a631807682 (commented):
When we use goroutine, another connection is created, causing `PRAGMA foreign_keys` to become invalid.
https://www.sqlite.org/foreignkeys.html#fk_enable

We can create a db instance specifically for it
```
	tx, err := OpenTestConnection()
	// your test
```

tests/associations_many2many_test.go
```diff
@@ -351,3 +354,40 @@ func TestDuplicateMany2ManyAssociation(t *testing.T) {
...
 	AssertEqual(t, user2, findUser2)
 }
+
+func TestConcurrentMany2ManyAssociation(t *testing.T) {
+	db, err := OpenTestConnection()
+	if err != nil {
+		t.Fatalf("open test connection failed, err: %+v", err)
+	}
+
+	var count = 3
```
a631807682, line 364: Please fix the formatting issues pointed out by gofumpt

a631807682 (approved):

demoManito (approved):

PR #6060, opened 2023-02-15
feat: disable prepare nested transaction

### LINKED ISSUE #6006: Create In Batches within transaction fails while using prepared statements
## GORM Playground Link

https://github.com/go-gorm/playground/pull/560

## Description

We would like to be able to execute a single transaction which creates a large number of table rows and then queries all the contents of that table to be returned. To do this we are using CreateInBatches within a Transaction call.

The issue occurs because we have PreparedStatement in the config set to true and the underlying driver, mysql in our case, does not allow for preparing SAVEPOINT statements

a631807682:
- [x] Do only one thing
- [x] Tested

close #6006

~~We still need to set it to true in the mysql driver.~~
<pre><code><del>config.DisablePrepareNestedTransaction = true</del>
</code></pre>
~~https://github.com/go-gorm/mysql/blob/master/mysql.go#L94~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ _book
 .idea
 vendor
 .vscode
-.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ _book
 .idea
 vendor
 .vscode
+.DS_Store
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -619,7 +619,6 @@ func (db *DB) Transaction(fc func(tx *DB) error, opts ...*sql.TxOptions) (err er
 		// nested transaction
 		if !db.DisableNestedTransaction && (!db.PrepareStmt ||
 			(db.PrepareStmt && !db.DisablePrepareNestedTransaction)) {
-
 			err = db.SavePoint(fmt.Sprintf("sp%p", fc)).Error
 			if err != nil {
 				return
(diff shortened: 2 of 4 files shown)


jinzhu:
Save point doesn't make sense for prepare statement, maybe we can just disable it...

a631807682:
> Save point doesn't make sense for prepare statement, maybe we can just disable it...

In other databases we can still use [savepoint in prepare](https://github.com/go-gorm/gorm/pull/6060/files#diff-0b1bf88d8ee1e7902b6c9242f1d9afef7fa7525bc7cb2a2e3849f2d0f4cba247R65), it may be a breaking change if it is just disabled it.

jinzhu:
> > Save point doesn't make sense for prepare statement, maybe we can just disable it...
> 
> In other databases we can still use [savepoint in prepare](https://github.com/go-gorm/gorm/pull/6060/files#diff-0b1bf88d8ee1e7902b6c9242f1d9afef7fa7525bc7cb2a2e3849f2d0f4cba247R65), it may be a breaking change if it is just disabled it.

yes, but caching savepoint statements don't really make sense, it would even cause leaks...

a631807682:
> > > Save point doesn't make sense for prepare statement, maybe we can just disable it...
> > 
> > 
> > In other databases we can still use [savepoint in prepare](https://github.com/go-gorm/gorm/pull/6060/files#diff-0b1bf88d8ee1e7902b6c9242f1d9afef7fa7525bc7cb2a2e3849f2d0f4cba247R65), it may be a breaking change if it is just disabled it.
> 
> yes, but caching savepoint statements don't really make sense, it would even cause leaks...

You are right, but the user may set `PreparedStmt` in `gorm.Open`, they don't know that savepoint is not allowed (actually the database supports it), they may even use `tx.SavePoint` directly.
maybe better to raise an error?

PR #6071, opened 2023-02-17
Allow use struct field tag in `Exec()` and `Raw()`

houbaron:
- [x] Do only one thing
- [x] Tested

Allow use struct field tag in `Exec()` and `Raw()`.

##### BEFORE
https://github.com/go-gorm/gorm/blob/540fb49bcbe07ee56c7a8a449a5504f40f50abc1/clause/expression.go#L110

##### AFTER
```go
fieldName, exist := utils.ParseTagSetting(fieldStruct.Tag.Get("gorm"), ";")["COLUMN"]
if !exist {
    fieldName = fieldStruct.Name
}
namedMap[fieldName] = reflectValue.Field(i).Interface()
```

### User Case Description

```go
var namedArgs = struct {
	Name1Alias string `gorm:"column:Name1;just:for;fun" not_exist:"meaningless"`
	Name2Alias int    `gorm:"column:Name2;use:less;whitespace "`
}{Name1Alias: "houbaron", Name2Alias: 2}

// SELECT * FROM tableName WHERE Name1="houbaron" AND Name2=2
db.Raw("SELECT * FROM tableName WHERE Name1=@Name1 AND Name2=@Name2", namedArgs)
// UPDATE tableName SET Name2=2 WHERE Name1="houbaron"
db.Exec("UPDATE tableName SET Name2=@Name2 WHERE Name1=@Name1", namedArgs)
```

### DIFF
--- a/clause/expression.go
+++ b/clause/expression.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"go/ast"
+	"gorm.io/gorm/utils"
 	"reflect"
 )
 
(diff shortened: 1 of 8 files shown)


jinzhu:
Thank you for your PR, but I prefer to use `Field` name for named arguments and the change would introduce incompatible changes

houbaron:
@jinzhu Maybe we can add a new function to avoid incompatible changes?

PR #6083, opened 2023-02-20
Distinguish custom types

black-06:
- [X] Do only one thing
- [X] Tested

Distinguish between detected and custom type (#6033).
In driver, we will skip to `getSchemaCustomType` if `CustomDataType` is specified.
See [MySQL#109](https://github.com/go-gorm/mysql/pull/109) for more context.

### User Case Description

```go
type Test struct {
    Date time.Time `gorm:"type:date"`
    Time time.Time `gorm:"type:time"`
    DefaultDateTime time.Time
}
```
```sql
CREATE TABLE `tests` (`date` date,`time` time,`default_date_time` datetime)
```

### DIFF
--- a/schema/field.go
+++ b/schema/field.go
@@ -56,7 +56,7 @@ type Field struct {
 	BindNames              []string
 	DataType               DataType
 	GORMDataType           DataType
-	CustomDataType         DataType
+	IsTaggedType           bool
 	PrimaryKey             bool
 	AutoIncrement          bool
 	AutoIncrementIncrement int64
@@ -311,7 +311,7 @@ func (schema *Schema) ParseField(fieldStruct reflect.StructField) *Field {
 		default:
 			field.DataType = DataType(val)
 		}
-		field.CustomDataType = DataType(val)
+		field.IsTaggedType = true
 	}
 
 	if field.Size == 0 {
--- a/schema/field.go
+++ b/schema/field.go
@@ -56,6 +56,7 @@ type Field struct {
 	BindNames              []string
 	DataType               DataType
 	GORMDataType           DataType
+	CustomDataType         DataType
 	PrimaryKey             bool
 	AutoIncrement          bool
 	AutoIncrementIncrement int64
@@ -310,6 +311,7 @@ func (schema *Schema) ParseField(fieldStruct reflect.StructField) *Field {
 		default:
 			field.DataType = DataType(val)
 		}
+		field.CustomDataType = DataType(val)
 	}
 
 	if field.Size == 0 {
--- a/schema/schema_test.go
+++ b/schema/schema_test.go
@@ -4,11 +4,37 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"gorm.io/gorm"
 	"gorm.io/gorm/schema"
 	"gorm.io/gorm/utils/tests"
 )
+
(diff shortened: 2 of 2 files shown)


schema/field.go
```diff
@@ -56,6 +56,7 @@ type Field struct {
 	BindNames              []string
 	DataType               DataType
 	GORMDataType           DataType
+	CustomDataType         DataType
```
a631807682, line 59: The meaning of the name CustomDataType is a bit vague to me.
It seems that we only need to know whether the user has specified the tag type, but not what the type is, because it will be equal to the datatype.

black-06: Do you have any good ideas about naming? 
How about "IsTaggedType"?

a631807682: I don't have a better suggestion, but IsTaggedType would work better for me than CustomDataType

a631807682 (approved):

jinzhu:
Currently, seems like the  `field.IsTaggedType` field is equal to checking the existing `field.TagSettings["TYPE"]`?

If so,  it doesn't really make sense to have it?

a631807682:
> Currently, seems like the `field.IsTaggedType` field is equal to checking the existing `field.TagSettings["TYPE"]`?
> 
> If so, it doesn't really make sense to have it?

This is the same thing. I hope to describe all the field information in parse. It is not easy to understand `TagSettings["type"]` in the driver, we need to see the gorm code to know why we do this.

jinzhu:
> > Currently, seems like the `field.IsTaggedType` field is equal to checking the existing `field.TagSettings["TYPE"]`?
> > If so, it doesn't really make sense to have it?
> 
> This is the same thing. I hope to describe all the field information in parse. It is not easy to understand the field.TagSettings["type"] and field.DataTyp in the driver, we need to see the gorm code to know why we do this.

The `Schema` and `Field` was thought of as low-level API, users need to know the details to use them.

But I understand your concerns, maybe we can define some development rules, and have a field to describe those information if it is used for two or three different purposes to avoid API overbloat？

a631807682:
> > > Currently, seems like the `field.IsTaggedType` field is equal to checking the existing `field.TagSettings["TYPE"]`?
> > > If so, it doesn't really make sense to have it?
> > 
> > 
> > This is the same thing. I hope to describe all the field information in parse. It is not easy to understand the field.TagSettings["type"] and field.DataTyp in the driver, we need to see the gorm code to know why we do this.
> 
> The `Schema` and `Field` was thought of as low-level API, users need to know the details to use them.
> 
> But I understand your concerns, maybe we can define some development rules, and have a field to describe those information if it is used for two or three different purposes to avoid API overbloat？

I agree

black-06:
> > > > Currently, seems like the `field.IsTaggedType` field is equal to checking the existing `field.TagSettings["TYPE"]`?
> > > > If so, it doesn't really make sense to have it?
> > > 
> > > 
> > > This is the same thing. I hope to describe all the field information in parse. It is not easy to understand the field.TagSettings["type"] and field.DataTyp in the driver, we need to see the gorm code to know why we do this.
> > 
> > 
> > The `Schema` and `Field` was thought of as low-level API, users need to know the details to use them.
> > But I understand your concerns, maybe we can define some development rules, and have a field to describe those information if it is used for two or three different purposes to avoid API overbloat？
> 
> I agree

Close it?

Issue #4966, opened 2021-12-29
Create View feature is missing, kind of very important


aukgit:
# Create View feature with Gorm

## Describe the feature

```
  db.
   View.
    CreateOrUpdate().
    Tables(....tables Names ). // strings
    Entities(....tables Names ). // interfaces 
    Select(columNames...). // 5 or 6 columns it is optional
    Join(joinClause).
    Alias(Entity, map[string]string{})
    Where(...., ...).
    ViewName("Give it a name"). // Optional, could come from tables or be placed
    Compile() // finisher method, builder, ends the chain
```

## Motivation

Database views are easily queried on compiled join queries. Currently, the framework doesn't support joins properly. It supports but requires manual work. If we have a view creation option it will reduce the work and query execution and improve the performance

## Related Issues

- [how to use gorm to create mysql view? · Issue #4617 · go-gorm/gorm](https://github.com/go-gorm/gorm/issues/4617)

## Tags

@riadevatix

## Keywords

View, ViewCreation, JoinFeauture, SqlInjection

longlihale:
thanks, It's a good idea, but maybe need to think about how to support it. :stuck_out_tongue:

nafeem-evatix:
@jinzhu and @ghost This should be a priority

sulthonzh:
Absolutely cannot wait for these features!

loeffel-io:
+1

devav2:
Please prioritize this for the upcoming releases.

egandro:
Is there any progress on this?

a631807682:
Already implemented in https://github.com/go-gorm/gorm/pull/6097

PR #6097, opened 2023-02-23
Create and drop view

black-06:
- [X] Do only one thing
- [X] Tested

create view #4966 

It's different from issue's suggestion.
I found the unimplemented `CreateView` function in `Migrator`,
then I implemented it.

About `ViewOption`:
- `Query` is a [subquey](https://gorm.io/docs/advanced_query.html#SubQuery)
- If `Replace` is true, exec `CREATE OR REPLACE`. sqlite may not support this.
- If `CheckOption` is not empty, append to sql, e.g. `WITH LOCAL CHECK OPTION`

### User Case Description

```go
DB.Migrator().CreateView("users_pets", gorm.ViewOption{Query: DB.Model(&User{}).Select("users.id, users.name")})
```

```sql
CREATE VIEW `users_pets` AS SELECT users.id, users.name FROM `users`
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -23,6 +23,8 @@ var (
 	ErrModelValueRequired = errors.New("model value required")
 	// ErrModelAccessibleFieldsRequired model accessible fields required
 	ErrModelAccessibleFieldsRequired = errors.New("model accessible fields required")
+	// ErrSubQueryRequired sub query required
+	ErrSubQueryRequired = errors.New("sub query required")
 	// ErrInvalidData unsupported data
 	ErrInvalidData = errors.New("unsupported data")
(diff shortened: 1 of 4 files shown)


a631807682 (commented):
Do we need to add parameter checking, such as what happens without option.Query? can we add comments to explain how to use it?

> sqlite may not support this.

It doesn't matter, just need to handle it in the sqlite driver

black-06:
> Do we need to add parameter checking, such as what happens without option.Query? can we add comments to explain how to use it?
> 
> > sqlite may not support this.
> 
> It doesn't matter, just need to handle it in the sqlite driver

done ~

a631807682 (approved):

PR #6127, opened 2023-03-07
Fix: Composite primary key with auto-increment value returns 0 after insert

### LINKED ISSUE #4930: composite key auto-increment returned 0 upon Create of new record
## GORM Playground Link

https://github.com/go-gorm/playground/pull/417

## Description

Use Case:
Define table with multiple primary keys (composite key)
Upon New record insert, Get ID of autoIncrement enabled field. 

Issue:
0 returned for autoIncrement enabled field. (Only when the model has 2 fields using primarykey tag)

Example:
Model Defined
image

Test Case:
image

Test Result:
image

truongns:
Fix #4930 workaround for databases that support auto-increment in composite primary key.

- [x] Do only one thing
- [x] Tested

Fix issue https://github.com/go-gorm/gorm/issues/4930

### User Case Description
In MySQL, you can have composite primary key with auto-increment 
For example:

https://github.com/nash90/playground/blob/7a82d16b1b1b626df01cf43089b302de7c2a6bd9/models.go#L63-L70
```go
type CompositeKeyProduct struct {
	ProductPointID int       `gorm:"primaryKey;autoIncrement:true;column:product_point_id"`
	LanguageCode   int       `gorm:"primaryKey;autoIncrement:false;column:language_code"`
	Code           string    `gorm:"column:code"`
	Name           string    `gorm:"column:name"`
	CreatedAt      time.Time `gorm:"column:created_at"`
	UpdatedAt      time.Time `gorm:"column:updated_at"`
}
```

Upon insert new record into `CompositeKeyProduct` with the following values:
https://github.com/nash90/playground/blob/7a82d16b1b1b626df01cf43089b302de7c2a6bd9/main_test.go#L26-L30
```go
prod := &CompositeKeyProduct{
		LanguageCode: 56,
		Code:         "56",
		Name:         "Test56",
}

res := DB.Create(prod)
```

expected 
```
prod.ProductPointID: 1
```

actual
```
prod.ProductPointID: 0
```

This Pull Request is aiming to fix the above case

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -225,6 +225,16 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 		schema.PrioritizedPrimaryField = schema.PrimaryFields[0]
 	}
 
+	// If there are multiple primary keys, the AUTOINCREMENT field is prioritized
+	if schema.PrioritizedPrimaryField == nil && len(schema.PrimaryFields) > 1 {
+		for _, field := range schema.PrimaryFields {
+			if _, ok := field.TagSettings["AUTOINCREMENT"]; ok {
+				schema.PrioritizedPrimaryField = field
+				break
+			}
+		}
+	}
+
 	for _, field := range schema.PrimaryFields {
 		schema.PrimaryFieldDBNames = append(schema.PrimaryFieldDBNames, field.DBName)
 	}
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -221,16 +221,16 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 		}
 	}
 
-	if schema.PrioritizedPrimaryField == nil && len(schema.PrimaryFields) == 1 {
-		schema.PrioritizedPrimaryField = schema.PrimaryFields[0]
-	}
-
-	// If there are multiple primary keys, the AUTOINCREMENT field is prioritized
-	if schema.PrioritizedPrimaryField == nil && len(schema.PrimaryFields) > 1 {
-		for _, field := range schema.PrimaryFields {
-			if _, ok := field.TagSettings["AUTOINCREMENT"]; ok {
-				schema.PrioritizedPrimaryField = field
-				break
+	if schema.PrioritizedPrimaryField == nil {
+		if len(schema.PrimaryFields) == 1 {
+			schema.PrioritizedPrimaryField = schema.PrimaryFields[0]
+		} else if len(schema.PrimaryFields) > 1 {
+			// If there are multiple primary keys, the AUTOINCREMENT field is prioritized
+			for _, field := range schema.PrimaryFields {
+				if field.AutoIncrement {
+					schema.PrioritizedPrimaryField = field
+					break
+				}
 			}
 		}
 	}
--- a/tests/create_test.go
+++ b/tests/create_test.go
@@ -547,3 +547,33 @@ func TestFirstOrCreateRowsAffected(t *testing.T) {
 		t.Fatalf("first or create rows affect err:%v rows:%d", res.Error, res.RowsAffected)
 	}
 }
+
+func TestCreateWithAutoIncrementCompositeKey(t *testing.T) {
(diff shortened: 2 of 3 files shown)


a631807682 (commented):
Can you add more tests to describe it? just like doing in playground.

tests/create_test.go
```diff
@@ -547,3 +547,33 @@ func TestFirstOrCreateRowsAffected(t *testing.T) {
...
+		t.Fatalf("failed to create, got error %v", err)
+	}
+
+	newProd := &CompositeKeyProduct{}
+	if err := DB.First(&newProd).Error; err != nil {
+		t.Fatalf("errors happened when query: %v", err)
+	} else {
+		AssertObjEqual(t, newProd, prod, "ProductID", "LanguageCode", "Code", "Name")
+	}
+}
```
truongns, line 579: @a631807682
I added this test here as per your request.
So with MySQL
Before the fix:
```
newProd.ProductID : 1
prod.ProductID : 0
```
After
```
newProd.ProductID : 1
prod.ProductID : 1
```

schema/schema.go
```diff
@@ -225,6 +225,16 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 		schema.PrioritizedPrimaryField = schema.PrimaryFields[0]
 	}
 
+	// If there are multiple primary keys, the AUTOINCREMENT field is prioritized
+	if schema.PrioritizedPrimaryField == nil && len(schema.PrimaryFields) > 1 {
+		for _, field := range schema.PrimaryFields {
+			if _, ok := field.TagSettings["AUTOINCREMENT"]; ok {
```
a631807682, line 231: Can we use `field.AutoIncrement`?  What happens when tag is set to `autoIncrement:false`?

tests/create_test.go
```diff
@@ -547,3 +547,33 @@ func TestFirstOrCreateRowsAffected(t *testing.T) {
...
+
+func TestCreateWithAutoIncrementCompositeKey(t *testing.T) {
+	type CompositeKeyProduct struct {
+		ProductID    int `gorm:"primaryKey;autoIncrement:true;"` // primary key
+		LanguageCode int `gorm:"primaryKey;"`                    // primary key
+		Code         string
+		Name         string
+	}
+
+	err := DB.Migrator().DropTable(&CompositeKeyProduct{})
```
a631807682, line 559: never used ?

schema/schema.go
a631807682: Can this code be merged with here?
https://github.com/go-gorm/gorm/pull/6127/files#diff-eafe62eae191e88faad4c223c4de9af268c175ee68695872e175b814d7c5baa5L224

truongns: @a631807682 Thank you for pointing this out, I have fixed the code based on your comment.
- use field.AutoIncrement instead
- merge with [#6127 (files)](https://github.com/go-gorm/gorm/pull/6127/files#diff-eafe62eae191e88faad4c223c4de9af268c175ee68695872e175b814d7c5baa5L224)
- add test for autoIncrement:false

tests/create_test.go
truongns: I have removed the unused code

a631807682 (approved):

tests/create_test.go
```diff
@@ -548,6 +548,35 @@ func TestFirstOrCreateRowsAffected(t *testing.T) {
...
+		LanguageCode: 56,
+		Code:         "Code56",
+		Name:         "ProductName56",
+	}
+	if err := DB.Create(&prod).Error; err != nil {
+		t.Fatalf("failed to create, got error %v", err)
+	}
+
+	newProd := &CompositeKeyProduct{}
+	if err := DB.First(&newProd).Error; err != nil {
```
black-06, line 573: This is not rigorous here. When I run the test repeatedly, the inserted record product id is self-increasing, so it will fail.
I suggest to do `DropTable` before `AutoMigrate`.

```golang
if err := DB.Migrator().DropTable(&CompositeKeyProduct{}); err != nil {
	t.Fatalf("failed to migrate, got error %v", err)
}
```

PR #6133, opened 2023-03-08
refactor: translate error only when it is not nil

saeidee:
- [x] Do only one thing
- [x] Tested

It will skip translating errors when thee error is nil.

### DIFF
--- a/gorm.go
+++ b/gorm.go
@@ -347,13 +347,13 @@ func (db *DB) Callback() *callbacks {
 
 // AddError add error to db
 func (db *DB) AddError(err error) error {
-	if errTranslator, ok := db.Dialector.(ErrorTranslator); ok {
-		err = errTranslator.Translate(err)
-	}
-
 	if db.Error == nil {
 		db.Error = err
 	} else if err != nil {
+		if errTranslator, ok := db.Dialector.(ErrorTranslator); ok {
+			err = errTranslator.Translate(err)
+		}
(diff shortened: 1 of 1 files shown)


gorm.go
```diff
@@ -347,8 +347,10 @@ func (db *DB) Callback() *callbacks {
 
 // AddError add error to db
 func (db *DB) AddError(err error) error {
-	if errTranslator, ok := db.Dialector.(ErrorTranslator); ok {
-		err = errTranslator.Translate(err)
+	if err != nil {
+		if errTranslator, ok := db.Dialector.(ErrorTranslator); ok {
+			err = errTranslator.Translate(err)
+		}
```
jinzhu, line 353: Move the below code to the if-check?

saeidee: Sorry, I didn't get it, do you mean adding it to the if & if else checks?

jinzhu: change it to something like:

```go
func (db *DB) AddError(err error) error {
	if err != nil {
		if errTranslator, ok := db.Dialector.(ErrorTranslator); ok {
			err = errTranslator.Translate(err)
		}

		if db.Error == nil {
			db.Error = err
		} else {
			db.Error = fmt.Errorf("%v; %w", db.Error, err)
		}
	}
	return db.Error
}
```

jinzhu (approved):

PR #6142, opened 2023-03-13
feat: add `When`

flc1125:
- [x] Do only one thing
- [x] Tested

I added a `When` to implement the logic when conditions exist

### User Case Description

```go
var age = 16

db.When(age > 18, func (db *gorm.DB) {
    return db.Where("show = ?", 1)
}).Where(...)

```

### DIFF
--- a/chainable_api.go
+++ b/chainable_api.go
@@ -205,6 +205,20 @@ func (db *DB) Where(query interface{}, args ...interface{}) (tx *DB) {
 	return
 }
 
+// When add conditions when condition is true
+//
+//	// Find the first user with name jinzhu
+//	db.When(true, func(db *gorm.DB) *gorm.DB {
+//		return db.Where("name = ?", "jinzhu")
+//	}).First(&user)
+func (db *DB) When(condition bool, fn func(*DB) *DB) *DB {
+	if condition {
+		return fn(db)
+	}
+
+	return db
+}
+
 // Not add NOT conditions
 //
 // Not works similarly to where, and has the same syntax.
--- a/tests/query_test.go
+++ b/tests/query_test.go
@@ -1380,3 +1380,40 @@ func TestQueryError(t *testing.T) {
 	}, Value: 1}).Scan(&p2).Error
 	AssertEqual(t, err, gorm.ErrModelValueRequired)
 }
+
+func TestWhen(t *testing.T) {
+	users := []User{
+		{Name: "when_1", Age: 10},
+		{Name: "when_2", Age: 20},
+		{Name: "when_3", Age: 30},
+		{Name: "when_3", Age: 40},
+	}
+
+	if err := DB.Create(&users).Error; err != nil {
+		t.Fatalf("errors happened when create users: %v", err)
+	}
+
+	t.Run("When True", func(t *testing.T) {
+		var trueUser User
+		if err := DB.Where("name LIKE ?", "when%").
+			When(true, func(db *gorm.DB) *gorm.DB {
+				return db.Where("age <= ?", 20)
+			}).Order("age Desc").First(&trueUser).Error; err != nil {
+			t.Errorf("Failed, got error: %v", err)
+		} else {
+			CheckUser(t, trueUser, users[1])
+		}
+	})
+
+	t.Run("When False", func(t *testing.T) {
+		var falseUser User
+		if err := DB.Where("name LIKE ?", "when%").
+			When(false, func(db *gorm.DB) *gorm.DB {
+				return db.Where("age <= ?", 20)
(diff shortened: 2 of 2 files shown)


a631807682 (commented):
Thanks for your PR, but it doesn't seem necessary, you can do it with `Scope`

flc1125:
> Thanks for your PR, but it doesn't seem necessary, you can do it with `Scope`

But when I'm associated with the business context, I need to add a lot of lines to solve this problem, in this way, I think it's more convenient.

flc1125:
> Thanks for your PR, but it doesn't seem necessary, you can do it with `Scope`

For Example:

```go
// When use `When`
DB.Where("age >= ?", 10).
    When(request.Get("name") != "", func(db *gorm.DB) *gorm.DB {
        return db.Where("name = ?", request.Get("name"))
    }).First(&user)

// When not use `When`
DB.Where("age >= ?", 10)

if request.Get("name") != "" {
    DB = DB.Where("name = ?", request.Get("name"))
}

DB.First(&user)
```

a631807682:
https://gorm.io/docs/scopes.html#content-inner
```go
func When(condition bool, fn func(*gorm.DB) *gorm.DB) func(*gorm.DB) *gorm.DB {
	if condition {
		return fn
	}

	return func(d *gorm.DB) *gorm.DB {
		return d
	}
}

DB.Scopes(When(age > 18, func(db *gorm.DB) *gorm.DB {
		return db.Where("show = ?", 1)
})).Where(...)
```

flc1125:
> I think there is no difference, and scope can be used in more scenarios. https://gorm.io/docs/scopes.html#content-inner
> 
> ```go
> func When(condition bool, fn func(*gorm.DB) *gorm.DB) func(*gorm.DB) *gorm.DB {
> 	if condition {
> 		return fn
> 	}
> 
> 	return func(d *gorm.DB) *gorm.DB {
> 		return d
> 	}
> }
> 
> DB.Scopes(When(age > 18, func(db *gorm.DB) *gorm.DB {
> 		return db.Where("show = ?", 1)
> })).Where(...)
> ```

The purpose of the framework is to make it easier for developers to use it, and some common and convenient method can be introduced into the framework so that other projects don't have to copy it every time.

And again, this is a function that a lot of people around you want to have.

Hopefully it will go through the merger.😁😁

a631807682:
Sorry, we must maintain the necessity of the API to ensure that it will not be unmaintainable due to the expansion of supported features.
Of course, we also welcome users to expand it, including drivers, plugins, clauses, tools or others, we will add links to good projects, and we will also evaluate and accept new projects into the go-gorm organization

flc1125:
> Sorry, we must maintain the necessity of the API to ensure that it will not be unmaintainable due to the expansion of supported features. Of course, we also welcome users to expand it, including drivers, plugins, clauses, tools or others, we will add links to good projects, and we will also evaluate and accept new projects into the go-gorm organization
> 
> https://gorm.io/community.html#Open-Sources

Okay, surprisingly, I think gorm has a long way to go in terms of ease of use.

jinzhu:
I think the `When` API is just a sugar method, and too many sugars are bad for project quality & maintainability.

So we should be very careful with any new APIs, and only add them if it helps a lot to make users write better code.


For this feature, I would suggest users extract business logic like https://gorm.io/docs/scopes.html#Pagination , for your application, it might be written like the following code:

```go
func Allowed(user User) func(*gorm.DB) *gorm.DB {
   if user.Age > 18 || user.Role == "admin" { 
       return func(tx *gorm.DB) *gorm.DB {
           return tx.Where("show = ?", 1)
      }
   }

   return func(tx *gorm.DB) *gorm.DB {
       return tx.Where("show = ?", 0)
   }
}

db.Where(Allowed(user)).Find(&articles)
```

flc1125:
> I think the `When` API is just a sugar method, and too many sugars are bad for project quality & maintainability.
> 
> So we should be very careful with any new APIs, and only add them if it helps a lot to make users write better code.
> 
> For this feature, I would suggest users extract business logic like https://gorm.io/docs/scopes.html#Pagination , for your application, it might be written like the following code:
> 
> ```go
> func Allowed(user User) func(*gorm.DB) *gorm.DB {
>    if user.Age > 18 || user.Role == "admin" { 
>        return func(tx *gorm.DB) *gorm.DB {
>            return tx.Where("show = ?", 1)
>       }
>    }
> 
>    return func(tx *gorm.DB) *gorm.DB {
>        return tx.Where("show = ?", 0)
>    }
> }
> 
> db.Scopes(Allowed(user)).Find(&articles)
> ```

Get, thanks for your time.

Perhaps `when` is better loaded in an extended form.

Issue #6136, opened 2023-03-10
Scan after using Scopes function panicks if *gorm.DB object is used for fetching other data.


peymanmortazavi:
## GORM Playground Link
https://github.com/go-gorm/playground/pull/577

## Description

We currently use the **Scopes** function, admittedly to my dislike, for some common joins or preparing long raw queries. The way we do it, is we first fetch some account attributes and then use them to return a raw query:

````
func ScopeFunc(id string) func (*gorm.DB) *gorm.DB {
   return func (d *gorm.DB) *gorm.DB {
        someModel := Model{}
        d.Raw("...").Scan(&someModel).Error
        // check error
        ...
       return d.Raw(preparedQuery)
  }
}
````
With the older version, this worked but after upgrading to 1.24.5, this panics.

It might be useful to note that I noticed in `finished_api.go` line 512 inside the `func (db *DB) Rows(*sql.Rows, error)`   we have this block:
```
	rows, ok := tx.Statement.Dest.(*sql.Rows)   <-- here we get *sql.Row instead of *sql.Rows
	if !ok && tx.DryRun && tx.Error == nil {
		tx.Error = ErrDryRunModeUnsupported
	}
``` 

We're not dry run, ok is false so nil result is returned with a nil error cause downstream methods to pass the error check and dereference a pointer that is pointing to nil.

**Comment by black-06** [comment]
use `Session` to copy a db when calling finisher_api in `Scopes`.
I don't know why, but it's okay...
```go
err = DB.Scopes(func(d *gorm.DB) *gorm.DB {
	a := Account{}
	if err := d.Session(&gorm.Session{}).Raw("select * from accounts limit 1").Scan(&a).Error; err != nil {
		return d
	}
	return d.Raw("select * from users limit 2")
}).Scan(&users).Error
```

peymanmortazavi:
@black-06 Yeah that definitely resolved the issue. My understand was that the chain API always takes a copy of the db instance. Though probably some pointers are used and those are just references.

My question is, should we always opt to use `Session` in similar situations or is this a bug that we should resolve in future.

a631807682:
When nested `Scan` is executed, the `rows` identifier will be deleted by mistake.
We may need to use an array to store the type of each query in the nested query.

https://github.com/go-gorm/gorm/blob/master/finisher_api.go#L512
https://github.com/go-gorm/gorm/blob/master/callbacks/row.go#L15

black-06:
[...]

PR #6143, opened 2023-03-13
fix: nested scan (#6136)

black-06:
- [X] Do only one thing
- [X] Tested

fix nested scan panic #6136 

### User Case Description

```go
var user User
if err := DB.Scopes(func(db *gorm.DB) *gorm.DB {
	var maxID int64
	if err := db.Raw("select max(id) from users").Scan(&maxID).Error; err != nil {
		return db
	}
	return db.Raw("select * from users where id = ?", maxID)
}).Scan(&user).Error; err != nil {
	t.Errorf("failed to find user, got err: %v", err)
}
```

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -35,7 +35,6 @@ type Statement struct {
 	Omits                []string // omit columns
 	Joins                []join
 	Preloads             map[string][]interface{}
-	QueryTypes           QueryTypes
 	Settings             sync.Map
 	ConnPool             ConnPool
 	Schema               *schema.Schema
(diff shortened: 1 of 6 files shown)


statement.go
```diff
@@ -46,6 +46,7 @@ type Statement struct {
 	attrs                []interface{}
 	assigns              []interface{}
 	scopes               []func(*DB) *DB
+	queryTypes           []bool
```
a631807682, line 49: Can we store it in `Settings` ? I haven't tested it, but could something wrong with this ?
```go
tx := db.Scope(...)
go func() {
	tx.Scan(...)
}()
go func() {
	tx.Scan(...)
}()
```

black-06: It has the same problem unless we use thread-safe slice/array

a631807682: > It has the same problem unless we use thread-safe slice/array

Or can we use lock?

a631807682 (approved):
This is ok for me, but I think since we implemented a list, it would be a bit wasteful if we can only deal with bools. Of course, I don't think of other scenarios that need to be used at the moment, so I think it's ok.

jinzhu:
If this PR is only used to fix the panic issue in the nested scan, I think don't panic but return an error should be enough.

We don't expect users to use it like this, for more details: https://gorm.io/docs/method_chaining.html

Going this close this PR for now, feel free to create a new one if those changes are required in other scenarios.

Thank you.

Issue #6139, opened 2023-03-10
Save (probably) should be idempotent


haroldbouley:
According to https://gorm.io/docs/update.html#Save-All-Fields

"Save is a combination function. If save value does not contain primary key, it will execute Create, otherwise it will execute Update (with all fields)."
The library is working as described in the quote, strictly speaking. But Save is not idempotent, and I may obviously be wrong but I don't think it is intuitive or convenient.
According to this code snippet at the end of the Save function : 

`updateTx := tx.callbacks.Update().Execute(tx.Session(&Session{Initialized: true}))

if updateTx.Error == nil && updateTx.RowsAffected == 0 && !updateTx.DryRun && !selectedUpdate {
	return tx.Create(value)
}`

If I call the Save function twice with the same object (assuming this object has an ID), the second time it is called no rows will be updated (because all fields keep the same value), thus the Create function will be called and likely provoke an error.

Obviously, I could append a pull request to the issue, simply deleting the if and its content, which should make the Save function idempotent, but I guess this code has a reason to be there. In case everything is working as intended, I fail to see a simple, clean and convenient way to perform a "if primary key is set, update else create". I looked into FirstOrCreate with Assign but I don't understand how to use it in a generic way. Is there something I am missing?

Regards

polly3d:
I am experiencing the same issue. 
Just added a 'created_at' field now.

Here is the same issue:
[6134](https://github.com/go-gorm/gorm/issues/6134)

Issue #6134, opened 2023-03-09
If the created_at field is not added, the db.Save() function will continue to perform insertion operations.


polly3d:
## GORM Playground Link

This is a fake URL to prevent this bug from being overlooked: https://github.com/go-gorm/playground/pull/1

## Description

The data model:

```go
type User struct {
	ID           uint
	Name         string
}
```

Here are the steps to reproduce：

```go
// Create first record
db.Create(&User{})

// Fetch the first record
var user *User
db.First(&user)

// Save record
user.Name = "jinzhu 2"
db.Save(&user)

// Save again with the same data
// It will throw an error: Error 1062 (23000): Duplicate entry '1' for key 'users.PRIMARY'
user.Name = "jinzhu 2"
db.Save(&user)

```

PR #6149, opened 2023-03-14
save should be idempotent

black-06:
- [X] Do only one thing
- [X] Tested

Save should be idempotent. #6139 #6134 

When call `Save` twice with the same object, `Save` should not raise an error.

### User Case Description

```golang
var company Company
DB.Frist(&company)

DB.Save(&company)
DB.Save(&company) // should not create new record or rasise an error.
DB.Save(&company) // should not create new record or rasise an error.
```


a631807682 (approved):

monoflash:
MySQL MariaDB database, same problem.
Save() does not work from version 1.24.6 to latest.
The latest version in which Save() works is v1.24.5.

black-06:
> I stopped using save as they keep having race data problems. The save should only update the column that has been changed.

Not, See [Save doc](https://gorm.io/docs/update.html#Save-All-Fields)
> `Save` will save all fields when performing the Updating SQL

Issue #6148, opened 2023-03-13
Error in generation groups queries in scopes


daashuun:
## GORM Playground Link

https://github.com/daashuun/playground

## Description

I tried to make
`SELECT * FROM "test" WHERE a = 1 AND (b = 2 OR c = 3)`

but this request
```
DB.Table("test").Scopes(
  func(d *gorm.DB) *gorm.DB {
	  return d.Where("a = 1")
  },
  func(d *gorm.DB) *gorm.DB {
	  return d.Where(d.Or("b = 2").Or("c = 3"))
  },
).Rows()
```
makes sql
`SELECT * FROM "test" WHERE a = 1 OR b = 2 OR c = 3 AND (a = 1 OR b = 2 OR c = 3)`

black-06:
It needs to be repaired and you can use this instead:
```go
	db.Table("test").Scopes(
		func(d *gorm.DB) *gorm.DB {
			return d.Where("a = 1")
		},
		func(d *gorm.DB) *gorm.DB {
			return d.Where(db.Or("b = 2").Or("c = 3")) // it's db
		},
	).Rows()
```

a631807682:
@black-06  This is a bug? Can it be assigned to you?

black-06:
> @black-06 This is a bug? Can it be assigned to you?

I think so, it can be assigned to me.
But it may require many changes to `Scopes`?  Do you have any idea

a631807682:
> But it may require many changes to `Scopes`? Do you have any idea

Sorry, I only read your comment and didn't check the question, I'll check and get back to you.

a631807682:
@black-06 This does not seem to be a scope problem, but a subquery processing problem. which will reuse Exprs to build a clause, even if it has been used externally.
https://github.com/go-gorm/gorm/blob/master/statement.go#L331
```go
	tx := DB.Where("a = 1").Session(&gorm.Session{})
	tx.Where(tx.Or("b = 2").Or("c = 3")).First(&test)
	//  SELECT * FROM `tests` WHERE a = 1 AND (a = 1 OR b = 2 OR c = 3) AND `tests`.`deleted_at` IS NULL ORDER BY `tests`.`id` LIMIT 1

```

black-06:
> @black-06 This does not seem to be a scope problem, but a subquery processing problem. which will reuse Exprs to build a clause, even if it has been used externally.
> https://github.com/go-gorm/gorm/blob/master/statement.go#L331
> ```go
[...]

PR #6152, opened 2023-03-15
fix cond in scopes

black-06:
- [X] Do only one thing
- [X] Tested

fix conditions in `Scope` #6148 

### User Case Description

```go
DB.Table("test").Scopes(
	func(d *gorm.DB) *gorm.DB {
		return d.Where("a = 1")
	},
	func(d *gorm.DB) *gorm.DB {
		return d.Where(d.Or("b = 2").Or("c = 3"))
	},
).Rows()
```
It should be:
```sql
SELECT * FROM `test` WHERE a = 1 AND (b = 2 OR c = 3)
```
But now it is:
```sql
SELECT * FROM `test` WHERE a = 1 OR b = 2 OR c = 3 AND (a = 1 OR b = 2 OR c = 3)
```

### DIFF
--- a/chainable_api.go
+++ b/chainable_api.go
@@ -366,7 +366,7 @@ func (db *DB) Scopes(funcs ...func(*DB) *DB) (tx *DB) {
 	return tx
 }
 
-func (db *DB) executeScopes(keepScopes bool) (tx *DB) {
+func (db *DB) executeScopes() (tx *DB) {
 	tx = db.getInstance()
 	scopes := db.Statement.scopes
 	if len(scopes) == 0 {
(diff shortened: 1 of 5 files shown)


chainable_api.go
```diff
@@ -366,6 +366,39 @@ func (db *DB) Scopes(funcs ...func(*DB) *DB) (tx *DB) {
 	return tx
 }
 
+func (db *DB) executeScopes(keepScopes bool) (tx *DB) {
+	tx = db.getInstance()
+	scopes := db.Statement.scopes
+	if len(scopes) == 0 {
```
jinzhu, line 372: in a scope condition, it might have other scope conditions, so we are writing the code like this:

```go
for len(db.Statement.scopes) > 0 {
	db = db.executeScopes(false)
		scopes := db.Statement.scopes
		db.Statement.scopes = nil
		for _, scope := range scopes {
			db = scope(db)
		}
	}
```

Seems like the change will break it.

black-06: Sorry, it's my fault.

a631807682 (approved):

PR #6167, opened 2023-03-20
feat: Implemented "easer" to reduce the DB load

ktsivkov:
- [x] Do only one thing
- [x] Tested

Implemented a feature to reduce Select queries to the database with a new bool option `Easer` that can be provided to the gorm when initializing.

### User Case Description

Imagine that you are having a system that handles multiple but identical Select queries.
Now let's say a query takes 500ms to execute, and during that time 30 more users have visited your app you will need to execute the exact same query 30 times and a total of 15 seconds of execution time.
This is wasteful! So what can be done, is to group them, and execute the query only once, and once the result of that query is ready, we distribute it to all of the users.

- User A, User B, and user C execute a query that runs 3 seconds.
- User A has requested it first
- User B & C request it to be executed, but since the query is still running they get subscribed to the result of it
- Once the query is done, the result of it is duplicated to all subscribed parties.


jinzhu:
Hi @ktsivkov 

I think it's better to implement this feature as a cache plugin?

Going to close this PR for now, let me know if you want to create the cache plugin, and I can create a repo under gorm org for you.

ktsivkov:
Hey @jinzhu, sure, it actually makes much more sense if it is done as a plugin. Would like to do that, so please create a repo for it, and I will try to do it this week!

jinzhu:
> Hey @jinzhu, sure, it actually makes much more sense if it is done as a plugin. Would like to do that, so please create a repo for it, and I will try to do it this week!

invited. https://github.com/go-gorm/caches/invitations

ktsivkov:
@jinzhu Today I released the first version of the plugin, are there any next steps I need to take in order to publish it?
I have also used the github.com/go-gorm/caches/invitations link in the go.mod, since the one under gorm.io/caches is not existing

PS: Did not add github actions in order not to consume any cpu time from your org. If needed just let me know, and I will setup actions for fmt & tests.

Issue #6181, opened 2023-03-24
`PreparedStmtDB` does not implement `Ping()`


phenixrizen:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/585

## Description

`PreparedStmtDB` does not implement `Ping()` to satisfy the interface.  This is allowing bad connection strings or a non reachable DB  to never be caught when using prepared statements and calling `Open()` because `Ping()` will never be called.  `Open()` returns a `*gomr.DB` and a `nil error`.

PR #6182, opened 2023-03-24
Add ping method to satisy interface

phenixrizen:
- [x] Do only one thing
- [x] Tested

There is a issue when using Prepared Statments Caching and opening a DB.

`PreparedStmtDB` does not implement `Ping()` to satisfy the interface. This is allowing bad connection strings or a non reachable DB to never be caught when using prepared statements and calling `Open()` because `Ping()` will never be called. `Open()` returns a `*gomr.DB` and a `nil error`.

This implements the `Ping()` method on the `gorm.PreparedStmtDB` allowing us to catch connection failures. The way gorm checks interfaces is not the best in my opinion.  i.e. https://github.com/phenixrizen/gorm/blob/master/gorm.go#L203-L207

Fixes https://github.com/go-gorm/gorm/issues/6181

### User Case Description

I want to ping the db when using the  Prepared Statments Caching.

### DIFF
--- a/prepare_stmt.go
+++ b/prepare_stmt.go
@@ -32,6 +32,15 @@ func (db *PreparedStmtDB) GetDBConn() (*sql.DB, error) {
 	return nil, ErrInvalidDB
 }
 
+func (db *PreparedStmtDB) Ping() error {
+	sqldb, err := db.GetDBConn()
+	if err != nil {
+		return err
+	}
+
+	return sqldb.Ping()
+}
+
 func (db *PreparedStmtDB) Close() {
 	db.Mux.Lock()
 	defer db.Mux.Unlock()


prepare_stmt.go
```diff
@@ -32,6 +32,15 @@ func (db *PreparedStmtDB) GetDBConn() (*sql.DB, error) {
 	return nil, ErrInvalidDB
 }
 
+func (db *PreparedStmtDB) Ping() error {
+	sqldb, err := db.GetDBConn()
```
a631807682, line 36: `GetDBConn` will return error if the underlying database connection is not a *sql.DB or does not implement `GetDBConnector`. We need to avoid this error because it is not a ping error but the driver does not implement the ping interface.

phenixrizen: You need to check this error otherwise if we can't get the connection we can't ping. Therefore it is a ping error. If we don't check this error open will return a DB and a nil error.

phenixrizen: @jinzhu What are your thoughts on this PR?

jinzhu:
Thank you for your PR, I would prefer to just leave users to use the generic interface, so we don't need to copy each method to GORM.

https://gorm.io/docs/generic_interface.html

PR #6189, opened 2023-03-27
feat: Added DSN

flc1125:
- [x] Do only one thing
- [x] Tested

Added a `DSN` tool to facilitate `DSN` character concatenation

### User Case Description

```go
dsn := gorm.DSN{
	Host: "127.0.0.1",
	Port: 3306,
	User: "root",
	Pass: "password",
	Db:   "gorm",
	Options: map[string]string{
		"charset":   "utf8mb4",
		"parseTime": "True",
		"loc":       "Local",
	},
}

gorm.Open(mysql.Open(dsn.String()), &gorm.Config{})
```

### DIFF
--- a/dsn.go
+++ b/dsn.go
@@ -1,0 +1,31 @@
+package gorm
+
+import (
+	"net/url"
+	"strconv"
+)
+
+type DSN struct {
+	Host    string
+	Port    int
+	User    string
+	Pass    string
+	Db      string
+	Options map[string]string
+}
+
+func (d DSN) String() string {
+	dsn := d.User + ":" + d.Pass + "@tcp(" + d.Host + ":" + strconv.Itoa(d.Port) + ")/" + d.Db
+
+	if d.Options != nil {
+		value := url.Values{}
+
+		for k, v := range d.Options {
+			value.Add(k, v)
+		}
+
+		dsn += "?" + value.Encode()
+	}
+
+	return dsn
+}
--- a/dsn_test.go
+++ b/dsn_test.go
@@ -1,0 +1,30 @@
+package gorm
+
+import "testing"
+
+func TestDsn(t *testing.T) {
+	dsn := DSN{
+		Host: "127.0.0.1",
+		Port: 3306,
+		User: "root",
+		Pass: "password",
+		Db:   "gorm",
+	}
+
+	t.Run("dsn string", func(t *testing.T) {
+		if dsn.String() != "root:password@tcp(127.0.0.1:3306)/gorm" {
+			t.Error("dsn string error")
+		}
+	})
+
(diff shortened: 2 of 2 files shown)


dsn.go
```diff
@@ -0,0 +1,31 @@
...
+	Host    string
+	Port    int
+	User    string
+	Pass    string
+	Db      string
+	Options map[string]string
+}
+
+func (d DSN) String() string {
+	dsn := d.User + ":" + d.Pass + "@tcp(" + d.Host + ":" + strconv.Itoa(d.Port) + ")/" + d.Db
```
demoManito, line 18: need to check whether the parameter is empty?

flc1125: I have tried, and if dsn is not valid, it will panic at mysql.Open.

There are no syntax problems because of the default values.

If you want to add a empty check, then you have to return error, so it doesn't iterate.

**Therefore, I recommend not adding empty check here.**

dsn.go
```diff
@@ -0,0 +1,31 @@
...
+
+	if d.Options != nil {
```
demoManito, line 20: Here should judge `len(d.Options) != 0`

dsn.go
```diff
@@ -0,0 +1,31 @@
...
+	"net/url"
+	"strconv"
+)
+
+type DSN struct {
+	Host    string
+	Port    int
+	User    string
+	Pass    string
+	Db      string
```
demoManito, line 13: I think it would be better to call this `DB`

dsn.go
flc1125: Fixed

dsn.go
flc1125: Fixed

dsn.go
```diff
@@ -0,0 +1,31 @@
...
+	User    string
+	Pass    string
+	DB      string
+	Options map[string]string
+}
+
+func (d DSN) String() string {
+	dsn := d.User + ":" + d.Pass + "@tcp(" + d.Host + ":" + strconv.Itoa(d.Port) + ")/" + d.DB
+
+	if d.Options != nil && len(d.Options) > 0 {
```
demoManito, line 20: `if len(d.Options) > 0 {` 即可

demoManito:
Should be added to the: https://github.com/go-gorm/mysql

demoManito:
I don't think it's necessary to add this feature

https://github.com/go-sql-driver/mysql (https://pkg.go.dev/github.com/go-sql-driver/mysql?utm_source=godoc#Config.FormatDSN)

a631807682 (commented):
Different database/drivers have different rules for parsing DataSourceName.

Issue #6098, opened 2023-02-23
Postgres Max Identifier length is 63 bytes


ghostsquad:
## GORM Playground Link

(no playground link)

## Description

Postgres has a max identifier length of 63 bytes. I see a test here that's asserting for 64 characters/bytes.

https://github.com/go-gorm/gorm/blob/master/schema/naming_test.go#L196

https://til.hashrocket.com/posts/8f87c65a0a-postgresqls-max-identifier-length-is-63-bytes

https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

ghostsquad:
Not stale. I don't think that a playground PR is strictly necessary here. Links to the documentation and current tested expectations have been mentioned.

ghostsquad:
Insistent little bot...

a631807682:
Since different databases support different lengths, we can add configuration items to schema.NamingStrategy and change the default configuration of Postgres drivers.
Contributions welcome, let me know if anyone is interested.

alidevhere:
> Since different databases support different lengths, we can add configuration items to `schema.NamingStrategy` and change the default configuration of Postgres driver. Contributions welcome, let me know if anyone is interested.

Hii !!
Please assign this issue to me, i would love to work on it.

alidevhere:
@a631807682 
Sorry if i sound like a fool, but this is my first contribution in this repo. And i still don't get how can we do that?
Let me explain my self what i mean

You asked to add configuration item in ```schema.NamingStrategy```
So i added ``` NamingStrategyConfig ``` [see here](https://github.com/go-gorm/gorm/pull/6196#issuecomment-1487024055) :

```
// NamingStrategy tables, columns naming strategy
type NamingStrategy struct {
	TablePrefix   string
	SingularTable bool
	NameReplacer  Replacer
	NoLowerCase   bool
	NamingStrategyConfig
}

// This struct is used to configure the behavior NamingStrategy, according to DB.
// For example, in MySQL, the maximum length of an identifier is 64 characters.
// In PostgreSQL, the maximum length of an identifier is 63 characters.
// In SQL Server, the maximum length of an identifier is 128 characters.
// In SQLite, the maximum length of an identifier is unlimited.
// In future, we may add more options to NamingStrategyConfig.
type NamingStrategyConfig struct {
	IdentifierMaxLength int
}

```

[...]

PR #6196, opened 2023-03-28
issue #6098

alidevhere:
- [*] Do only one thing
- [*] Tested

It solves the issue [Postgres Max Identifier length is 63 bytes](https://github.com/go-gorm/gorm/issues/6098) 

NamingStrategyConfig struct is added in schema.NamingStrategy, for supporting Max Identifier Lengths in different databases
Identifier Length support is added for Mysql(64),MS Sql Server ( 128),Sqlite (it has unlimited but default 64 is used), Postgres (63)
### User Case Description

All test cases are updated. All tests have passed

### DIFF
--- a/gorm.go
+++ b/gorm.go
@@ -144,7 +144,20 @@ func Open(dialector Dialector, opts ...Option) (db *DB, err error) {
 	}
 
 	if config.NamingStrategy == nil {
-		config.NamingStrategy = schema.NamingStrategy{}
+		// Set default value of IdentifierMaxLength according to the database type
+		identifierMaxLength := 64
+		switch dialector.Name() {
+		case "mysql":
+			identifierMaxLength = 64
+		case "postgres":
+			identifierMaxLength = 63
+		case "sqlite":
+			identifierMaxLength = 64
+		case "sqlserver":
+			identifierMaxLength = 128
+		}
+
+		config.NamingStrategy = schema.NamingStrategy{NamingStrategyConfig: schema.NamingStrategyConfig{IdentifierMaxLength: identifierMaxLength}}
 	}
 
 	if config.Logger == nil {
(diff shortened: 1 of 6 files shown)


a631807682 (commented):
Similar to the reasons in the comments, we do not recommend judging the dialector name in gorm.
https://github.com/go-gorm/gorm/issues/6098#issuecomment-1454740584
https://github.com/go-gorm/gorm/pull/6004#issuecomment-1435446332

alidevhere:
> gorm

Then how we can know which length  to assign ?

a631807682:
> > gorm
> 
> Then how we can know which length to assign ? @a631807682 Need you help. Can you elaborate please.
> 
> thanks

We can use `Apply` to change some config in every drivers. 

https://github.com/go-gorm/mysql/blob/master/mysql.go#L85
https://github.com/go-gorm/gorm/blob/master/gorm.go#L143

schema/naming.go
```diff
@@ -32,6 +32,17 @@ type NamingStrategy struct {
...
+}
+
+// This struct is used to configure the behavior NamingStrategy, according to DB.
+// For example, in MySQL, the maximum length of an identifier is 64 characters.
+// In PostgreSQL, the maximum length of an identifier is 63 characters.
+// In SQL Server, the maximum length of an identifier is 128 characters.
+// In SQLite, the maximum length of an identifier is unlimited.
+// In future, we may add more options to NamingStrategyConfig.
+type NamingStrategyConfig struct {
+	IdentifierMaxLength int
```
jinzhu, line 45: Don't need the `NamingStrategyConfig` struct? we can just move the `IdentifierMaxLength` field to `NamingStrategy`

gorm.go
```diff
@@ -144,7 +144,20 @@ func Open(dialector Dialector, opts ...Option) (db *DB, err error) {
 	}
 
 	if config.NamingStrategy == nil {
-		config.NamingStrategy = schema.NamingStrategy{}
+		// Set default value of IdentifierMaxLength according to the database type
```
jinzhu, line 147: those logic should move to each driver.

Issue #6209, opened 2023-04-04
Automatically analyze the number of batches in `CreateInBatches`


ken2403:
## Describe the feature

I would like the number of batches in `CreateInBatch` to be automatically set appropriately based on the size of the array to Insert and the number of query parameters.
Or, implement a new method to do so.

## Motivation

When operating a service using SQLServer, an event has occurred in which Create cannot be performed on huge arrays because the upper limit of parameters is 2100 ([ref](https://learn.microsoft.com/en-us/sql/relational-databases/stored-procedures/specify-parameters?view=sql-server-ver16)).
Since the number of structure fields in an array is dynamic, the number of parameters must be identified, e.g., by parsing SQL, and the number of batches must be determined appropriately for the size of the array to fit within the 2100 parameter limit.

PR #6210, opened 2023-04-04
add `CreateInAutoBatches` method

ken2403:
- [x] Do only one thing
- [x] Tested

API implemented for this [ issue](https://github.com/go-gorm/gorm/issues/6209).

### User Case Description

A method to execute `Create` with appropriate batch size depending on the number of fields of the struct and array size when the parameter limit is restricted.

```go
// big array
values := [10000]{User{Name: "Smith", Age: 33, From: "Canada"}, ...}
// For SQLServer, the maximum number of parameters is 2100
err := db.CreateInAutoBatches(values, 2100).Error
```

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"math"
 	"reflect"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


a631807682 (commented):
Similar to https://github.com/go-gorm/gorm/pull/6142#issuecomment-1465742524 ,this feat can be implemented by users in their own utils.
In fact, `NumField` does not represent the number of parameters, gorm also supports `embedded` and `field permissions`

ken2403:
Thanks for your comment.
I agree that this PR is overengineering, as I have tried many things and it seems that I can implement something equivalent to this by making full use of Gorm's API.

I will continue to support Gorm's simple but powerful API!
Thanks for the kind review and thank you for your time.

By the way, could you please provide a reference to `embedded` and `field permissions`?

a631807682:
`embedded`
```go
 type Author struct {
  Name  string
  Email string
}

type Blog struct {
  ID      int
  Author  Author `gorm:"embedded"`
  Upvotes int32
}
```
https://gorm.io/docs/models.html#Embedded-Struct
https://github.com/go-gorm/gorm/blob/master/schema/field.go#L83
`field permissions`
```go
	type CustomizeFieldStruct struct {
		gorm.Model
		Name                    string
		FieldAllowCreate        string `gorm:"<-:create"`
		FieldAllowUpdate        string `gorm:"<-:update"`
		FieldAllowSave          string `gorm:"<-"`
		FieldAllowSave2         string `gorm:"<-:create,update"`
		FieldAllowSave3         string `gorm:"->:false;<-:create"`
		FieldReadonly           string `gorm:"->"`
		FieldIgnore             string `gorm:"-"`
	}
```
https://gorm.io/docs/models.html#Fields-Tags
https://github.com/go-gorm/gorm/blob/master/schema/field.go#L62

PR #6220, opened 2023-04-07
fix(nested transaction): SavePoint SQL Statement not support in Prepared Statements

wangliuyang520:
- [x] Do only one thing
- [x] Tested

fix(nested transaction): SavePoint SQL Statement not support in Prepared Statements

### User Case Description
Nested transactions are implemented through savepoint, but savepoint does not support precompiled statements, and if the session is enabled for precompilation, nested transactions will fail

### DIFF
--- a/tests/transaction_test.go
+++ b/tests/transaction_test.go
@@ -57,6 +57,19 @@ func TestTransaction(t *testing.T) {
 	if err := DB.First(&User{}, "name = ?", "transaction-2").Error; err != nil {
 		t.Fatalf("Should be able to find committed record, but got %v", err)
 	}
+
+	t.Run("this is test nested transaction and prepareStmt coexist case", func(t *testing.T) {
+		// enable prepare statement
+		tx3 := DB.Session(&gorm.Session{PrepareStmt: true})
+		if err := tx3.Transaction(func(tx4 *gorm.DB) error {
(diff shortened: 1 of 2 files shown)


tests/transaction_test.go
```diff
@@ -57,6 +57,20 @@ func TestTransaction(t *testing.T) {
...
+		tx3 := DB.Session(&gorm.Session{PrepareStmt: true})
+		if err := tx3.Transaction(func(tx4 *gorm.DB) error {
+			// nested transaction
+			return tx4.Transaction(func(tx5 *gorm.DB) error {
+				return tx5.First(&User{}, "name = ?", "transaction-2").Error
+			})
+		}); err != nil {
+			t.Fatalf("prepare statement and nested transcation coexist" + err.Error())
+		}
+		println("ok")
```
chiyutianyi, line 72: 无需输出

a631807682 (commented):
Please fix the failing test.
related to https://github.com/go-gorm/gorm/pull/6060
seems https://github.com/go-gorm/gorm/commit/cc2d46e5be425300e064a39868cfdb333f24e4ac only solves the memory overflow, but does not solve the prepare nested transation in mysql.

jinzhu (approved):

PR #6227, opened 2023-04-11
feat: on soft delete, update additional fields from model/struct in the same update operation by using "updateOnSoftDelete" field tag

DrOctavius:
Added new feature - on soft delete, update additional fields from model/struct in the same update operation by using "updateOnSoftDelete" field tag

### User Case Description

example:
```
const postgresDSN = "user=root password= dbname=gorm host=localhost port=26257 sslmode=disable TimeZone=Europe/Chisinau"

type User struct {
	ID uuid.UUID `gorm:"primaryKey;type:uuid;not null;->;default:gen_random_uuid()"`

	DeletedByID *uuid.UUID `gorm:"type:uuid;updateOnSoftDelete"`
	DeletedAt   *gorm.DeletedAt
}

func main() {
	db, err := gorm.Open(postgres.Open(postgresDSN), &gorm.Config{})
	if err != nil {
		log.Fatalln(err)
	}

	var id = "60ea2d8f-4fac-42ba-b363-4536f6a8edf4"

	// Retrieve the user
	var user User
	result := db.First(&user, "id = ?", id)
	if result.Error != nil {
		log.Fatalln(result.Error)
	}

	// Set the user who will delete the record
	deletedByID := uuid.New()
	user.DeletedByID = &deletedByID

	// Delete
	result = db.Delete(&user)
	if result.Error != nil {
		log.Fatalln(result.Error)
	}
}
```

### DIFF
--- a/schema/field.go
+++ b/schema/field.go
@@ -62,6 +62,7 @@ type Field struct {
 	Creatable              bool
 	Updatable              bool
 	Readable               bool
+	UpdateOnSoftDelete     bool
 	AutoCreateTime         TimeType
 	AutoUpdateTime         TimeType
 	HasDefaultValue        bool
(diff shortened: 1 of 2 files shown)


a631807682 (commented):
Thank you for your PR, but whether it is `SoftDeleteUpdateClause` or `DeletedAt`, it is just a type, which is independent of `schema.Field`. This seems to support it by defining the `CustomDeletedAt` type.

PR #6230, opened 2023-04-12
feat: SQLite3 uses memory to speed up CI

wangliuyang520:
Change-Id: I2f1b7cb4830e3240d461c11b26a9c66e413d3d42

- [*] Do only one thing
- [*] Tested

feat: SQLite3 uses memory to speed up CI

### User Case Description
feat: SQLite3 uses memory to speed up CI


jinzhu:
AFAIK, memory mode is not exactly the same as normal mode, maybe better to test with production settings.

So going to close this PR.

PR #6276, opened 2023-04-28
omit() support called in chain

Lxxxxt:
- [✅] Do only one thing
- [✅] Non breaking API changes
- [✅] Tested

Usually, function omit() is used in a huge database table to improve performance. We got an instinct that gorm support method chaining like where() clause, but omit() does not. When omit() was used, the next call would rewrite the content of the last one. It is dangerous, especially while updating, which could cause unexpected inserts. It is easy to make a fix to support method chaining calls as while as forward compatibility. Just use append().

### User Case Description
Table Doc has association tables "Versions" and "Team". When I tried to update a "Doc" record, gorm would help upsert Versions and Team. To avoid it, I use the function omit() like the following:
image
Omit ("Versions") did overwrite by omit("Team"), which caused thousands of upsert request occurs every time I clicked the button on my site, and finally, my database broke down.
There was not any warning message or tips telling me it may be a wrong usage.

### DIFF
--- a/chainable_api.go
+++ b/chainable_api.go
@@ -178,9 +178,9 @@ func (db *DB) Omit(columns ...string) (tx *DB) {
 	tx = db.getInstance()
 
 	if len(columns) == 1 && strings.ContainsRune(columns[0], ',') {
-		tx.Statement.Omits = strings.FieldsFunc(columns[0], utils.IsValidDBNameChar)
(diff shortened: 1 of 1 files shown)


a631807682:
Thank you for your contribution, currently `Select` and `Omit` behaviors will be overridden, this is a breaking change with a relatively large impact, we cannot accept it in this version. This may be accepted in the next major version, but not now.

PR #6291, opened 2023-05-02
feat: add handling for the interface field, when an Interface field is an object that has already been instantiated, use the type of the instantiated object itself instead of interface{}.

jinlongchen:
feat: add handling for the interface field, when an Interface field is an object that has already been instantiated, use the type of the instantiated object itself instead of interface{}.

```go
feat: add handling for the interface field, when an Interface field is an object that has already been instantiated, use the type of the instantiated object itself instead of interface{}.

feat: add handling for the interface field, when an Interface field is an object that has already been instantiated, use the type of the instantiated object itself instead of interface{}.

type BaseMeta interface {
    // Scan....
}
type Meta1 struct {
}
// meta1 Scan ...
type Meta2 struct {
}
// meta2 Scan ...
type Hello struct {
    Meta BaseMeta `gorm:"type:JSON;column:meta"`
}

// ...
// hello := &Hello{ Meta: &Meta1{} }
// db.First(hello).Error...
```


jinzhu:
Hi @jinlongchen

Can you add more information about what's it used for?

jinlongchen:
sometimes, we need to store different types of data in a JSON field. however, using a `json.RawMessage` in code can be inconvenient. to address this issue, we define multiple types of structs which inherit from a base interface. we then use this base interface as the field type, enabling us to easily work with these structs via a type assertion without having to unmarshal the `json.RawMessage`.

jinzhu:
Checkout gorm.io/docs/serializer.html, it should work for your cases.

PR #6294, opened 2023-05-05
fix: save with hook (#6285)

### LINKED ISSUE #6285: BeforeSave is executed twice in transaction
## GORM Playground Link

https://github.com/go-gorm/playground/pull/593

## Description

When you save a model which `BeforeSave` is implemented in a transaction, `BeforeSave` will be executed twice.

```go
package main_test

import (
	"testing"

	"github.com/stretchr/testify/assert"
	"gorm.io/driver/mysql"
	"gorm.io/gorm"
)

func Test_main(t *testing.T) {
	dsn := "root:pass@tcp(localhost:3306)/dbname?charset=utf8mb4&parseTime=True&loc=Local"
	db, err := gorm.Open(mysql.Open(dsn))
	assert.NoError(t, err)
	assert.NoError(t, db.AutoMigrate(&User{}, &Token{}))

	u := User{
		Name: "user",
		Token: Token{
			Content: "token",
		},
	}
	u1, err := saveUser(db, &u)
	assert.NoError(t, err)
	assert.Equal(t, "token_encrypted", u1.Token.Content)

	u = User{
		ID:   u.ID,
		Name: "user",
		Token: Token{
			Content: "token2",
		},
	}
	u2, err := saveUser(db, &u)
	assert.NoError(t, err)
	assert.Equal(t, "token2_encrypted", u2.Token.Content) // FAIL: actual is token2_encrypted_encrypted
}

func saveUser(db *gorm.DB, u *User) (*User, error) {
	var newUser User
	if err := db.Transaction(func(tx *gorm.DB) error {
		if err := tx.Debug().Session(&gorm.Session{FullSaveAssociations: true}).Save(u).Error; err != nil {
			return err
		}

		if err := tx.Preload("Token").First(&newUser, u.ID).Error; err != nil {
			return err
		}

		return nil
	}); err != nil {
		return nil, err
	}

	return &newUser, nil
}

// - Models

type User struct {
	ID    int    `gorm:"primary_key"`
	Name  string `gorm:"type:varchar(100)"`
	Token Token  `gorm:"foreignKey:UserID"`
}

type Token struct {
	UserID  int    `gorm:"primary_key"`
	Content string `gorm:"type:varchar(100)"`
}

// This method should be called only once
func (t *Token) BeforeSave(tx *gorm.DB) error {
	t.Content += "_encrypted"
	return nil
}
```

```bash
$ go test main_test.go

2023/05/01 11:00:35 /home/ras/misc/go/gorm/main_test.go:42
[2.085ms] [rows:1] INSERT INTO `tokens` (`content`,`user_id`) VALUES ('token_encrypted',7) ON DUPLICATE KEY UPDATE `content`=VALUES(`content`)

2023/05/01 11:00:35 /home/ras/misc/go/gorm/main_test.go:42
[5.084ms] [rows:1] INSERT INTO `users` (`name`) VALUES ('user')

2023/05/01 11:00:35 /home/ras/misc/go/gorm/main_test.go:42
[2.312ms] [rows:2] INSERT INTO `tokens` (`content`,`user_id`) VALUES ('token2_encrypted',7) ON DUPLICATE KEY UPDATE `content`=VALUES(`content`)

2023/05/01 11:00:35 /home/ras/misc/go/gorm/main_test.go:42
[4.827ms] [rows:0] UPDATE `users` SET `name`='user' WHERE `id` = 7

2023/05/01 11:00:35 /home/ras/misc/go/gorm/main_test.go:42
[2.382ms] [rows:2] INSERT INTO `tokens` (`content`,`user_id`) VALUES ('token2_encrypted_encrypted',7) ON DUPLICATE KEY UPDATE `content`=VALUES(`content`)

2023/05/01 11:00:35 /home/ras/misc/go/gorm/main_test.go:42
[5.107ms] [rows:0] INSERT INTO `users` (`name`,`id`) VALUES ('user',7) ON DUPLICATE KEY UPDATE `name`=VALUES(`name`)
--- FAIL: Test_main (0.12s)
    main_test.go:36: 
                Error Trace:    /home/ras/misc/go/gorm/main_test.go:36
                Error:          Not equal: 
                                expected: "token2_encrypted"
                                actual  : "token2_encrypted_encrypted"
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -token2_encrypted
                                +token2_encrypted_encrypted
                Test:           Test_main
FAIL
FAIL    command-line-arguments  0.125s
FAIL
```

black-06:
- [X] Do only one thing
- [X] Tested

Hook is executed twice in `Save` func, close #6285 

### User Case Description

```go
type Token struct {
	ID  int
	Content string `gorm:"type:varchar(100)"`
}

func (t *Token) BeforeSave(tx *gorm.DB) error {
	t.Content += "_encrypted"
	return nil
}

token := Token{Content: "token"}
DB.Save(&token) // "token_encrypted"

token = Token{
	ID:           token.ID,
	Content: "token"
}
DB.Save(&token) // mysql RowsAffected is 0, it should be "token_encrypted" instead of "token_encrypted_encrypted"
```

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -107,7 +107,7 @@ func (db *DB) Save(value interface{}) (tx *DB) {
 		updateTx := tx.callbacks.Update().Execute(tx.Session(&Session{Initialized: true}))
 
 		if updateTx.Error == nil && updateTx.RowsAffected == 0 && !updateTx.DryRun && !selectedUpdate {
-			return tx.Clauses(clause.OnConflict{UpdateAll: true}).Create(value)
+			return tx.Session(&Session{SkipHooks: true}).Clauses(clause.OnConflict{UpdateAll: true}).Create(value)
 		}
 
 		return updateTx
--- a/tests/update_test.go
+++ b/tests/update_test.go
@@ -805,3 +805,76 @@ func TestUpdateWithDiffSchema(t *testing.T) {
 	AssertEqual(t, err, nil)
 	AssertEqual(t, "update-diff-schema-2", user.Name)
 }
+
+type TokenOwner struct {
+	ID    int
+	Name  string
+	Token Token `gorm:"foreignKey:UserID"`
+}
+
+func (t *TokenOwner) BeforeSave(tx *gorm.DB) error {
+	t.Name += "_name"
+	return nil
+}
+
+type Token struct {
+	UserID  int    `gorm:"primary_key"`
+	Content string `gorm:"type:varchar(100)"`
+}
+
+func (t *Token) BeforeSave(tx *gorm.DB) error {
+	t.Content += "_encrypted"
+	return nil
+}
+
+func TestSaveWithHooks(t *testing.T) {
+	DB.Migrator().DropTable(&Token{}, &TokenOwner{})
+	DB.AutoMigrate(&Token{}, &TokenOwner{})
+
+	saveTokenOwner := func(owner *TokenOwner) (*TokenOwner, error) {
+		var newOwner TokenOwner
+		if err := DB.Transaction(func(tx *gorm.DB) error {
+			if err := tx.Debug().Session(&gorm.Session{FullSaveAssociations: true}).Save(owner).Error; err != nil {
+				return err
+			}
+			if err := tx.Preload("Token").First(&newOwner, owner.ID).Error; err != nil {
+				return err
+			}
+			return nil
+		}); err != nil {
+			return nil, err
+		}
+		return &newOwner, nil
+	}
+
+	owner := TokenOwner{
+		Name:  "user",
+		Token: Token{Content: "token"},
+	}
+	o1, err := saveTokenOwner(&owner)
+	if err != nil {
+		t.Errorf("failed to save token owner, got error: %v", err)
+	}
+	if o1.Name != "user_name" {
+		t.Errorf(`owner name should be "user_name", but got: "%s"`, o1.Name)
+	}
+	if o1.Token.Content != "token_encrypted" {
+		t.Errorf(`token content should be "token_encrypted", but got: "%s"`, o1.Token.Content)
+	}
+
+	owner = TokenOwner{
+		ID:    owner.ID,
+		Name:  "user",
+		Token: Token{Content: "token2"},
+	}
+	o2, err := saveTokenOwner(&owner)
+	if err != nil {
+		t.Errorf("failed to save token owner, got error: %v", err)
+	}
+	if o2.Name != "user_name" {
+		t.Errorf(`owner name should be "user_name", but got: "%s"`, o2.Name)
+	}
+	if o2.Token.Content != "token2_encrypted" {
+		t.Errorf(`token content should be "token2_encrypted", but got: "%s"`, o2.Token.Content)
+	}
+}


a631807682 (commented):
Does this mean that AfterSave has been executed during Update even if it is not updated the record (in this case AfterSave becomes BeforeSave)

black-06:
> Does this mean that AfterSave has been executed during Update even if it is not updated the record (in this case AfterSave becomes BeforeSave)

Yes.

a631807682:
I seem to have other problems here, such as Create will not be executed when using Select, is Update necessary when we use OnConflict?
Of course, it's usually necessary to use Update and Create to avoid OnConflict, but if still need to use Create and OnConflict after Update, I'm not sure it's necessary.

black-06:
> I seem to have other problems here, such as Create will not be executed when using Select, is Update necessary when we use OnConflict? Of course, it's usually necessary to use Update and Create to avoid OnConflict, but if still need to use Create and OnConflict after Update, I'm not sure it's necessary.

Yes, it is necessary. `OnConflict` is added at #6149 .
 
For example, `token` in database (MySQL):
| id | content |
| -- | -- |
| 1 | str |
```sql
UPDATE `token` SET `content`='str' WHERE `id` = 1
-- rows: 0 (other db RowsAffected is 1)
-- Save will exec Create
-- Error 1062 (23000): Duplicate entry '1' for key 'token_owners.PRIMARY'
```

---

And this pr adds `SkipHooks: true`

a631807682:
> > I seem to have other problems here, such as Create will not be executed when using Select, is Update necessary when we use OnConflict? Of course, it's usually necessary to use Update and Create to avoid OnConflict, but if still need to use Create and OnConflict after Update, I'm not sure it's necessary.
> 
> Yes, it is necessary. `OnConflict` is added at #6149 .
> 
> For example, `token` in database (MySQL):
> 
> id	content
> 1	str
> ```sql
> UPDATE `token` SET `content`='str' WHERE `id` = 1
> -- rows: 0 (other db RowsAffected is 1)
> -- Save will exec Create
> -- Error 1062 (23000): Duplicate entry '1' for key 'token_owners.PRIMARY'
> ```
> 
> And this pr adds `SkipHooks: true`

My question is not whether OnConflict is necessary, but whether Update is necessary

black-06:
> > > I seem to have other problems here, such as Create will not be executed when using Select, is Update necessary when we use OnConflict? Of course, it's usually necessary to use Update and Create to avoid OnConflict, but if still need to use Create and OnConflict after Update, I'm not sure it's necessary.
> > 
> > Yes, it is necessary. `OnConflict` is added at #6149 .
> > 
> > For example, `token` in database (MySQL):
> > 
> > id	content
> > 1	str
> > ```sql
> > UPDATE `token` SET `content`='str' WHERE `id` = 1
> > -- rows: 0 (other db RowsAffected is 1)
> > -- Save will exec Create
> > -- Error 1062 (23000): Duplicate entry '1' for key 'token_owners.PRIMARY'
> > ```
> > 
> > And this pr adds `SkipHooks: true`
> 
> My question is not whether OnConflict is necessary, but whether Update is necessary

I have tried to create directly without update, but many tests have failed. 
I guess it's related to perm and  OnConflict will ignore them

a631807682:
I see, thank you.

jinzhu (approved):

PR #6303, opened 2023-05-09
fix: migrator run with nil schema (#5795)

### LINKED ISSUE #5795: DropColumn with string table panics with a nil dereference
## GORM Playground Link

https://github.com/go-gorm/playground/pull/532

## Description

When using `Migrator.DropColumn`, the table cannot be provided as a string. Providing it as a model works fine. I expect it to work even when the table is provided as a string.

As far as my shallow analysis can tell, the issue is caused by this:
- In `Migrator.RunWithValue`, the `Statement` is setup with a `nil` Schema. The Schema is only set in the `else` block visited in case the `value` parameter is *not* a string.
- `DropColumn`'s callback provided to `Migrator.RunWithValue` uses `stmt.Schema.LookUpField(name)`, which leads to a `nil` dereference in this case.

What I'm unsure about is how this should be fixed, hence no fixing PR. If you give me some pointers, I can give it a go, though.
What I'm equally unsure about is whether this issue happens with other methods, too. If it does, they should all disappear at once if the fix is done within `RunWithValue`.

black-06:
- [X] Do only one thing
- [X] Tested

fix `DropColumn` `RenameColumn` `HasColumn` with string table param.
Close #5795 

### User Case Description

It should work instead of panic
```go
migrator := DB.Migrator()
migrator.RenameColumn("column_structs", "new_name", "new_new_name")
migrator.HasColumn("column_structs", "new_new_name")
migrator.DropColumn("column_structs", "new_new_name")
```


a631807682 (commented):
This is a breaking change, because we use `if stmt.Schema == nil `in many places to judge

black-06:
> This is a breaking change, because we use `if stmt.Schema == nil `in many places to judge

Okay, I'll modify it

a631807682 (commented):
Test cases need to wait for drivers to be merged.
Can tests cover changes? For example, AlterColumn needs to provide datatype, so it does not support specifying by table name, and an error needs to be returned at this time.

black-06:
> Test cases need to wait for drivers to be merged. Can tests cover changes? For example, AlterColumn/AddColumn needs to provide datatype, so it does not support specifying by table name, and an error needs to be returned at this time.

modified~

black-06:
When `RenameIndex(&IndexStruct{}, "idx_index_structs_name", "idx_users_name_1")` in sqlite, it actually exec 
```sql
CREATE INDEX `idx_users_name_1` ON `index_structs`(`name`)
```

This is a bug, right? Because idx_index_structs_name still exist. 
We need to fix it first

a631807682:
> When `RenameIndex(&IndexStruct{}, "idx_index_structs_name", "idx_users_name_1")` in sqlite, it actually exec
> 
> ```sql
> CREATE INDEX `idx_users_name_1` ON `index_structs`(`name`)
> ```
> 
> This is a bug, right? Because idx_index_structs_name still exist. We need to fix it first

yes

Issue #6288, opened 2023-05-02
embedded should be nil if not exists


hykuan:
## Your Question

refer to the pr #6219 

the test from tests/embedded_struct_test.go works fine.

but after i added one field with data type int, the test will failed.

from 
``` golang
type Author struct {
	ID    string
	Name  string
	Email string
}
```

to
``` golang
type Author struct {
	ID    string
	Name  string
	Email string
	Age   int
}
```

the output will show: `Expected to get back a nil Author but got: &{  0}`

## Expected answer

the author should also be nil.

## More information

I test int & uint types, here is the result

data type | int64 | int32 | int16 | int8 | int | uint64 | uint32 | uint16 | uint8 | uint
--- | --- | --- | --- |--- |--- |--- |--- |--- |--- |--- 
test result | pass | failed | failed | failed | failed | pass | failed | failed | failed | failed

hykuan:
maybe we need to add something like this ?

from gorm.io/gorm/schema/field.go

```
	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
		field.Set = func(ctx context.Context, value reflect.Value, v interface{}) (err error) {
			switch data := v.(type) {
			case **int64:
				if data != nil && *data != nil {
					field.ReflectValueOf(ctx, value).SetInt(**data)
				}
+			case **int:
+				if data != nil && *data != nil {
+					field.ReflectValueOf(ctx, value).SetInt(int64(**data))
+				}

```

aclich:
Hi
I'm facing similar issue too. After tested, found it was cause by the following datatypes
- interface{} (custom data types)
- *interface{} (custom data types)
- *time.time

I add those types in tests/embedded_struct_test.go, and it will also make the test failed.
Like following example
```go
type Content struct {
	Content interface{} `gorm:"type:String"`
}

func (c Content) Value() (driver.Value, error) {
	return json.Marshal(c)
}

func (c *Content) Scan(src interface{}) error {
	...
}

type Author struct {
	ID          string
	Name        string
	Email       string
	Age         int
	Content     Content
	ContentPtr  *Content
	BirthDay    time.Time
	BirthDayPtr *time.Time
}
```
I've trying to change the reflect logic on struct scanner & pointer scanner to resolve this problem. Not sure if it's gorm wants?
`gorm.io/gorm/schema/field.go`
<pre><b>diff --git a/schema/field.go b/schema/field.go</b>
[...]

PR #6311, opened 2023-05-11
fix: 🐛 embedded struct test failed with custom datatypes

aclich:
- [x] Do only one thing
- [x] Tested

Refer to https://github.com/go-gorm/gorm/issues/6288
Fix embedded struct with custom datatypes and *time.time will failed the test.

### User Case Description

Test should pass.

### DIFF
--- a/schema/field.go
+++ b/schema/field.go
@@ -846,7 +846,7 @@ func (field *Field) setupValuerAndSetter() {
 			field.Set = func(ctx context.Context, value reflect.Value, v interface{}) error {
 				switch data := v.(type) {
 				case **time.Time:
-					if data != nil {
+					if data != nil && *data != nil {
 						field.ReflectValueOf(ctx, value).Set(reflect.ValueOf(*data))
 					}
 				case time.Time:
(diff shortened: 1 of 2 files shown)


aclich:
mssql will have type issue while inserting null value into `varbinary(max)` type.
```sh
/home/diad/Developer/hbs4/gorm/tests/embedded_struct_test.go:130 mssql: Implicit conversion from data type nvarchar to varbinary(max) is not allowed. Use the CONVERT function to run this query.
[16347.552ms] [rows:0] INSERT INTO "hn_posts" ("title","url","upvotes","user_id","user_name","user_email","user_age","user_content","user_content_ptr","user_birthday","user_birthday_ptr") OUTPUT INSERTED."id" VALUES ('embedded_pointer_type','',0,NULL,NULL,NULL,NULL,NULL,NULL,NULL,NULL);
```
reference https://stackoverflow.com/questions/11411854/implicit-conversion-from-data-type-nvarchar-to-varbinarymax-is-not-allowed

Need to fix gorm driver first.

aclich:
commit https://github.com/go-gorm/gorm/commit/dc149c1127c45e889263a3164fa9757f69ceca09
Modify the custom datatypes to string for preventing mssql driver issue.
Reopen PR, fixing embedded struct should be nil.

a631807682 (approved):

PR #6337, opened 2023-05-20
max identifier length changed to 63

alidevhere:
- [x ] Do only one thing
- [ x] Non breaking API changes
- [ x] Tested

default max identifier length changed to 63

See issue [here](https://github.com/go-gorm/gorm/issues/6098)
It adds max Identifier Length in Naming strategy which can be overridden by individual drivers of gorm.


alidevhere:
@jinzhu once these changes are merged i will create pull request in all drivers to support this change.
Thanks

jinzhu:
Hi @alidevhere 

Thank you for the max length support, but maybe we should keep the old configuration by default to make sure old applications not using Postgres don't break.

alidevhere:
> Hi @alidevhere
> 
> Thank you for the max length support, but maybe we should keep the old configuration by default to make sure old applications not using Postgres don't break.

Should we change default max length identifier to old 64 instead of 63 ?

jinzhu:
> > Hi @alidevhere
> > Thank you for the max length support, but maybe we should keep the old configuration by default to make sure old applications not using Postgres don't break.
> 
> @jinzhu Should we change default max length identifier to old 64 instead of 63 ?

Better to keep the current setting to keep compatibility.

alidevhere:
> > > Hi @alidevhere
> > > Thank you for the max length support, but maybe we should keep the old configuration by default to make sure old applications not using Postgres don't break.
> > 
> > 
> > @jinzhu Should we change default max length identifier to old 64 instead of 63 ?
> 
> Better to keep the current setting to keep compatibility.

@jinzhu  Changes reverted to current settings. Ready for review

jinzhu (approved):

PR #6366, opened 2023-05-31
feat: add SQLConnector

lzakharov:
- [x] Do only one thing
- [ ] Tested (**looks unnecessary**)

This PR adds ability to provide `*sql.DB` connector which can resolve current connect based on the database context. It works the same as `GetDBConnector` but additionally have `*gorm.DB` state.

### User Case Description

Since we use [DBResolver](https://gorm.io/docs/dbresolver.html) plugin to provide multiple databases support, sometimes it's necessary to get raw `*sql.DB`. At this moment all these statements return the same connection pool, ignoring provided clauses:

```go
db.DB()
db.Clauses(dbresolver.Write).DB()
db.Clauses(dbresolver.Read).DB()
db.Clauses(dbresolver.Use("slow")).DB()
```

However, it seems to us more logical behavior to return different connections based on the current database state.

### DIFF
--- a/interfaces.go
+++ b/interfaces.go
@@ -77,6 +77,12 @@ type GetDBConnector interface {
 	GetDBConn() (*sql.DB, error)
 }
 
+// SQLConnector represents SQL db connector which takes into account the current
+// database context
+type SQLConnector interface {
+	GetSQLConn(db *DB) (*sql.DB, error)
+}
+
 // Rows rows interface
 type Rows interface {
 	Columns() ([]string, error)
--- a/gorm.go
+++ b/gorm.go
@@ -375,6 +375,10 @@ func (db *DB) AddError(err error) error {
 func (db *DB) DB() (*sql.DB, error) {
 	connPool := db.ConnPool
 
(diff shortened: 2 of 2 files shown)


interfaces.go
```diff
@@ -77,6 +77,12 @@ type GetDBConnector interface {
 	GetDBConn() (*sql.DB, error)
 }
 
+// SQLConnector represents SQL db connector which takes into account the current
+// database context
+type SQLConnector interface {
```
a631807682, line 82: This is a good point, but since it expresses a similar meaning to GetDBConnector (returns SQL db connector), I feel that the current naming is not easy to distinguish between them. Can we make it clear in naming that this interface needs to provide context?

lzakharov: Agree, I feel the same about the naming. Maybe:

```go
type GetDBConnectorWithContext interface {
	GetDBConnWithContext(db *DB) (*sql.DB, error)
}
```

a631807682: I think it's ok

a631807682 (approved):

saeidee (approved):

PR #6373, opened 2023-06-03
reafactor: add nil detection when sqldb return

KantaHasegawa:
- [x] Do only one thing
- [x] Tested

This pull request addresses a bug that caused a panic when connecting to the database using an incorrect DSN. The bug has already been fixed in this commit (https://github.com/go-gorm/gorm/commit/c1ea73036715018a1bb55cdb8690441044e13a76). However, to prevent potential future bugs, I have refactored the code to include null detection during type assertion.

### User Case Description

https://github.com/go-gorm/gorm/blob/c1ea73036715018a1bb55cdb8690441044e13a76/gorm.go#L374-L387

Before merging this PR (c1ea730), `connPool.(*sql.DB)` returned `nil, nil` which had a negative impact on the called method.

### DIFF
--- a/gorm.go
+++ b/gorm.go
@@ -379,7 +379,7 @@ func (db *DB) DB() (*sql.DB, error) {
 		return dbConnector.GetDBConn()
 	}
 
-	if sqldb, ok := connPool.(*sql.DB); ok {
+	if sqldb, ok := connPool.(*sql.DB); ok && sqldb != nil {
 		return sqldb, nil
 	}
 
--- a/gorm.go
+++ b/gorm.go
@@ -376,7 +376,9 @@ func (db *DB) DB() (*sql.DB, error) {
 	connPool := db.ConnPool
 
 	if dbConnector, ok := connPool.(GetDBConnector); ok && dbConnector != nil {
-		return dbConnector.GetDBConn()
+		if sqldb, err := dbConnector.GetDBConn(); sqldb != nil && err != nil {
+			return sqldb, err
+		}
 	}
 
 	if sqldb, ok := connPool.(*sql.DB); ok && sqldb != nil {
--- a/gorm.go
+++ b/gorm.go
@@ -181,7 +181,7 @@ func Open(dialector Dialector, opts ...Option) (db *DB, err error) {
 		err = config.Dialector.Initialize(db)
 
 		if err != nil {
-			if db, _ := db.DB(); db != nil {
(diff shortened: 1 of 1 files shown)


a631807682 (commented):
https://github.com/go-gorm/gorm/commit/c1ea73036715018a1bb55cdb8690441044e13a76 is ok to me, and this change is also the same, but in order to achieve the purpose of your appeal, we also need to detect the content returned by `dbConnector.GetDBConn()` to ensure that if `*sql.DB` is `nil`, an error must be returned , and you can revert the error judged code in https://github.com/go-gorm/gorm/commit/c1ea73036715018a1bb55cdb8690441044e13a76

KantaHasegawa:
Thank you for the review.
Does this code meet your requirements?

```
if dbConnector, ok := connPool.(GetDBConnector); ok && dbConnector != nil {
	if sqldb, err := dbConnector.GetDBConn(); sqldb != nil && err != nil {
		return sqldb, err
	}
}
```

gorm.go
```diff
@@ -376,10 +376,12 @@ func (db *DB) DB() (*sql.DB, error) {
 	connPool := db.ConnPool
 
 	if dbConnector, ok := connPool.(GetDBConnector); ok && dbConnector != nil {
-		return dbConnector.GetDBConn()
+		if sqldb, err := dbConnector.GetDBConn(); sqldb != nil && err != nil {
+			return sqldb, err
+		}
 	}
```
KantaHasegawa, line 382: Thank you for the review.
Does this code meet your requirements?

gorm.go
```diff
@@ -376,10 +376,12 @@ func (db *DB) DB() (*sql.DB, error) {
 	connPool := db.ConnPool
 
 	if dbConnector, ok := connPool.(GetDBConnector); ok && dbConnector != nil {
-		return dbConnector.GetDBConn()
+		if sqldb, err := dbConnector.GetDBConn(); sqldb != nil && err != nil {
```
a631807682, line 379: Is `&& err== nil` here?

KantaHasegawa: I apologize for the careless mistake.
In the case where `dbConnector.GetDBConn()` returns `nil, error`, checking `sqldb != nil` alone is not sufficient to return the error properly.
Therefore, `sqldb != nil || error != nil` is the correct condition.

a631807682 (approved):
thanks for your contribution

Issue #5401, opened 2022-06-06
Support `unique` and `uniqueIndex` together on the same field.


ihucos:
## Describe the feature

Allow uniqueIndex and index struct tags to be used at the time. At the moment it seems I can do one or the other, but not both.

## Motivation

SQL supports this, so I expect gorm to support it. I need it for my software.

a631807682:
https://github.com/go-gorm/gorm/issues/6381#issuecomment-1929427042

Issue #6381, opened 2023-06-06
Refactor: Distinguish between unique and UniqueIndex


a631807682:
## Describe the feature

Distinguish between unique index and index to better support them.

## Motivation
Because we can't distinguish them at present, it causes such as https://github.com/go-gorm/gorm/issues/6224#issuecomment-1547623416, and column will not be able to set [unique index and index at the same time](https://github.com/go-gorm/gorm/issues/5401)

The following tasks are open 

- [x] Distinguish between unique index and index in gorm @black-06 
- [x] Distinguish between unique index and index in postgres @black-06 
- [x] Distinguish between unique index and index in sqlserver @black-06 
- [x] Distinguish between unique index and index in sqlite @black-06 
- [x] Distinguish between unique index and index in mysql @black-06 

If anyone is interested in the above tasks, please leave a message in the comments.

## Related Issues

related issues:
https://github.com/go-gorm/gorm/issues/6224
https://github.com/go-gorm/gorm/issues/5401
https://github.com/go-gorm/gorm/issues/5715
https://github.com/go-gorm/gorm/issues/5681
https://github.com/go-gorm/gorm/issues/6378
https://github.com/go-gorm/gorm/issues/6407
related pull request:
https://github.com/go-gorm/gorm/pull/5341

black-06:
Let me do the distinguish in gorm (I touched the checkbox by mistake

black-06:
Is the index created repeatedly because of this? (#5715 & #6224)
Or is there any fix in Dialector ?

Here's a playground I created: https://github.com/go-gorm/playground/pull/609
It says no duplicate indexes are created.

a631807682:
> Is the index created repeatedly because of this? (#5715 & #6224) Or is there any fix in Dialector ?
> 
> Here's a playground I created: [go-gorm/playground#609](https://github.com/go-gorm/playground/pull/609) It says no duplicate indexes are created.

Here is a mysql example
```go
type Test struct {
	ID   uint   `gorm:"primaryKey"`
	Name string `gorm:"type:varchar(10);uniqueIndex"`
}

type Test1 struct {
	ID   uint   `gorm:"primaryKey"`
	Name string `gorm:"type:varchar(20);uniqueIndex"`
}

func TestGORM(t *testing.T) {
	DB.Migrator().DropTable(&Test{})
	DB.AutoMigrate(&Test{})
	DB.Table("tests").AutoMigrate(&Test1{})

[...]

Issue #6378, opened 2023-06-06
AutoMigrate creating different unique Index on subsequent runs


cabeedelhertz:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/608

## Description

I have an existing database and consistently run `Automigrate` but after upgrading my gorm version from 1.28.3, the automigrate is creating different unique indexes than the ones I have defined and already exist in my database.

To reproduce, pull this branch and run `GORM_DIALECT=postgres go test`. This should work and pass tests. But if you run it again, it will fail to make migrations because it is trying to create a uniqueIndex and uniqueConstraint that doesn't have the same `name` or `where` clause as the unique index that I defined:
`ERROR: could not create unique index "idx_users_description" (SQLSTATE 23505)`

After the initial run of `AutoMigrate` the unique indexes that I defined are correctly created:
<img width="500" alt="image" src="https://github.com/go-gorm/playground/assets/60052974/c2c1b06a-119d-4940-9fe3-345a0403bd23">

But after the next time I run `AutoMigrate`, it created new unique indexes with a different name and without the where clause:
<img width="500" alt="image" src="https://github.com/go-gorm/playground/assets/60052974/b58cf2b0-c873-4fc2-a0ab-ac2fc8e19538">

I can only get to the state of the second picture if I don't have any data that conflicts the `where` clause of the unique index of the `description` column. In my use case, I have a table with a unique index with a `where deleted_at is NULL` condition but now that my `AutoMigrate` behaves differently, I cannot run migrations.

Did something change with how `AutoMigrate` handles unique indexes?

a631807682:
What version of postgres are you using? This seems to be an error caused by the handling of the index name, but in fact I cannot reproduce the wrong creation of this index from the link you provided.

cabeedelhertz:
> What version of postgres are you using? This seems to be an error caused by the handling of the index name, but in fact I cannot reproduce the wrong creation of this index from the link you provided.

using `gorm.io/driver/postgres v1.5.2`

cabeedelhertz:
> fact I cannot reproduce the wrong creation of this index from the link you provided.

And yeah sorry with the given tests you can't actually create the wrong index because the tests insert conflicting data. But you should see errors that it is failing to create those indexes. I just included the "wrong" index in the description to show what `autoMigrate` is _trying_ to create

a631807682:
refer to https://github.com/go-gorm/gorm/issues/6381#issuecomment-1580164621
[...]

Issue #6224, opened 2023-04-08
Unnecessarily unique indexes created when running `AutoMigrate`


shanehou:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/589

## Description

Model `User` has only one field `Name`, with `uniqueIndex` tag. First time I run `AutoMigrate`, GORM will create two unique indexes called `idx_users_name` and `name`. Then every time I run `AutoMigrate`, GORM will create a new unique index called `name_2`, `name_3` and so on.

If I use `uniqueIndex:i_u_name` to specify the name of the unique index, the first run would still create `i_u_name` and `name` indexes, and the following runs would still create `name_2`, `name_3` indexes.

a631807682:
This error is caused by the lack of distinction between Unique and UniqueIndex in schema.Filed and gorm.ColumnType.
A good way is to record UniqueIndex in them and compare them, and a not so good way is to re-compare them in AlterColumn (like https://github.com/go-gorm/postgres/pull/157).

shanehou:
> This error is caused by the lack of distinction between Unique and UniqueIndex in schema.Filed and gorm.ColumnType. A good way is to record UniqueIndex in them and compare them, and a not so good way is to re-compare them in AlterColumn (like [go-gorm/postgres#157](https://github.com/go-gorm/postgres/pull/157)).

How should I do that in go struct? I see it as a bug, because I just followed the doc to define UniqueIndex. No more operations.

shanehou:
> 

When `AlterColumn`, it'll run `ALTER TABLE {table_name} MODIFY COLUMN {field} {fullDataType}`. If `fullDataType` has `UNIQUE`, MySQL will always creates a new unique index. So the solution is either checking if the unique index is already created before running the `ALTER TABLE` statement (your "not-so-good-way" I suppose?), or removing `UNIQUE` in `fullDataType` early (your "good-way" I suppose?). But I think the former way seems more reasonable, because `fullDataType` will clearly be used in other places like `CreateTable`.

a631807682:
> When `AlterColumn`, it'll run `ALTER TABLE {table_name} MODIFY COLUMN {field} {fullDataType}`. If `fullDataType` has `UNIQUE`, MySQL will always creates a new unique index.
> 
[...]

PR #6386, opened 2023-06-07
refactor: distinguish between Unique and UniqueIndex

black-06:
- [X] Do only one thing
- [X] Tested

See #6381

Related issues:
#6224
#5401
#5715
#5681
#6378
#6407


a631807682 (commented):
Can we add more tests showing what this change fixes? The test doesn't need to be successful as I can imagine this would involve a lot of driver changes.
I take mysql as an example
1. We need to ensure that the field A marked as unique index is not `a varchar(20) unique` when creating table, but `UNIQUE INDEX idx_a (a)`
Link https://github.com/go-gorm/gorm/issues/6378
2. When comparing field A, we may need to compare `field.Unique` and `field.UniqueIndex` respectively (this may cause failure, because currently mysql does not handle the columnKey value `'MUL'`, and unique index may have multiple)
Link https://github.com/go-gorm/gorm/issues/5681
https://github.com/go-gorm/mysql/blob/master/migrator.go#L264
3. When comparing, it is necessary to ensure the deletion and addition of the unique index
Link https://github.com/go-gorm/gorm/issues/6224

Since this PR may contain relatively large changes, I want to describe it here. This is what I can think of so far. If this PR will fix more problems, welcome to add.

black-06:
Can I use `"github.com/stretchr/testify/assert"` in "gorm.io/gorm/tests" ?
I need `ElementsMatch`, or copy code to tests package ?

a631807682:
> Can I import `"github.com/stretchr/testify/assert"` to "gorm.io/gorm/tests" ? I need `ElementsMatch`, or copy code to tests package ?

Both ok for me.

black-06:
What should we do with 
```
`gorm:"uniqueIndex;index"`
```

a631807682:
> What should we do with
> 
> ```
> `gorm:"uniqueIndex;index"`
> ```

This situation is meaningless, it is equivalent to `gorm:"uniqueIndex;"`, but the following scenarios will exist
```
Name string `gorm:"uniqueIndex:idx_name_code"`
Code string `gorm:"uniqueIndex:idx_name_code;index"`
```

black-06:
> > What should we do with
> > ```
> > `gorm:"uniqueIndex;index"`
> > ```
> 
> This situation is meaningless, it is equivalent to `gorm:"uniqueIndex;"`, but the following scenarios will exist
> 
> ```
> Name string `gorm:"uniqueIndex:idx_name_code"`
> Code string `gorm:"uniqueIndex:idx_name_code;index"`
> ```

In this case:
```go
type UserIndex struct {
	Name8        string `gorm:"index:,length:10;index:,collate:utf8"`
}
```

It got index 
```go
"idx_user_indices_name8": {
	Name: "idx_user_indices_name8",
	Type: "",
	Fields: []schema.IndexOption{
		{Field: &schema.Field{Name: "Name8"}, Length: 10},
		{Field: &schema.Field{Name: "Name8"}, Collate: "utf8"},
	},
}
```
Maybe Index.Fields ([]IndexOption) should be changed to Map[string]IndexOption which key is field name.

a631807682:
> ```go
> type UserIndex struct {
> 	Name8        string `gorm:"index:,length:10;index:,collate:utf8"`
> }
> ```

The same index may indeed be created multiple times, but I am not sure whether the actual usage scenario exists. In this case, I can only think that some scenarios may need to be specified using FORCE INDEX, but in this case, I think users can specify an index name instead of using the default name.
In short, I think there are too few usage scenarios to consider this situation, of course if my thinking is wrong, please let me know.

black-06:
> In short, I think there are too few usage scenarios to consider this situation

I mean, it's an unreported bug.
We don't have to  fix it (because there are very few usage scenarios).


What is your opinion?  fix or not?

a631807682:
> > In short, I think there are too few usage scenarios to consider this situation
> 
> I mean, it's an unreported bug. We don't have to fix it (because there are very few usage scenarios).
> 
> What is your opinion? fix or not?

We don't have to fix it, specifying the index name can avoid.

black-06:
> 1. We need to ensure that the field A marked as unique index is not `a varchar(20) unique` when creating table, but `UNIQUE INDEX idx_a (a)`

For `unique`, we should do
```sql
CREATE TABLE `index_tests` (
  `name` VARCHAR(10),
  UNIQUE KEY `uni_index_tests_name` (`name`)
);
```
or
```sql
CREATE TABLE `index_tests` (
  `name` VARCHAR(10),
  UNIQUE INDEX `idx_index_tests_name` (`name`)
);
```
There is no difference between them in mysql,
but I don't know the situation of other databases.

For option 1, I should add a `ParseUniqueConstraints` like `ParseCheckConstraints`.
For option 2, I should modify in `ParseIndexes`.

a631807682:
> > 1. We need to ensure that the field A marked as unique index is not `a varchar(20) unique` when creating table, but `UNIQUE INDEX idx_a (a)`
> 
> For `unique`, we should do
> 
> ```sql
> CREATE TABLE `index_tests` (
>   `name` VARCHAR(10),
>   CONSTRAINT `uni_index_tests_name` UNIQUE (`name`)
> );
> ```
> 
> or
> 
> ```sql
> CREATE TABLE `index_tests` (
>   `name` VARCHAR(10),
>   UNIQUE INDEX `idx_index_tests_name` (`name`)
> );
> ```
> 
> There is no difference between them in mysql, but I don't know the situation of other databases.
> 
> For option 1, I should add a `ParseUniqueConstraints` like `ParseCheckConstraints`. For option 2, I should modify in `ParseIndexes`.

```go
type User struct {
	U  string `gorm:"type:varchar(20);uniqueIndex"`
	UI string `gorm:"type:varchar(20);unique"`
}
```

Consistent with the current, we need to remove the redundant unique statement
```diff
CREATE TABLE `users` (
	`id` BIGINT UNSIGNED AUTO_INCREMENT,
	`u` VARCHAR ( 20 ),
-	`ui` VARCHAR ( 20 ) UNIQUE,
+	`ui` VARCHAR ( 20 ), 
	PRIMARY KEY ( `id` ),
INDEX `idx_users_deleted_at` ( `deleted_at` ),
UNIQUE INDEX `idx_users_u` ( `u` ))
```

a631807682:
> > 1. We need to ensure that the field A marked as unique index is not `a varchar(20) unique` when creating table, but `UNIQUE INDEX idx_a (a)`
> 
> For `unique`, we should do
> 
> ```sql
> CREATE TABLE `index_tests` (
>   `name` VARCHAR(10),
>   CONSTRAINT `uni_index_tests_name` UNIQUE (`name`)
> );
> ```
> 
> or
> 
> ```sql
> CREATE TABLE `index_tests` (
>   `name` VARCHAR(10),
>   UNIQUE INDEX `idx_index_tests_name` (`name`)
> );
> ```
> 
> There is no difference between them in mysql, but I don't know the situation of other databases.
> 
> For option 1, I should add a `ParseUniqueConstraints` like `ParseCheckConstraints`. For option 2, I should modify in `ParseIndexes`.

Consistent with the current, we need to remove the redundant unique statement.

```go
type User struct {
	ID uint `gorm:"primarykey"`
	UI string `gorm:"type:varchar(20);uniqueIndex"`
	U  string `gorm:"type:varchar(20);unique"`
}
```

```diff
CREATE TABLE `users` (
	`id` BIGINT UNSIGNED AUTO_INCREMENT,
-	`ui` VARCHAR ( 20 ) UNIQUE,
+	`ui` VARCHAR ( 20 ),
	`u` VARCHAR ( 20 ) UNIQUE,
	PRIMARY KEY ( `id` ),
INDEX `idx_users_deleted_at` ( `deleted_at` ),
UNIQUE INDEX `idx_users_ui` ( `ui` ))
```

black-06:
There is no doubt that we need to remove the redundant unique statement.
My question is, should we declare this field as `UNIQUE INDEX` or `UNIQUE CONSTRAINT` ?

```diff
+ UNIQUE INDEX `idx_users_ui` ( `ui` ))
or
+ CONSTRAINT `idx_users_ui` UNIQUE (`ui`)
```

a631807682:
> There is no doubt that we need to remove the redundant unique statement. My question is, should we declare this field as `UNIQUE INDEX` or `UNIQUE CONSTRAINT` ?
> 
> ```diff
> + UNIQUE INDEX `idx_users_ui` ( `ui` ))
> or
> + CONSTRAINT `idx_users_ui` UNIQUE (`ui`)
> ```

```sql
 UNIQUE INDEX `idx_users_ui` ( `ui` ))
```

black-06:
> ```sql
>  UNIQUE INDEX `idx_users_ui` ( `ui` ))
> ```

I think it should be a constraint.
If it's an index, then we distinguish them without meaning.
And I queried some [docs](https://www.w3schools.com/sql/sql_unique.asp), they said the `Unique` is constraint

a631807682:
> > ```sql
> >  UNIQUE INDEX `idx_users_ui` ( `ui` ))
> > ```
> 
> I think it should be a constraint. If it's an index, then we distinguish them without meaning. And I queried some [docs](https://www.w3schools.com/sql/sql_unique.asp), they said the `Unique` is constraint

When the user uses the `uniqueIndex` tag, I think what the user wants to create is an index (even if they are the same in most cases), including when we define the `uniqueIndex` tag, we parse and create a `partial index`, which may fail when creating a unique constraint?

It's a `constraint` when the user uses the `unique` tag, but since we're currently only receiving `boolean` values, it doesn't make much sense to change it.

https://stackoverflow.com/questions/23542794/postgres-unique-constraint-vs-index
https://dba.stackexchange.com/questions/144/when-should-i-use-a-unique-constraint-instead-of-a-unique-index/147#147?newreg=e5609e35473c4d0b912f18996c349674

black-06:
> When the user uses the uniqueIndex tag, I think what the user wants to create is an index (even if they are the same in most cases), including when we define the uniqueIndex tag, we parse and create a partial index, which may fail when creating a unique constraint?

For `uniqueIndex` tag, there is no doubt that we created UniqueIndex.

> It's a constraint when the user uses the unique tag

+1

> but since we're currently only receiving boolean values, it doesn't make much sense to change it.

So we use the `unique` tag as the unique index?
```diff
GORM tag: unique
- unique constraint
+ unique index
```
But as you shared said, they have some subtle differences, such as error code, foreign key. 

summary:
| GORM tag | DB | 
| -- | -- |
| unique | unique constraint (or index?) <br/>Here I think it should be a constraint | 
| uniqueIndex | unique index |
| index | index |
| index:,unique | unique index |

a631807682:
> So we use the unique tag as the unique index?

It should be `unique constraint` , but since we don't support multiple columns and we don't support patch constraint , so I don't think it makes sense to change it, can we change it in a later version or PR?

https://github.com/go-gorm/gorm/issues/5559#issuecomment-1200718658

black-06:
> can we change it in a later version or PR?

Okay.

> Or changing it now will make subsequent work becomes easier?

No, Just more clearly

a631807682:
black-06 proposes that if unique is defined as unique index, it will be indistinguishable (and compared) in database query, and the database query result unique and uniqueIndex tags are the same, so unique needs to be changed into a constraint.
Can't resolve #5401

Record this comment.

migrator/migrator.go
```diff
@@ -523,13 +522,105 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
...
+					if columns := index.Columns(); len(columns) != 1 || columns[0] != field.DBName {
+						continue
+					}
+					if name := index.Name(); name == constraint || name == field.UniqueIndex {
+						continue
+					}
+					if err := execTx.Migrator().DropIndex(value, index.Name()); err != nil {
+						return err
+					}
+				}
```
black-06, line 572: Because Unique is affected by UniuqeIndex,
I had to clear redundant indexes here.

But this makes different `Dialector` behave differently.
For example, in unit tests, mysql behaves differently from other dialectors.
See https://github.com/go-gorm/gorm/pull/6386/files#diff-23864390402317a89bc6a5197a8144403c082c4e2a3d0132bac131454831b5f9R1717-R1760

chrismoran-mica: Just watching this unfold as a spectator eagerly awaiting the outcome. We are currently locked on v1.24.2 because upgrading that version breaks our migrations miserably. Investigation has shown the break is due an issue that this PR appears to fix. Thank you both!

migrator/migrator.go
```diff
@@ -523,13 +522,105 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
...
+				if field.UniqueIndex != "" {
+					if err := execTx.Migrator().CreateIndex(value, field.UniqueIndex); err != nil {
+						return err
+					}
+				}
+				return nil
+			}
+		} else {
+			if unique && !field.Unique {
```
a631807682, line 613: Do we need to handle this situation?
```diff
-	Name  string `gorm:"type:varchar(20);uniqueIndex:udx_name_name1"`
-	Name1 string `gorm:"type:varchar(20);uniqueIndex:udx_name_name1"`
+	Name  string `gorm:"type:varchar(20);uniqueIndex"`
+	Name1 string `gorm:"type:varchar(20);unique"`
```

migrator/migrator.go
```diff
@@ -523,13 +522,105 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
...
+	}
+
+	if err := m.MigrateColumnUnique(value, field, columnType); err != nil {
+		return err
 	}
 
 	return nil
 }
 
+func (m Migrator) MigrateColumnUnique(value interface{}, field *schema.Field, columnType gorm.ColumnType) error {
```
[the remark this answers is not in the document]
black-06: > do we need to add to the interface, looks like MigrateConstraint?

We can add to the interface, but `MigrateConstraint` may not be a good name...

migrator/migrator.go
```diff
@@ -599,37 +690,35 @@ func (m Migrator) DropView(name string) error {
...
-	for _, field := range constraint.ForeignKeys {
-		foreignKeys = append(foreignKeys, clause.Column{Name: field.DBName})
-	}
-
-	for _, field := range constraint.References {
-		references = append(references, clause.Column{Name: field.DBName})
+// GuessConstraintAndTable guess statement's constraint and it's table based on name
+//
+// Deprecated: use GuessConstraintInterfaceAndTable instead.
+func (m Migrator) GuessConstraintAndTable(stmt *gorm.Statement, name string) (*schema.Constraint, *schema.CheckConstraint, string) {
```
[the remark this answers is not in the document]
black-06: > If it is marked as Deprecated, do I need to change it at the same time in the driver?

Yes, I have changed it in the pr of Dialectors.

migrator/migrator.go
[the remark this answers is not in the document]
a631807682: > > do we need to add to the interface, looks like MigrateConstraint?
> 
> We can add to the interface, but `MigrateConstraint` may not be a good name...

Naming is a difficult for me, you can decide for yourself.

migrator/migrator.go
a631807682: > What do you think is better?

When the unique of the field is equal, it seems that the index needs to be checked. If the Columns of the index do not contain the name of the field, the unique index may need to be deleted? I haven't researched it yet, because it may be more complicated, such as how to deal with adding or deleting a field in the composite index.
Of course, this doesn't have to be done in a PR, it's just a question, and I think it needs to be investigated and concluded.

black-06: > When the unique of the field is equal, it seems that the index needs to be checked. 

It's not,  because `UniqueAffectedByUniqueIndex` is false.
In here, even if there is a UniqueIndex, it reports that Unique in ColumnType is false.

When `UniqueAffectedByUniqueIndex` is false:

```diff
- Name  string `gorm:"type:varchar(20);uniqueIndex:udx_name_name1"`
+ Name  string `gorm:"type:varchar(20);uniqueIndex"`
```
only the UniqueIndex names are different (not handle yet)

```diff
- Name1 string `gorm:"type:varchar(20);uniqueIndex:udx_name_name1"`
+ Name1 string `gorm:"type:varchar(20);unique"`
```
only add Unique constraints.


Behaviors: (That's actually what the test is about)

|  From | To |  UniqueAffected<br/>ByUniqueIndex | Behavior in MigrateColumnUnique |
| -- | -- | -- | -- |
| `gorm:"size:10"` | `gorm:"size:30;unique"` | true/false | add unique constraint |
| `gorm:"size:10"` | `gorm:"size:50;uniqueIndex"` | true | add unique index |
| `gorm:"size:10"` | `gorm:"size:50;uniqueIndex"` | false | - |
| `gorm:"size:30;unique"` | `gorm:"size:10"` | true/false | drop unique constraint |
| `gorm:"size:30;unique"` | `gorm:"size:50;uniqueIndex"` | true | drop unique constraint<br/>add unique index |
| `gorm:"size:30;unique"` | `gorm:"size:50;uniqueIndex"` | false | drop unique constraint |
| `gorm:"size:50;uniqueIndex"` | `gorm:"size:10"` | true | drop unique index |
| `gorm:"size:50;uniqueIndex"` | `gorm:"size:10"` | false | - |

black-06: Once `UniqueAffectedByUniqueIndex`  is false, we don't care about UniqueIndex

a631807682 (commented):
# en
There doesn't seem to be much of a problem for me, next I need time to check the following
1. Test cases and driver changes are used to ensure that unique behavior meets expectations
2. The driver that involves mutual influence between unique and unique index, whether the behavior of unique index is also in line with expectations.
3. Does the test case contain the scenario pointed out by the associated issue?

This is a breaking change. Tables created by lower versions may be affected by the behavior of the new version, but we may not pay much attention to it, so we need to pay special attention to the version number to avoid causing users to upgrade by mistake.

# zh
对我来说似乎没有太多问题，接下来我需要时间检查以下内容
1. 测试用例以及驱动的更改用于保证 unique 行为符合预期
2. 涉及 unique 和 unique index 互相影响的驱动，unique index 行为是否也符合预期。
3. 测试用例是否包含关联 issue 指出的场景

这是一个破坏更改，可能通过低版本创建的表会被新版本的行为所影响，但我们可能并不太关注它，因此需要特别注意版本号避免导致用户错误升级。

migrator/migrator.go
```diff
@@ -99,15 +95,20 @@ func (m Migrator) FullDataTypeOf(field *schema.Field) (expr clause.Expr) {
 	return
 }
 
+func (m Migrator) GetQueryAndExecTx() (queryTx, execTx *gorm.DB) {
```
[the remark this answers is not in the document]
black-06: > # en
> Is this a public function? Will it be a problem to change it to private?
> 
> # zh
> 这是个公开的函数吗？改为私有会有问题吗？

I want distinguish queryTx & execTx in MySQL's MigrateColumnUnique, just like AutoMigrate.
So I made this func public

migrator.go
```diff
@@ -87,6 +87,8 @@ type Migrator interface {
 	DropColumn(dst interface{}, field string) error
 	AlterColumn(dst interface{}, field string) error
 	MigrateColumn(dst interface{}, field *schema.Field, columnType ColumnType) error
+	// MigrateColumnUnique migrate column's UNIQUE constraint, it's part of MigrateColumn.
+	MigrateColumnUnique(dst interface{}, field *schema.Field, columnType ColumnType) error
```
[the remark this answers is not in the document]
black-06: The advantage is that we don't need to add  `UniqueAffectedbyUniqueIndex` parameter

schema/field.go
```diff
@@ -69,6 +69,7 @@ type Field struct {
 	DefaultValueInterface  interface{}
 	NotNull                bool
 	Unique                 bool
+	UniqueIndex            string
```
a631807682, line 72: add some code comments? doesn't seem to contain mul uniqueIndex?

schema/relationship.go
```diff
@@ -577,6 +578,30 @@ type Constraint struct {
...
+	sql = "CONSTRAINT ? FOREIGN KEY ? REFERENCES ??"
+	if constraint.OnDelete != "" {
+		sql += " ON DELETE " + constraint.OnDelete
+	}
+
+	if constraint.OnUpdate != "" {
+		sql += " ON UPDATE " + constraint.OnUpdate
+	}
+
+	var foreignKeys, references []interface{}
```
a631807682, line 593: fix the format pointed out by ci

tests/migrate_test.go
```diff
@@ -1598,3 +1606,209 @@ func TestMigrateExistingBoolColumnPG(t *testing.T) {
...
+		checkMulIndex = func(t *testing.T) {
+			checkColumnType(t, "name", false)
+			checkColumnType(t, "nick_name", false)
+			if !DB.Migrator().HasIndex(table, mulIndex.Name()) {
+				t.Errorf("%v: should has index %s but not", utils.FileWithLineNum(), mulIndex.Name())
+			}
+		}
+	}
+
+	tests := []TestCase{
```
a631807682, line 1762: Comments can be added to indicate the issue associated with the test case.

a631807682 (approved):

starudream:
Are there any plans for an upcoming release?

black-06:
I found that ci of GORM and drivers are deadlocked, they will never succeed.

So I will split this mr.

black-06:
I rolled back a small part of them, otherwise this mr would never have passed the ci.
After that, I will modify Dialectors' mr, 
when they are merged, I will reopen a new mr to apply the part that was rolled back.

Guys, now that ci is successful, let's review and merge it

a631807682:
> I rolled back a small part of them, otherwise this mr would never have passed the ci. After that, I will modify Dialectors' mr, when they are merged, I will reopen a new mr to apply the part that was rolled back.
> 
> Guys, now that ci is successful, let's review and merge it

Can you rebase 9ad102b - ab75e42 to one commit? It has been too long since I approved these changes. I just want to see what has been changed after that.

black-06:
@a631807682 I have compressed them, please take time to review~

PR #6453, opened 2023-07-08
Fix begin in an exist transaction error

c5ms:
- [x] Do only one thing
- [x] Tested

Allow begin method call on a sql.Tx without a ErrInvalidTransaction error .
So  we can control the transaction manually with nested.

### User Case Description

```go
// here is an example to call begin in nested.

func f1(db *gorm.DB) {
	tx := db.Begin()
	var err error
	defer func() {
		if err != nil {
			tx.Commit()
		} else {
			tx.Rollback()
		}
	}()
}

func fr(db *gorm.DB) {
	tx := db.Begin()
	var err error
	defer func() {
		if err != nil {
			tx.Commit()
		} else {
			tx.Rollback()
		}
	}()

	f1(tx)
}

// code above is similar to :

	tx := context.Db.Begin()
	tx2 := tx.Begin()
        // now ,it's safe
	if nil != tx2.Error {
		err = tx2.Error
		panic(err)
	}

```


jinzhu:
Going to close this PR as tests failed.

PR #6467, opened 2023-07-19
add dynamic tableName support

wuxq:
Add Dynamic TableName Support

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -90,6 +90,10 @@ func (schema Schema) LookUpFieldByBindName(bindNames []string, name string) *Fie
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @wuxq 

Which is not supported intentionally, which cause lots of misuses

jinzhu:
You can use `Scopes` for dynamic table name support.

wuxq:
Scopes can't supported in  "join" or "Association"

wuxq:
@jinzhu 
Thx  your reply.

But  Scopes can't supported  when use "join" or "Association"

Is there any other solution in this situation ?

PR #6488, opened 2023-07-31
feat:add before find hook

liaosunny123:
- [x] Do only one thing
- [x] Tested

Add Before Find Hook.
I find that gorm do not implemention before find hook, but the document mentioned.


jinzhu:
We don't support BeforeFind by design. https://gorm.io/docs/hooks.html#Querying-an-object

The dest is usually an empty struct before querying, so we don't support `BeforeFind` due to it might increase misusages

PR #6502, opened 2023-08-04
fix(clause): clause equal empty array

whcao:
- [ ] Do only one thing
- [ ] Tested

when the value of `clause.Eq` is an empty array, the SQL should be `IN (NULL)`

### User Case Description
CASE: `clause.Eq{Column: column, Value: []string{}}`
ERROR: syntax error at or near \")\" (SQLSTATE 42601)

### DIFF
--- a/clause/expression.go
+++ b/clause/expression.go
@@ -246,15 +246,19 @@ func (eq Eq) Build(builder Builder) {
 
 	switch eq.Value.(type) {
 	case []string, []int, []int32, []int64, []uint, []uint32, []uint64, []interface{}:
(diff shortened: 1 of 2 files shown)


jinzhu:
Hi @whcao 

Thank you for your PR, `NOT IN (NULL)` will match nothing, so we want to raise this case as an error to avoid misunderstanding.

jinzhu:
Maybe we can only change the `Eq` expression, which looks good to me and it don't cause the misunderstanding.

Would you mind to create another pull request?

PR #6540, opened 2023-08-23
Fix handling of unknown column types

looi:
- [x] Do only one thing
- [x] Tested

When using the go clickhouse driver, scanning into sql.RawBytes is not supported for most columns. I think the database/sql standard only [guarantees](https://pkg.go.dev/database/sql#Rows.Scan) that all columns can be converted to *interface{}, not RawBytes, so gorm should not make this assumption.

This issue doesn't affect most drivers like sqlite, mysql, postgres, but it does affect clickhouse. The following test will pass once this change is merged: https://github.com/go-gorm/clickhouse/pull/128

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -257,9 +257,11 @@ func Scan(rows Rows, db *DB, mode ScanMode) {
 								continue
 							}
 						}
-						values[idx] = &sql.RawBytes{}
+						var val interface{}
+						values[idx] = &val
 					} else {
-						values[idx] = &sql.RawBytes{}
+						var val interface{}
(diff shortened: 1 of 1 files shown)


jinzhu:
This might broken query with maps? 

If it is possible to make the clickhouse driver support sql.RawBytes, I think this might be a better solution?

looi:
Why should it break maps? I think any driver would be happy to scan a map into interface{}?

looi:
I don't think changing the clickhouse driver is the right solution because the clickhouse driver isn't violating the spec

PR #6579, opened 2023-09-06
Update from

FrancoLiberali:
- [x] Do only one thing
- [x] Tested

To demonstrate that the changes in the following pull requests make sense:
https://github.com/go-gorm/sqlserver/pull/115, https://github.com/go-gorm/sqlite/pull/166, https://github.com/go-gorm/postgres/pull/213

Add a test to show that "update from" is possible without making changes to the code (by adding the clause.From manually).

This test will not work until the new versions of the dialectors are used. I don't know what flow you follow to make this update.

### User Case Description

Add an EXISTS condition to an UPDATE in a simpler way:

UPDATE table_name_1
SET attribute = "value"
FROM table_name_2
WHERE table_name_2.table_1_id = table_name_1.id

This is equivalent to:

UPDATE table_name_1
SET attribute = "value"
WHERE EXISTS (select 1 FROM table_name_2 WHERE table_name_2.table_1_id = table_name_1.id)

### DIFF
--- a/tests/update_test.go
+++ b/tests/update_test.go
@@ -882,3 +882,52 @@ func TestSaveWithHooks(t *testing.T) {
 		t.Errorf(`token content should be "token2_encrypted", but got: "%s"`, o2.Token.Content)
 	}
 }
+
+// only postgres, sqlserver, sqlite support update from
+func TestUpdateFrom(t *testing.T) {
+	if DB.Dialector.Name() != "postgres" && DB.Dialector.Name() != "sqlite" && DB.Dialector.Name() != "sqlserver" {
+		return
+	}
+
+	users := []*User{
+		GetUser("update-from-1", Config{Account: true}),
+		GetUser("update-from-2", Config{Account: true}),
+		GetUser("update-from-3", Config{}),
+	}
+
+	if err := DB.Create(&users).Error; err != nil {
+		t.Fatalf("errors happened when create: %v", err)
+	} else if users[0].ID == 0 {
+		t.Fatalf("user's primary value should not zero, %v", users[0].ID)
+	} else if users[0].UpdatedAt.IsZero() {
+		t.Fatalf("user's updated at should not zero, %v", users[0].UpdatedAt)
+	}
+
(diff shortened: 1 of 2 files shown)


tests/update_test.go
```diff
@@ -882,3 +882,52 @@ func TestSaveWithHooks(t *testing.T) {
 		t.Errorf(`token content should be "token2_encrypted", but got: "%s"`, o2.Token.Content)
 	}
 }
+
+// only postgres, sqlserver, sqlite support update from
+func TestUpdateFrom(t *testing.T) {
```
jinzhu, line 887: the test failed when running all test cases

FrancoLiberali: @jinzhu It seems that the problem is not these tests, but the new versions of the drivers that were taged and are used to run the tests in the ci (in master the tests passed because the ci was run before the new tags were created). For the postgree driver the v1.5.3 tag was created but the tests do not pass using this version (they do not pass because of the changes in the commit https://github.com/go-gorm/postgres/commit/7ba909e839c99062711d10be7f1a03f19f8d9694 specifically). In the same way, in the mysql driver the v1.5.2 tag was created but the tests don't pass since the changes of the commit https://github.com/go-gorm/mysql/commit/2a61ba01095c98f4610c8d841050ec39c04b7ea2. In my fork (https://github.com/ditrit/gorm) you can see the master tests do not pass with the new driver tags

a631807682: @FrancoLiberali The test failure has nothing to do with this PR. You seem to have investigated the cause of the failure. Are you interested in creating other PRs to fix it?

FrancoLiberali: @a631807682 To be honest, I only ran the tests on the different driver's commits to find out from which one the tests don't work, but I don't know what is the purpose of the changes in those commits to fix the generated problems

haihanj:
Is it possible you can write some sample code how to translate this SQL sentence into GORM?

FrancoLiberali:
@haihanj As you can see in the added tests, the way is to add the From clause manually:

`DB.Model(&User{}).Clauses(clause.From{Tables: []clause.Table{{Name: "accounts"}}}).Where("accounts.user_id = users.id AND accounts.number = ? AND accounts.deleted_at IS NULL", users[0].Account.Number).Update("name", "franco")`

I hope this is helpful

PR #6583, opened 2023-09-07
Update and Delete returning preload

FrancoLiberali:
- [x] Do only one thing
- [x] Tested

Be able to do preloading with an UPDATE or DELETE that does RETURNING

### DIFF
--- a/callbacks/callbacks.go
+++ b/callbacks/callbacks.go
@@ -68,6 +68,7 @@ func RegisterDefaultCallbacks(db *gorm.DB, config *Config) {
 	updateCallback.Register("gorm:before_update", BeforeUpdate)
 	updateCallback.Register("gorm:save_before_associations", SaveBeforeAssociations(false))
 	updateCallback.Register("gorm:update", Update(config))
+	updateCallback.Register("gorm:preload", Preload)
 	updateCallback.Register("gorm:save_after_associations", SaveAfterAssociations(false))
 	updateCallback.Register("gorm:after_update", AfterUpdate)
 	updateCallback.Match(enableTransaction).Register("gorm:commit_or_rollback_transaction", CommitOrRollbackTransaction)
--- a/callbacks/callbacks.go
+++ b/callbacks/callbacks.go
@@ -58,6 +58,7 @@ func RegisterDefaultCallbacks(db *gorm.DB, config *Config) {
 	deleteCallback.Register("gorm:before_delete", BeforeDelete)
 	deleteCallback.Register("gorm:delete_before_associations", DeleteBeforeAssociations)
 	deleteCallback.Register("gorm:delete", Delete(config))
+	deleteCallback.Register("gorm:preload", Preload)
 	deleteCallback.Register("gorm:after_delete", AfterDelete)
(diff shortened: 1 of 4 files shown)


saeidee:
@FrancoLiberali thanks for your contribution, I see your point of adding it for the Returning clause case but not sure if it is logical to add it for update & delete generally.

FrancoLiberali:
@saeidee In the case of not adding a returning, the preload would not affect the executed queries at all, since there is no preload to preload on. Prube:
Code:

    var result User

	log.Println("returning but empty results")
	if err := db.Model(&result).Where("name = ?", "not_exists").Clauses(clause.Returning{}).Preload("Account").Update("name", "order_name").Error; err != nil {
		log.Fatalln(err)
	}

	log.Println("no returning with empty results")
	if err := db.Model(&result).Where("name = ?", "not_exists").Preload("Account").Update("name", "order_name").Error; err != nil {
		log.Fatalln(err)
	}

	log.Println("no returning with not empty results")
	user := &User{Name: "exists"}
	if err := db.Create(user).Error; err != nil {
		log.Fatalln(err)
	}
	if err := db.Create(&Account{Name: "account", UserID: user.ID}).Error; err != nil {
		log.Fatalln(err)
	}
	if err := db.Model(&result).Where("name = ?", "exists").Preload("Account").Update("name", "exists").Error; err != nil {
		log.Fatalln(err)
	}
	log.Println(result)

	log.Println("returning with not empty results")
	if err := db.Model(&result).Where("name = ?", "exists").Clauses(clause.Returning{}).Preload("Account").Update("name", "exists").Error; err != nil {
		log.Fatalln(err)
	}
	log.Println(result)

Result:

returning but empty results
[0.669ms] [rows:0] UPDATE `users` SET `name`="order_name" WHERE name = "not_exists" RETURNING *

no returning with empty results
[0.153ms] [rows:0] UPDATE `users` SET `name`="order_name" WHERE name = "not_exists"

no returning with not empty results
[8.350ms] [rows:1] INSERT INTO `users` (`name`) VALUES ("exists") RETURNING `id`
[5.060ms] [rows:1] INSERT INTO `accounts` (`name`,`user_id`) VALUES ("account",2) RETURNING `id`
[0.235ms] [rows:2] UPDATE `users` SET `name`="exists" WHERE name = "exists"

{0 exists {0  0}}

returning with not empty results
[0.081ms] [rows:1] SELECT * FROM `accounts` WHERE `accounts`.`user_id` = 1
[0.121ms] [rows:1] INSERT INTO `accounts` (`name`,`user_id`,`id`) VALUES ("account",1,1) ON CONFLICT (`id`) DO UPDATE SET `user_id`=`excluded`.`user_id` RETURNING `id`
[0.729ms] [rows:1] UPDATE `users` SET `name`="exists" WHERE name = "exists" RETURNING *

{1 exists {1 account 1}}

a631807682 (commented):
Can you describe why update statements need to use preload? In my understanding, they are not used together.

jinzhu:
Preload is not designed works with Delete and Update, so going to close this pr. thank you for your contribution

PR #6627, opened 2023-10-08
feat: add IF NOT EXISTS config for migrator

zhtiyissr:
- [x] Do only one thing
- [x] Tested

When migrator needs to be executed concurrently, it is not enough to judge by ```HasTable()```, which will cause ```CreateTable()``` to return the ```table is already exist``` error. To avoid this error, hopefully use the ```IF NOT EXISTS``` statement. This change is forward compatible and will not cause changes to existing code.

### User Case Description
Assume that in an APP, user login, registration, browsing, click and other events need to be processed, and these events need to be recorded in the table, then the table needs to be dynamically divided according to user_id. In a distributed scenario, when browsing and clicking events are processed concurrently, When, it may cause Migrator to be abnormal.

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -35,6 +35,7 @@ type Migrator struct {
 // Config schema config
 type Config struct {
 	CreateIndexAfterCreateTable bool
+	CreateTableUseIfNotExists   bool
 	DB                          *gorm.DB
 	gorm.Dialector
 }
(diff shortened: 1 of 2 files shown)


jinzhu:
you can use AutoMigrate in this case, and this pr seems doesn't works as explained, so going to close it for now.

PR #6649, opened 2023-10-19
Adds AfterError callback hook

guilhermefbarbosa:
- [x] Do only one thing
- [x] Tested

This pull request adds an new callback hook: AfterError that should run if any other main callback returns an error.

Currently, AfterXYZ hooks like AfterCreate, AfterSave and AfterUpdate doesn't get executed when an error is found, this leads to some unnecessary (but not always a problem) levels of code coupling between applications and errors returned by gorm. 

### User Case Description

There are many possible use cases, but mainly:

1. Use AfterError hooks to translate/wrap gorm and database errors to application domain errors, decoupling other application layers from duty of knowing external errors.
2. Use AfterError to handle specific table-wise errors, like UniqueIndex errors to redirect application flow

### DIFF
--- a/callbacks/callbacks.go
+++ b/callbacks/callbacks.go
@@ -42,6 +42,7 @@ func RegisterDefaultCallbacks(db *gorm.DB, config *Config) {
 	createCallback.Register("gorm:before_create", BeforeCreate)
 	createCallback.Register("gorm:save_before_associations", SaveBeforeAssociations(true))
 	createCallback.Register("gorm:create", Create(config))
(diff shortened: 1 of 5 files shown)


guilhermefbarbosa:
i'm looking over these test failures

a631807682 (commented):
What is the difference between this feat and https://gorm.io/docs/write_plugins.html#Register-Callback-with-orders

jinzhu:
Doesn't seem necessary for me.

PR #6650, opened 2023-10-20
fix: SQL `UPDATE SET` is missing spaces

demoManito:


jinzhu:
thank you for your PR, the space is not necessary, so we removed them for performance.

PR #6662, opened 2023-10-27
map insert support return increment id

FangSqing:
- 

- [x] Do only one thing
- [x] Tested

This pull request aims to support returning the primary key value, when using a map (or map list) to call GORM to insert record, in auto-increment primary key case. 

If the primary key value is not set, the returned map or map list will be filled with a custom column "@id".

### User Case Description

The following cases are all using **auto-increment primary key**
**Case 1:**
&ensp; Input: map or map list without primary key
&ensp; Output: map or map list will be filled with "@id"
**Case 2:**
&ensp; Input: map or map list with primary key
&ensp; Output: map or map list will be filled with "@id"

The following cases are all using **non-auto-increment primary key**
**Case 1:**
&ensp; Input: map or map list without primary key
&ensp; Output: error, the same as before.
**Case 2:**
&ensp; Input: map or map list with primary key
&ensp; Output: the same as before

Last but not least, it has no impact on non-auto-increment primary keys.

### DIFF
--- a/callbacks/create.go
+++ b/callbacks/create.go
@@ -107,15 +107,47 @@ func Create(config *Config) func(db *gorm.DB) {
 			return
 		}
 
-		if db.Statement.Schema != nil {
-			if db.Statement.Schema.PrioritizedPrimaryField == nil || !db.Statement.Schema.PrioritizedPrimaryField.HasDefaultValue {
-				return
-			}
+		insertID, err := result.LastInsertId()
+		insertOk := err == nil && insertID > 0
+		if !insertOk {
+			db.AddError(err)
+			return
+		}
 
-			insertID, err := result.LastInsertId()
-			insertOk := err == nil && insertID > 0
-			if !insertOk {
-				db.AddError(err)
+		var (
+			pkField     *schema.Field
+			pkFieldName = "@id"
+		)
+
+		if db.Statement.Schema != nil && db.Statement.Schema.PrioritizedPrimaryField != nil && db.Statement.Schema.PrioritizedPrimaryField.HasDefaultValue {
+			pkField = db.Statement.Schema.PrioritizedPrimaryField
+			pkFieldName = db.Statement.Schema.PrioritizedPrimaryField.DBName
+		}
+
+		// append @id column with value for auto-increment primary key
+		// the @id value is correct, when: 1. without setting auto-increment primary key, 2. database AutoIncrementIncrement = 1
+		switch values := db.Statement.Dest.(type) {
+		case map[string]interface{}:
+			values[pkFieldName] = insertID
+		case *map[string]interface{}:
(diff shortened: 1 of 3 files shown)


callbacks/create.go
```diff
@@ -103,9 +103,15 @@ func Create(config *Config) func(db *gorm.DB) {
...
+		if db.RowsAffected == 0 {
+			return
+		}
+
+		if db.Statement.Schema != nil {
+			if db.Statement.Schema.PrioritizedPrimaryField == nil || !db.Statement.Schema.PrioritizedPrimaryField.HasDefaultValue {
+				return
+			}
+
 			insertID, err := result.LastInsertId()
```
jinzhu, line 115: In some cases,  the `Dest` might be a map value, but `db.Statement.Schema` is not nil, for example:


`db.Model(&User{}).First(map[string]interface{}{})`

jinzhu: btw, can you add some test cases for the feature, thank you.

a631807682 (commented):
Since the detection of `RETURNING` does not support map, an error will occur if the driver only supports returning but not LastInsertId (the database usually only supports one).
We need to change the detection of `RETURNING` if we want to support this feat.

related to https://github.com/go-gorm/gorm/issues/6812

sujit-baniya:
This PR creates bug for inserting rows (slice of map) in batches for Postgres

Error: `LastInsertId is not supported by this driver`

FangSqing:
Sorry, I’ll check it and fix it with a new PR.

thesiti92:
I also cannot update gorm because this breaks inserting a map of values in postgres. Thanks for working on a fix!

sujit-baniya:
For the time being, till the issue is fixed, I've been checking for the error message.

> https://github.com/go-gorm/gorm/compare/master...sujit-baniya:gorm:master

zh929:
if I use createForUpdate logic，and the same logic is executed simultaneously in multiple databases，if not set Model, the same map will add “@id”，causing some SQL statements to fail execution. For example:
`
        updateList := make([]map[string]interface{}, 0)  // the map will be add '@id'
        columns := make([]clause.Column, 0)
        updateColumns := make([]string, 0)
        wg := new(sync.WaitGroup)
	wg.Add(len(req.dbList))
	for _, v := range req.dbList {
		go func(dbName string) {
			defer wg.Done()
			db := global.GVA_DBList[dbName]
			err := db.Table(req.table).Clauses(clause.OnConflict{
				Columns:   columns,                                
				DoUpdates: clause.AssignmentColumns(updateColumns), 
			}).Create(&updateList).Error
			if err != nil {
				errDbList = append(errDbList, dbName)
				isAllDone = false
			}
		}(v)
	}
	wg.Wait()
`
Error Msg: Error 1054 (42S22): Unknown column '@id' in 'field list'

PR #6676, opened 2023-11-07
fix(logger): Fixed `fc` is nil

flc1125:
- [x] Do only one thing
- [ ] Tested

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -153,6 +153,10 @@ func (l *logger) Error(ctx context.Context, msg string, data ...interface{}) {
 //
 //nolint:cyclop
 func (l *logger) Trace(ctx context.Context, begin time.Time, fc func() (string, int64), err error) {
+	if fc == nil {
+		return
+	}
+
 	if l.LogLevel <= Silent {
 		return
 	}


shedyfreak (approved):
lftm, although I would suggest 
l.Printf(l.traceWarnStr, utils.FileWithLineNum(), err, float64(elapsed.Nanoseconds())/1e6, "-", "-")
before returning

flc1125:
> lftm, although I would suggest l.Printf(l.traceWarnStr, utils.FileWithLineNum(), err, float64(elapsed.Nanoseconds())/1e6, "-", "-") before returning

If I return from there, I have code snippets of my switch, almost every one of them, will that cause redundant confusion.

jinzhu:
Can you let me know why this is needed? in which cases the `fc` will be nil when using GORM.

flc1125:
> Can you let me know why this is needed? in which cases the `fc` will be nil when using GORM.

I was writing unit tests when I discovered the problem and fixed it together.

jinzhu:
If we can't reproduce this when using GORM's API directly, I think it is safe to keep it now to avoid unnecessary condition check.

Thank you.

PR #6678, opened 2023-11-07
Allow filtering/hiding SQL params when using a custom logger implementation

shedyfreak:
- [X] Do only one thing
- [X] Tested

adding a function to interface definition

### User Case Description

I have my own implementation of the logger interface, but in my logging I want to hide params when logging in PROD

### More

I didn't find a way to set the ParameterizedQueries in config when using my own Logger implementation.


shedyfreak:
I just saw that params filter is an interface in it self, thank you for the good design I can handle it without this PR

PR #6700, opened 2023-11-20
feat: order supports []clause.OrderByColumn

lucasbbb:
- [x] Do only one thing
- [x] Tested

This PR supports the []clause.OrderByColumn as input for the Order function.

### User Case Description

db.Order([]clause.OrderByColumn{{Column: clause.Column{Name: "name"}, Desc: true}})

### DIFF
--- a/chainable_api.go
+++ b/chainable_api.go
@@ -299,6 +299,7 @@ func (db *DB) Having(query interface{}, args ...interface{}) (tx *DB) {
 //
 //	db.Order("name DESC")
 //	db.Order(clause.OrderByColumn{Column: clause.Column{Name: "name"}, Desc: true})
+//	db.Order([]clause.OrderByColumn{{Column: clause.Column{Name: "name"}, Desc: true}})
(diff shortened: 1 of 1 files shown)


a631807682 (commented):
Thanks for the contribution, but it doesn't seem necessary, we can use `.Order(...).Order(...)` to achieve it.

lucasbbb:
> Thanks for the contribution, but it doesn't seem necessary, we can use `.Order(...).Order(...)` to achieve it.

Let me describe my usage scenario:

I control the order in which the restful api returns results with a single parameter, such as `GET /users?sort_by=-last_modified,+email`. Then we use a generic method to process this parameter as `[]clause.OrderByColumn{...}}`.

If the `.Order()` doesn't support `[]clause.OrderByColumn{}` as input, I had to use a for loop, which broke the chain.
```
for _, ele := range orders {
    db.Order(ele)
}
```
```
db.Order(orders)
```

PR #6766, opened 2023-12-27
fix: ExplainSQL using consecutive pairs of escaper in SQL string represents an escaper

iTanken:
- [x] Do only one thing
- [x] Tested

[`logger.ExplainSQL`](https://github.com/iTanken/gorm/blob/dc066a1d5637a60eb25d897411753b837b3a2b4f/logger/sql.go#L38) using consecutive pairs of escaper in SQL string represents an escaper, preventing it from being interpreted as the string terminator. This is a widely used escape mechanism in SQL standards and is applicable in most relational databases.

https://github.com/go-gorm/postgres/issues/236

### User Case Description

https://github.com/go-gorm/playground/pull/677


iTanken:
@jinzhu Please review this PR when you have time.

Issue #6755, opened 2023-12-18
Composite Primary Key should understand field priority like unique index


zeevt-at-jfrog:
## Describe the feature

Composite Primary Key should understand field priority like unique index.

https://github.com/go-gorm/playground/pull/674

## Motivation

The order of the fields in the index is important, if I can have a composite primary key I want to be able to control it like I can for unique indexes

PR #6784, opened 2024-01-07
feat: allow priority tag in composite PK

yaofeng-wang:
- [x] Do only one thing
- [x] Tested

Fixes [issue#6755](https://github.com/go-gorm/gorm/issues/6755) to allow priority tag for composite primary key, similar to unique index.

### User Case Description

```
type Bad struct {
	A   ItemA
	AId uint `gorm:"primaryKey,priority:2;autoIncrement:false"`
	B   ItemB
	BId uint `gorm:"primaryKey,priority:1;autoIncrement:false"`
}
```
Composite primary key will be created and it follows ordering specified by priority tag:
image

### DIFF
--- a/schema/field.go
+++ b/schema/field.go
@@ -127,16 +127,21 @@ func (schema *Schema) ParseField(fieldStruct reflect.StructField) *Field {
 	}
 
 	if field.PrimaryKey {
+		// default priority is 10, to be consistent with composite index
 		field.PrimaryKeyPriority = 10
-	}
 
-	if field.TagSettings["PRIMARYKEY,PRIORITY"] != "" {
-		primaryKeyPriority, err := strconv.Atoi(tagSetting["PRIMARYKEY,PRIORITY"])
-		if err == nil {
-			field.PrimaryKeyPriority = primaryKeyPriority
+		priority := ParseTagSetting(field.TagSettings["PRIMARYKEY"], ",")["PRIORITY"]
+		if priority == "" {
+			priority = ParseTagSetting(field.TagSettings["PRIMARY_KEY"], ",")["PRIORITY"]
+		}
+
+		if priority != "" {
+			primaryKeyPriority, err := strconv.Atoi(priority)
+			if err == nil {
+				field.PrimaryKeyPriority = primaryKeyPriority
+			}
 		}
 	}
-
 	for field.IndirectFieldType.Kind() == reflect.Ptr {
 		field.IndirectFieldType = field.IndirectFieldType.Elem()
 	}
--- a/schema/field.go
+++ b/schema/field.go
@@ -59,6 +59,7 @@ type Field struct {
 	DataType               DataType
 	GORMDataType           DataType
 	PrimaryKey             bool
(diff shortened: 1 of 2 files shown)


schema/field.go
```diff
@@ -103,19 +104,20 @@ func (schema *Schema) ParseField(fieldStruct reflect.StructField) *Field {
...
+		IndirectFieldType: fieldStruct.Type,
+		StructField:       fieldStruct,
+		Tag:               fieldStruct.Tag,
+		TagSettings:       tagSetting,
+		Schema:            schema,
+		Creatable:         true,
+		Updatable:         true,
+		Readable:          true,
+		PrimaryKey: utils.CheckTruth(tagSetting["PRIMARYKEY"], tagSetting["PRIMARY_KEY"],
+			tagSetting["PRIMARYKEY,PRIORITY"]),
```
a631807682, line 120: Thank you for your contribution, can we keep the style of tags consistent?

https://github.com/go-gorm/gorm/blob/master/schema/index_test.go#L16

schema/field.go
```diff
@@ -124,6 +126,17 @@ func (schema *Schema) ParseField(fieldStruct reflect.StructField) *Field {
 		AutoIncrementIncrement: DefaultAutoIncrementIncrement,
 	}
 
+	if field.PrimaryKey {
+		field.PrimaryKeyPriority = 10
```
a631807682, line 130: Does this have any special meaning? Why is it 10?

yaofeng-wang: https://gorm.io/docs/indexes.html#Fields-Priority

> You can specify the order with the priority option, the default priority value is 10, if priority value is the same, the order will be based on model struct’s field index

The value 10 is intended to be the default value, similar to composite index.

schema/field.go
yaofeng-wang: Do you mean that the key for `tagSetting` should be `PRIMARYKEY` and `PRIORITY` instead of `PRIMARY,PRIORITY` ?

a631807682: `gorm:"primaryKey:,priority:1"`

yaofeng-wang: I've updated the parsing logic according to the tag format. I've also added a comment for the default priority value.

a631807682:
```go
PrimaryKey: utils.CheckTruth(tagSetting["PRIMARYKEY"], tagSetting["PRIMARY_KEY"],
			tagSetting["PRIMARYKEY,PRIORITY"])
```
`tagSetting["PRIMARYKEY,PRIORITY"]` Is it unnecessary?

jinzhu:
Why not use a unique index? Doesn't it seem more intuitive?

yaofeng-wang:
> Why not use a unique index? Doesn't it seem more intuitive?

I think it's better to be able to choose between composite primary key and unique index.

jinzhu:
We're planning to close this PR because we aim to minimize the complexity introduced by some edge cases. If necessary, we can ensure that this field is ordered in the model definition.

Thank you for your PR.

Issue #6799, opened 2024-01-20
endless loop in Save


omniskop:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/684

## Description

Calling Save with a pointer to an interface{} results in an endless loop.

This is due to it trying to dereference the interfacing using reflect.Indirect which only works with pointers.
[See finisher_api.go:78](https://github.com/go-gorm/gorm/blob/0123dd45094295fade41e13550cd305eb5e3a848/finisher_api.go#L78)

I tried two different approaches to fix this:

### Solution 1: Return an error

Change the loop to not dereference interfaces. This will result in an "unsupported data" error down the line.
This can be an acceptable solution and is better than an endless loop.

### Solution 2: Support it

Use reflectValue.Elem() instead of reflect.Indirect() which works with interfaces and pointers.
Just doing that will also result in an "unsupported data" error though.
This is due to [callbacks.go](https://github.com/go-gorm/gorm/blob/0123dd45094295fade41e13550cd305eb5e3a848/callbacks.go#L117) only supporting pointers and no interfaces.
I changed that place to also accept interfaces but the code later tries to use CanAddr on the interface and that doesn't work. All of this is due to it trying to modify the data passed to Save.

I don't know enough about this code to try to implement solution 2 I have thus created a Pull Request for solution 1. If that should not be the desired solution feel free to reject it and alternatives can be discussed here.

PR #6800, opened 2024-01-20
fix: endless loop in save

omniskop:
- [X] Do only one thing
- [X] Tested

Save attempts to dereference an interface using reflect.Indirect this doesn't work as that is only intended for pointers.
This change stops it from trying to do that. Trying to do this will instead result in an error.

I don't think this should cause any breaking changes, but I don't know why this code was written like.

This would close Issue #6799 but if another solution for the problem would be preferred feel free to close this pull request.

### User Case Description

While it might not be a supported case to call Save with a pointer to an interface it should not cause an endless loop.

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -75,7 +75,7 @@ func (db *DB) Save(value interface{}) (tx *DB) {
 	tx.Statement.Dest = value
 
 	reflectValue := reflect.Indirect(reflect.ValueOf(value))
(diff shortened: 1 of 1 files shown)


a631807682 (commented):
Thanks for your contribution, we are sure this is a bug and it comes from https://github.com/go-gorm/gorm/commit/0729261b627d0f73ab0e9bccc5b548d5e55fae88, can you add some tests for it?

daheige (approved):
lgtm!

PR #6806, opened 2024-01-26
let limit and offset use bind parameter

jasonchuan:
- [ ] Do only one thing
- [ ] Tested

 two  changes for solving the limt and offset problem which is  possible to cause  continuously increasing of  prepared_stmt_count of mysql , max_prepared_stmt_count  value range  is 0 - 1048576 and  default 16382.

### User Case Description

the caller used  while true ugly code to call our api , our api has bugs for no checking if the limit and offset is below the  total row number  . limit and offset was going far away from the row count . a number of sql would be  prepared in mysql ,much more than  max_prepared_stmt_count  .
FATAL: MySQL error: 1461 "Can't create more than max_prepared_stmt_count statements (current value: 1048576 )"


jinzhu:
Hi @jasonchuan 

thank you for your pr, can you please fix those failed tests?

jasonchuan:
> Hi @jasonchuan
> 
> thank you for your pr, can you please fix those failed tests?



> Hi @jasonchuan
> 
> thank you for your pr, can you please fix those failed tests?

ok  I'll fix those later!

jasonchuan:
@jinzhu got  the job done , please approve the postgress commit (https://github.com/go-gorm/postgres/pull/239) first.  because the stmt.Vars is polluted  by  gorm.io/driver/postgres@v1.5.4/migrator.go:598 ,we must get the vars ok for bindvar method of postgress driver  first.

a631807682 (approved):

tests/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/jinzhu/now v1.1.5
 	github.com/lib/pq v1.10.9
 	gorm.io/driver/mysql v1.5.2
-	gorm.io/driver/postgres v1.5.4
+	gorm.io/driver/postgres v1.5.5
```
saeidee, line 10: This tag has been deleted FYI

jasonchuan: rollbacked

saeidee (approved):

haunt98:
Hi @jasonchuan, thanks for PR. But can I ask one more thing? Is the change `LIMIT 123` to `LIMIT ?` helps MySQL checking with total row limit? Or do I need to check it also in layer above db?

PR #6828, opened 2024-02-06
feat: dryrun migration should only print not-select sqls

shushenghong:
some driver migration may run select querys to compare schemal change, like postgresql. we should only print not-select sqls when dry run migration

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -47,7 +47,9 @@ type printSQLLogger struct {
 
 func (l *printSQLLogger) Trace(ctx context.Context, begin time.Time, fc func() (sql string, rowsAffected int64), err error) {
 	sql, _ := fc()
-	fmt.Println(sql + ";")
+	if len(sql) < 6 || strings.ToUpper(sql[0:6]) != "SELECT" {
+		fmt.Println(sql + ";")
+	}
 	l.Interface.Trace(ctx, begin, fc, err)
 }
 


migrator/migrator.go
```diff
@@ -47,7 +47,9 @@ type printSQLLogger struct {
 
 func (l *printSQLLogger) Trace(ctx context.Context, begin time.Time, fc func() (sql string, rowsAffected int64), err error) {
 	sql, _ := fc()
-	fmt.Println(sql + ";")
+	if len(sql) < 6 || strings.ToUpper(sql[0:6]) != "SELECT" {
+		fmt.Println(sql + ";")
```
a631807682, line 51: The select statement is executed by queryTx and will not be printed. The logs you see are all printed by Trace. https://github.com/go-gorm/gorm/blob/master/migrator/migrator.go#L122
In fact, I don't agree with using fmt.Println here. It is equivalent to searching for the record I need in a lot of logs, and it may be inconsistent with the output of the original log settings, and level control cannot be used.
I think we can provide log implementation for migration through documentation, such as writing the change log to a file through the log.

shushenghong: postgresql dry run mode still need to run many select method and will printout by this method

jinzhu: https://github.com/go-gorm/gorm/blob/3e2c4fc446f0a601124771d012ee167cbb0d7de0/migrator/migrator.go#L112-L114 It is expected to work in dry run mode.

PR #6837, opened 2024-02-13
Fix null setting on reused objects

turip:
- [x] Do only one thing
- [ ] Tested

Starting from v1.25.1 GORM seems to not reset null fields, causing incosistent behavior compared to the previous way GORM was working.

The fixes I have added are fixing this for Scanner interfaces and time.Time, however there's still a test failure in case of the int and most likely other primitive variables.

I am opening this PR just to illustrate the fixes, but I think we need to debate it further what's the expected behavior.

### DIFF
--- a/tests/scan_test.go
+++ b/tests/scan_test.go
@@ -1,10 +1,12 @@
 package tests_test
 
 import (
+	"fmt"
 	"reflect"
 	"sort"
 	"strings"
 	"testing"
+	"time"
 
 	"gorm.io/gorm"
 	. "gorm.io/gorm/utils/tests"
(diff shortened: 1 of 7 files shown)


turip:
The regression seems to be coming from this PR: https://github.com/go-gorm/gorm/pull/6311

The unit tests are still failing on the integer default value setting. I can dig that out too @jinzhu, but this partially reverts #6311. What do you think about this fix?

tests/scan_test.go
```diff
@@ -240,3 +242,62 @@ func TestScanToEmbedded(t *testing.T) {
...
+	test := NullValue{}
+	for rows.Next() {
+		if err := DB.ScanRows(rows, &test); err != nil {
+			t.Fatalf("cannot scan nullvalues: %v", err)
+		}
+
+		if test.ID%2 == 0 {
+			AssertEqual(t, test.ScannerValue, NewDummyString(""))
+			AssertEqual(t, test.NullScannerValue, nil)
+			AssertEqual(t, test.IntValue, int(0))
```
turip, line 295: This is still problematic:
```
-- FAIL: TestScanNilHandling (9.77s)
    utils.go:40: /Users/turip/src/gorm/tests/scan_test.go:295: expect: 0, got 1
    utils.go:40: /Users/turip/src/gorm/tests/scan_test.go:295: expect: 0, got 3
    utils.go:40: /Users/turip/src/gorm/tests/scan_test.go:295: expect: 0, got 5
    utils.go:40: /Users/turip/src/gorm/tests/scan_test.go:295: expect: 0, got 7
```

jinzhu:
This PR is being closed due to failed tests.

henryjcee:
Just came here to say that I've also concluded that this is problem for us and it's not 100% clear to me what the previous PR was trying to solve. Will try and get another PR up to fix this but will also make sure I understand completely what's going on first.

PR #6844, opened 2024-02-22
Fix regression in db.Not introduced in v1.25.6.

### LINKED ISSUE #6845: db.Not method when used with multiple Where conditions has changed unexpectedly
## GORM Playground Link

- https://github.com/go-gorm/playground/pull/695
- https://github.com/go-gorm/playground/pull/696

## Description

A regression introduced in Gorm v1.25.6, specifically in commit https://github.com/go-gorm/gorm/commit/940358e0ddd1e6d5e893184943a4fa71a0c28521.

The behavior of the db.Not method when used with multiple Where conditions has changed unexpectedly. Prior to v1.25.6, the query `db.Not(db.Where("a=1").Where("b=1"))` would correctly translate to SQL as `NOT (a=1 AND b=1)`. However, after the specified commit, the same query started translating to `(NOT a=1 AND NOT b=1)`, which is a significant alteration in the logic.

I created PR (https://github.com/go-gorm/gorm/pull/6844) to address this issue.

tsuba3:
- [x] Do only one thing
- [x] Tested

Fix #6845

Address a regression introduced in Gorm v1.25.6, specifically in commit 940358e.

The behavior of the `db.Not` method when used with multiple `Where` conditions has changed unexpectedly. Prior to v1.25.6, the query `db.Not(db.Where("a=1").Where("b=1"))` would correctly translate to SQL as `NOT (a=1 AND b=1)`. However, after the specified commit, the same query started translating to `(NOT a=1 AND NOT b=1)`, which is a significant alteration in the logic.

This PR reverts the `db.Not` method's behavior back to its original state when *gorm.DB passed as arguments.

### DIFF
--- a/clause/where.go
+++ b/clause/where.go
@@ -21,11 +21,11 @@ func (where Where) Name() string {
 
 // Build build where clause
 func (where Where) Build(builder Builder) {
-    if len(where.Exprs) == 1 {
-        if andCondition, ok := where.Exprs[0].(AndConditions); ok {
-           where.Exprs = andCondition.Exprs
-        }
-    }
+	if len(where.Exprs) == 1 {
+		if andCondition, ok := where.Exprs[0].(AndConditions); ok {
+			where.Exprs = andCondition.Exprs
+		}
+	}
 
 	// Switch position if the first query expression is a single Or condition
 	for idx, expr := range where.Exprs {
@@ -152,11 +152,6 @@ func (or OrConditions) Build(builder Builder) {
 func Not(exprs ...Expression) Expression {
 	if len(exprs) == 0 {
 		return nil
-	}
-	if len(exprs) == 1 {
-		if andCondition, ok := exprs[0].(AndConditions); ok {
-			exprs = andCondition.Exprs
-		}
 	}
 	return NotConditions{Exprs: exprs}
 }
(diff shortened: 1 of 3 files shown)


clause/where.go
```diff
@@ -166,19 +166,58 @@ type NotConditions struct {
...
+		if _, ok := c.(NegationExpressionBuilder); ok {
+			anyNegationBuilder = true
+			break
+		}
 	}
 
-	for idx, c := range not.Exprs {
-		if idx > 0 {
-			builder.WriteString(AndWithSpace)
+	if anyNegationBuilder {
```
a631807682, line 177: Do we need to distinguish negative expressions when not.Exprs length is greater than 1? I think the following statements should be the same?
```
	DB.Not(DB.Where("a=1").Where("b=1"))
	DB.Not(DB.Where("a=1").Where(clause.Eq{Column: "b", Value: 1}))
```

tsuba3: I agree with your point of view, but implementing it that way would result in compatibility issues with the following code:
`DB.Not(&Table{A: 1, B: 1})`
This code, even in versions prior to v1.25.5, results in WHERE (a <> 1 AND b <> 2). Therefore, while your proposed logic seems correct, we need to consider backward compatibility with existing code implementations.

a631807682: I got it, can you add some test about this case ?

a631807682 (approved):

PR #6846, opened 2024-02-22
CHORE add unittest test function  ConvertMapToValueForCreate

naruchet:
Adds unit tests for ConvertMapToValuesForCreate function in GORM, ensuring correct value conversion from maps to database columns for CREATE operations.


a631807682 (approved):

a631807682:
This test and the original `visit_map_test.go` can be moved to `helper_test.go`

a631807682:
> > This test and the original `visit_map_test.go` can be moved to `helper_test.go`
> 
> @a631807682 Hi
> 
> The function loadOrStoreVisitMap in the callbacks package is a private function. This means that it can only be accessed and used by code within the same package.
> 
> If you move the unit test for this function from `visite_map_test.go` to `helper_test.go`, the unit test will no longer be able to access the function because it will be in a different package.
> 
> do you want to move only `convert_map_test.go` first

Sorry, I didn't make it clear, I meant move to `callbacks/helper_test.go` .

naruchet:
> > > This test and the original `visit_map_test.go` can be moved to `helper_test.go`
> > 
> > 
> > @a631807682 Hi
> > The function loadOrStoreVisitMap in the callbacks package is a private function. This means that it can only be accessed and used by code within the same package.
> > If you move the unit test for this function from `visite_map_test.go` to `helper_test.go`, the unit test will no longer be able to access the function because it will be in a different package.
> > do you want to move only `convert_map_test.go` first
> 
> Sorry, I didn't make it clear, I meant move to `callbacks/helper_test.go` .

@a631807682
Hi. i already move the test cases located in the files convert_map_test.go and visit_map_test.go into the file helper_test.go.

a631807682 (approved):

PR #6855, opened 2024-02-28
fix(create): fix insert column order

archever:
- [x] Do only one thing
- [x] Tested

make insert sql column order stable for unittest

### User Case Description
gorm create() interface generated sql not stable for each test run. I changed the map interater to a stable slice in order to keep the column order.

### DIFF
--- a/callbacks/create_test.go
+++ b/callbacks/create_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 	"gorm.io/gorm/schema"
@@ -23,7 +22,10 @@ func TestConvertToCreateValues_DestType_Slice(t *testing.T) {
 	}
 
 	s, err := schema.Parse(&user{}, schemaCache, schema.NamingStrategy{})
-	assert.NoError(t, err)
+	if err != nil {
+		t.Errorf("parse schema error: %v, is not expected", err)
+		return
+	}
 	dest := []*user{
 		{
 			ID:    1,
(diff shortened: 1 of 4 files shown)


a631807682 (commented):
Can you add some unit tests to describe what is being fixed?

archever:
> Can you add some unit tests to describe what is being fixed?

add a ut case

a631807682 (commented):
The order of `defaultValueFieldsHavingValue` elements is based on `FieldsWithDefaultDBValue`, I don't understand why the order needs to be maintained here again, and the unit test passed before the change, can you describe why we need it?

archever:
the type of defaultValueFieldsHavingValue is a map. before the change, column values is appended by iterating a map, which is not in a stable order. after the change,  column values is appended following the order of FieldsWithDefaultDBValue, which is in  stable order.

a631807682 (approved):

jinzhu:
Hi @archever,

Thank you for your PR. Could you please remove those dependencies in go.mod? We aim to keep GORM's dependencies clean and tidy.

```
	github.com/stretchr/testify v1.8.4
)

require (
	github.com/davecgh/go-spew v1.1.1 // indirect
	github.com/pmezard/go-difflib v1.0.0 // indirect
	gopkg.in/yaml.v3 v3.0.1 // indirect
)
```

archever:
@jinzhu done

callbacks/create_test.go
```diff
@@ -0,0 +1,71 @@
...
+		ID    int `gorm:"primaryKey"`
+		Name  string
+		Email string `gorm:"default:(-)"`
+		Age   int    `gorm:"default:(-)"`
+	}
+
+	s, err := schema.Parse(&user{}, schemaCache, schema.NamingStrategy{})
+	if err != nil {
+		t.Errorf("parse schema error: %v, is not expected", err)
+		return
```
demoManito, line 27: use `t.Fatalf`

PR #6876, opened 2024-03-06
fix: trans error rollback

### LINKED ISSUE #6826: Rollback logic error in "func (db *DB) Transaction"
## GORM Playground Link

Reference issue - playground link to follow

## Description

In `finisher_api.go`:

```go
    644    defer func() {
    645    // Make sure to rollback when panic, Block error or Commit error
    646        if panicked || err != nil {
    647        tx.Rollback()
    648        }
    649    }()
    650
    651    if err = fc(tx); err == nil {
    652        panicked = false
    653        return tx.Commit().Error
    654    }
```

Function `tx.Rollback()` is not called in case of failure on line 653.

a631807682:
- [ ] Do only one thing
- [ ] Tested

close #6826

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -650,7 +650,8 @@ func (db *DB) Transaction(fc func(tx *DB) error, opts ...*sql.TxOptions) (err er
 
 		if err = fc(tx); err == nil {
 			panicked = false
-			return tx.Commit().Error
+			err = tx.Commit().Error
+			return
 		}
 	}
 


jinzhu:
If a commit fails, is a rollback needed? Does it automatically rollback?

iTanken:
> If a commit fails, is a rollback needed? Does it automatically rollback?

https://github.com/go-gorm/gorm/blob/e4e23d26d2d16cbfb575da0bfa9a400e2ad29b92/finisher_api.go#L638-L655

`// Make sure to rollback when panic, Block error or Commit error `

jinzhu:
> `// Make sure to rollback when panic, Block error or Commit error `

We code for reality, not for comments ;)

After doing some research, I think that modern databases no longer require calling rollback after a failed commit. We might need to revise the comment above.

a631807682:
Commit are usually divided into writing **commit packet** and **reading results**. If writing fails, we not need to rollback. Failure of reading results usually comes from network errors, which means our rollback may not succeed.
This does not seem to be necessary, is there any actual case?

PR #6880, opened 2024-03-07
fix(scan.go): reflect.MakeSlice may pass in the wrong type reflect.Array

demoManito:
- [x] Do only one thing
- [x] Tested

reflect.MakeSlice can only initialize slices

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -274,12 +274,14 @@ func Scan(rows Rows, db *DB, mode ScanMode) {
 
 			if !update || reflectValue.Len() == 0 {
 				update = false
-				// if the slice cap is externally initialized, the externally initialized slice is directly used here
-				if reflectValue.Cap() == 0 {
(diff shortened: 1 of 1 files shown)


scan.go
```diff
@@ -274,12 +274,14 @@ func Scan(rows Rows, db *DB, mode ScanMode) {
 
 			if !update || reflectValue.Len() == 0 {
 				update = false
-				// if the slice cap is externally initialized, the externally initialized slice is directly used here
-				if reflectValue.Cap() == 0 {
-					db.Statement.ReflectValue.Set(reflect.MakeSlice(reflectValue.Type(), 0, 20))
-				} else if !isArrayKind {
-					reflectValue.SetLen(0)
-					db.Statement.ReflectValue.Set(reflectValue)
```
demoManito, line 282: If it is an array, this logic should not be used

a631807682 (approved):

PR #6910, opened 2024-03-18
fix: behavior when getting removed callback

snackmgmg:
- [x] Do only one thing
- [x] Tested

Resolved the issue where `Get()` retrieves a callback that has been removed by `Remove()`. 
This was due to the search continuing even when a callback with `v.removed: true` was found, eventually reaching another callback of the same name with `v.removed: false`.

### User Case Description

If a callback named `cb1` is registered, it is expected that calling `Get("cb1")` after `Remove("cb1")` should return `nil`. However, a callback function registered as `cb1` is being returned instead.

```golang
db, _ := gorm.Open(tests.DummyDialector{}, nil)
createCallback := db.Callback().Create()

_ = createCallback.Register("cb1", func(*gorm.DB){})
_ = createCallback.Remove("cb1")

cb := createCallback.Get("cb1") // not nil
```

### DIFF
--- a/tests/callbacks_test.go
+++ b/tests/callbacks_test.go
@@ -212,7 +212,6 @@ func TestCallbacksRemoveAndGet(t *testing.T) {
 	createCallback := db.Callback().Create()
 
 	err := createCallback.Register("plugin_1_fn1", c1)
-
 	if err != nil {
 		t.Errorf("callbacks tests failed, got error: %v", err)
 	}
--- a/tests/callbacks_test.go
+++ b/tests/callbacks_test.go
@@ -206,3 +206,25 @@ func TestPluginCallbacks(t *testing.T) {
 		t.Errorf("callbacks tests failed, got %v", msg)
 	}
 }
+
+func TestCallbacksRemoveAndGet(t *testing.T) {
+	db, _ := gorm.Open(nil, nil)
+	createCallback := db.Callback().Create()
+
+	err := createCallback.Register("plugin_1_fn1", c1)
+
+	if err != nil {
+		t.Errorf("callbacks tests failed, got error: %v", err)
+	}
+	if cb := createCallback.Get("plugin_1_fn1"); reflect.DeepEqual(cb, c1) {
+		t.Errorf("callbacks tests failed, got: %p, want: %p", cb, c1)
+	}
+
+	err = createCallback.Remove("plugin_1_fn1")
+	if err != nil {
+		t.Errorf("callbacks test failed, got error: %v", err)
+	}
+	if cb := createCallback.Get("plugin_1_fn1"); cb != nil {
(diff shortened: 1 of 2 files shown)


demoManito (approved):

tests/callbacks_test.go
```diff
@@ -206,3 +206,25 @@ func TestPluginCallbacks(t *testing.T) {
 		t.Errorf("callbacks tests failed, got %v", msg)
 	}
 }
+
+func TestCallbacksRemoveAndGet(t *testing.T) {
+	db, _ := gorm.Open(nil, nil)
+	createCallback := db.Callback().Create()
+
+	err := createCallback.Register("plugin_1_fn1", c1)
+
```
demoManito, line 215: Hello, this blank line can be deleted?

snackmgmg: @demoManito Hello, thank you for the review. I have removed the blank line.

a631807682 (commented):
Obtaining key values based on order does not solve this problem, because the deleted element may be at the front or at the back.

Failed case
```diff	
- err := createCallback.Register("plugin_1_fn1", c1)
+ err := createCallback.Before("*").Register("plugin_1_fn1", c1)
```

I don't know the history of this part of the code, and I don't understand why the deleted callback is not removed when `compiling`.

jinzhu:
previously, the callbacks system allowed a new callback to overwrite a previously deleted one, which is no longer supported now~

jinzhu:
maybe we can remove the callback from `processor.callbacks` when call the `Remove` method

snackmgmg:
Thank you for your review. You are right, I overlooked that part. 
Additionally, after reviewing your comments, I realized that the `Remove()` is not working properly either. 
For instance, it seems that callbacks are not being removed in certain cases.

```golang
func TestRemoveNotWorking(t *testing.T) {
	db, _ := gorm.Open(nil, nil)
	rawCallback := db.Callback().Raw()

	rawCallback.Before("*").Register("cb1", func(*gorm.DB) { panic("Remove does not work") })
	rawCallback.Remove("cb1")

	db.Exec("SELECT 1") // panic
}

// In this case, it works correctly.
func TestRemoveWorking(t *testing.T) {
	db, _ := gorm.Open(nil, nil)
	rawCallback := db.Callback().Raw()

	rawCallback.Before("*").Register("cb1", func(*gorm.DB) { panic("Remove does not work") })
	rawCallback.Before("*").Remove("cb1")

	db.Exec("SELECT 1") // not panic
}
```

I think that adopting the method of removing callbacks that have been marked for removal is not only a solution for the `Get()` issue but also addresses the problem with `Remove()`. 
I will try this approach. Thank you!

snackmgmg:
Sorry, I forgot to close this.

PR #6916, opened 2024-03-19
fix: remove `callback` from `callbacks` if `Remove()` called

snackmgmg:
- [x] Do only one thing
- [x] Tested

ref: https://github.com/go-gorm/gorm/pull/6910#pullrequestreview-1945116356

Modified the `Remove()` method so that it not only sets the removed flag but also removes the `callback` from the `callbacks` within `compile()`. 
This resolves the following issues:

- The problem where the `Get()` method could retrieve callbacks that have already been removed.
- The problem where removed callbacks could still operate due to the `Remove()` method.

In addition, I have made some modifications to the existing tests.
Previously, the specified `before` and `after` for a `callback` continued to affect the overall order even after being `Remove()`d. I believe that continuing to have an effect after being removed is an unintended behavior.
What do you think? I would appreciate any advice.

#### The problem where the `Get()` method

If a callback named `cb1` is registered, it is expected that calling `Get("cb1")` after `Remove("cb1")` should return `nil`. However, a callback function registered as `cb1` is being returned instead.

```golang
db, _ := gorm.Open(nil, nil)
createCallback := db.Callback().Create()

createCallback.Register("cb1", func(*gorm.DB){})
createCallback.Remove("cb1")

cb := createCallback.Get("cb1") // not nil
```

#### The problem where the `Remove()` method

The expectation is that callbacks marked as removed via `Remove()` should not be executed. However, we encountered an issue where callbacks registered through methods like `Before("*").Register()` continue to operate even after being `Remove()`d. 
This typically occurs in scenarios such as the following:

```golang
db, _ := gorm.Open(nil, nil)
rawCallback := db.Callback().Raw()

rawCallback.Before("*").Register("cb1", func(*gorm.DB) { panic("Remove does not work") })
rawCallback.Remove("cb1")

db.Exec("SELECT 1") // panic
```

#### workaround

For your reference, using `Before("*").Remove()` can be mentioned as a workaround for the above issue.

### DIFF
--- a/callbacks.go
+++ b/callbacks.go
@@ -186,21 +186,19 @@ func (p *processor) Replace(name string, fn func(*DB)) error {
 }
 
 func (p *processor) compile() (err error) {
-	var (
-		callbacks []*callback
-		removed   []string
-	)
+	var callbacks []*callback
+	removedMap := map[string]bool{}
 	for _, callback := range p.callbacks {
 		if callback.match == nil || callback.match(p.db) {
 			callbacks = append(callbacks, callback)
 		}
 		if callback.remove {
-			removed = append(removed, callback.name)
+			removedMap[callback.name] = true
 		}
 	}
 
-	if len(removed) > 0 {
-		callbacks = removeCallbacks(callbacks, removed)
+	if len(removedMap) > 0 {
+		callbacks = removeCallbacks(callbacks, removedMap)
 	}
 
 	p.callbacks = callbacks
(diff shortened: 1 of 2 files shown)


a631807682 (commented):
Maybe we can reduce the number of loops through `map` ?

snackmgmg:
Thanks! I have improved the part you mentioned in your comment.
https://github.com/go-gorm/gorm/pull/6916/commits/f255f362be9ef06b868a8cb3f6a99779b509bf1e

a631807682 (approved):

Issue #6898, opened 2024-03-14
SQL printed to terminal is error 


wangzeping722:
## GORM Playground Link
https://github.com/go-gorm/playground/pull/702

## Description
When I use the where statement, the printed sql in console is wrong that makes me think there is something wrong with the sql statement.

 It should print number rather than string when the underlying type is int.
``` go 
type UserStatus int

const (
	Normal UserStatus = 1
	Forbid UserStatus = 2
)
// print: SELECT * FROM `users` WHERE status = '1' AND `users`.`deleted_at` IS NULL
DB.Where("status = ?", Normal).Find(&users)

// print: SELECT * FROM `users` WHERE status = 1 AND `users`.`deleted_at` IS NULL
DB.Where("status = ?", 1).Find(&users)
```

PR #6917, opened 2024-03-21
fix: 'type XXXX int' will print wrong sql to terminal

wangzeping722:
- [x] Do only one thing
- [x] Tested

ref: https://github.com/go-gorm/gorm/issues/6898

The wrong sql printed to the terminal can be confusing when use `type XXX int`, so fix it.

### User Case Description

```go
type UserStatus int

const (
	Normal UserStatus = 1
	Forbid UserStatus = 2
)

DB.Where("status = ?", Normal).Find(&users)
```
It print: 
``` go
SELECT * FROM `users` WHERE status = '1' AND `users`.`deleted_at` IS NULL
```

The expected print is: 
``` go 
SELECT * FROM `users` WHERE status = 1 AND `users`.`deleted_at` IS NULL
```


jinzhu:
thank you for your PR

PR #6933, opened 2024-03-30
Add debug loglevel

fnless:
Hi, I am new to gorm, and I am very curious that debug log level missed in the logger interface. If I Missed Anything, My Bad.

- [x] Do only one thing
- [x] Tested

As an logger `Interface`, I suppose that a `debug` log level should be owned to let it free extend.

### User Case Description

I want to use my own logger implementation, and use `debug` level to output the useful message only for developers.


a631807682 (commented):
Logically we should indeed use debug level logging in the Debug method.
But this is a breaking change and may not have enough benefit. There is no debug level log printing in gorm (because we have not implemented it), so we recommend that you use custom logs.

Issue #6819, opened 2024-02-02
ScanRows doesn't reset fields in target struct for incoming null values


turip:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/686

## Description

Starting with gorm v1.25.1 (v1.25.0 was not affected), if I use ScanRows on a struct that is not initialized to its default value for all fields GORM will not overwrite those fields that are NULL on SQL side.

An example codpiece can be found at https://github.com/go-gorm/playground/pull/686, but generally speaking we have encountered it when using a non-builtin type too (https://github.com/shopspring/decimal).

Xyon75:
Same problem using gorm.io/gorm v1.23.6 and gorm.io/driver/mysql v1.3.4.
Gorm model has bool or int values, table has corrisponding nullable fields.
When reading from DB multiple records using find(), if a record has a non null value, the field is set. For all subsequent records having the same field set to null, the returned model has the value set to the prior value instead of the default null value for the variable (false or 0).

Xyon75:
Updated deps to:
- gorm.io/driver/mysql v1.1.1
- gorm.io/gorm v1.25.7

Problem solved

turip:
I have opened a PR with some fixes: https://github.com/go-gorm/gorm/pull/6837/files

Waldeedle:
I just ran into this issue as well when upgrading gorm. It breaks the pattern of
```golang
    var datasetRow T
    for rows.Next() {
        err := db.ScanRows(rows, &datasetRow)
```
now I need to reallocate for each scan
```golang
    for rows.Next() {
    var datasetRow T
        err := db.ScanRows(rows, &datasetRow)
```

Waldeedle:
Actually this is broken since v1.25.0 (strings scanning NULL values still have the same error)

PR #6938, opened 2024-04-03
fix(scan): update Scan function to reset structs to zero values for each scan

Waldeedle:
- [x] Do only one thing
- [x] Tested

Addresses #6819
Adding the else condition to use the base type when data is nil ensures that the struct field is always set by the scanIntoStructMethod

### User Case Description
It fixes the example from the docs where you initialize the var once:
```golang
    var datasetRow T
    for rows.Next() {
        err := db.ScanRows(rows, &datasetRow)
```

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -244,7 +244,7 @@ func Scan(rows Rows, db *DB, mode ScanMode) {
 								rel = rel.FieldSchema.Relationships.Relations[name]
 								relFields = append(relFields, rel.Field)
 							}
-							// lastest name is raw dbname
+							// latest name is raw dbname
 							dbName := names[subNameCount-1]
 							if field := rel.FieldSchema.LookUpField(dbName); field != nil && field.Readable {
 								fields[idx] = field
--- a/tests/embedded_struct_test.go
+++ b/tests/embedded_struct_test.go
@@ -279,6 +279,6 @@ func TestEmbeddedTagSetting(t *testing.T) {
 	err = DB.Save(&t1).Error
 	AssertEqual(t, err, nil)
 	if t1.Tag1.Id == 0 {
-		t.Errorf("embedded struct's primary field should be rewrited")
+		t.Errorf("embedded struct's primary field should be rewritten")
 	}
 }
--- a/tests/non_std_test.go
+++ b/tests/non_std_test.go
@@ -37,7 +37,7 @@ func TestNonStdPrimaryKeyAndDefaultValues(t *testing.T) {
 	}
 
 	animal = Animal{From: "somewhere"}              // No name fields, should be filled with the default value (galeone)
-	DB.Save(&animal).Update("From", "a nice place") // The name field shoul be untouched
+	DB.Save(&animal).Update("From", "a nice place") // The name field should be untouched
 	DB.First(&animal, animal.Counter)
 	if animal.Name != "galeone" {
 		t.Errorf("Name fields shouldn't be changed if untouched, but got %v", animal.Name)
--- a/tests/query_test.go
+++ b/tests/query_test.go
@@ -1337,7 +1337,7 @@ func TestQueryResetNullValue(t *testing.T) {
 		Number1 int64      `gorm:"default:NULL"`
 		Number2 uint64     `gorm:"default:NULL"`
 		Number3 float64    `gorm:"default:NULL"`
-		Now     *time.Time `gorm:"defalut:NULL"`
+		Now     *time.Time `gorm:"default:NULL"`
 		Item1Id string
 		Item1   *QueryResetItem `gorm:"references:ID"`
 		Item2Id string
@@ -1427,7 +1427,7 @@ func TestQueryScanToArray(t *testing.T) {
 		t.Fatal(err)
 	}
 	if users[0] == nil || users[0].Name != "testname1" {
-		t.Error("users[0] not covere")
+		t.Error("users[0] not covered")
 	}
 	if users[1] != nil {
 		t.Error("users[1] should be empty")
--- a/tests/transaction_test.go
+++ b/tests/transaction_test.go
@@ -67,7 +67,7 @@ func TestTransaction(t *testing.T) {
 				return tx5.First(&User{}, "name = ?", "transaction-2").Error
 			})
 		}); err != nil {
-			t.Fatalf("prepare statement and nested transcation coexist" + err.Error())
+			t.Fatalf("prepare statement and nested transaction coexist" + err.Error())
 		}
 	})
 }


Waldeedle:
This code will overwrite every type of field except for embedded pointer type structs. 
It is difficult to determine whether or not that field should be reset to nil since if user is selecting a field i.e. NULL AS `embedded_ptr_struct_int_field` then the field should be initialized with its zero value but that would make a breaking change

Waldeedle:
@jinzhu any concerns with this PR?

scan.go
```diff
@@ -66,13 +66,21 @@ func (db *DB) scanIntoStruct(rows Rows, reflectValue reflect.Value, values []int
...
 		if field == nil {
 			continue
 		}
 
 		if len(joinFields) == 0 || len(joinFields[idx]) == 0 {
-			db.AddError(field.Set(db.Statement.Context, reflectValue, values[idx]))
+			fieldIsEmbeddedPointerTypeStruct := len(field.BindNames) > 1 && len(field.StructField.Index) > 0 && field.StructField.Index[0] < 0
+			fieldValue := reflect.ValueOf(values[idx]).Elem()
+			if !fieldIsEmbeddedPointerTypeStruct && fieldValue.Kind() == reflect.Ptr && fieldValue.IsNil() {
+				db.AddError(field.Set(db.Statement.Context, reflectValue, field.DefaultValueInterface))
```
jinzhu, line 79: Why use field.DefaultValueInterface? I don't think it is an expected behavior for most users.

Waldeedle: `field.DefaultValueInterface` ensures that I return the struct back with default/zero values for each field. i.e. nil for pointers and 0 for ints. It helps to reset the struct.

Waldeedle: @jinzhu any concerns with that?

Waldeedle: @jinzhu
I think it is expected usage based on your docs: https://gorm.io/docs/sql_builder.html#Scan-sql-Rows-into-struct
```golang
var user User
for rows.Next() {
  // ScanRows scan a row into user
  db.ScanRows(rows, &user)

  // do something
}
```

Waldeedle: @a631807682 @jinzhu ? any comments on the above? This is valid usage and should lower the memory allocations needed from my understanding

Waldeedle: @a631807682 @jinzhu ? any comments on the above? This is valid usage and should lower the memory allocations needed from my understanding

a631807682 (commented):
I didn’t understand what this PR was going to do from the test cases.
1. Why do we need to reset the value to the default value? Its means there a default value when the type is `Struct`, but not when the type is `Ptr`, and this is a breaking change.

```diff
type NestedEmbeddedStruct struct {
-	NestedEmbeddedIntField int
+	NestedEmbeddedIntField int `gorm:"default:2"`
}

+ result = Result{}

for rows.Next() {
	if err := DB.ScanRows(rows, &result); err != nil {
```
2. Can we reset all values via `reflect.Zero` before `scanIntoStruct`?

Waldeedle:
@a631807682 
1. the PR is because there is a breaking change in v1.24.6, for example if User is scanned 1 time, then the next scan if there is a Ptr field but the value from the db is null, the scan skips the field. Meaning User2 still has User1 data. This is because now for each row scan all values are not being reset.

```golang
var user User
for rows.Next() {
  // ScanRows scan a row into user
  db.ScanRows(rows, &user)

  // do something
}
```
2. I am not sure what you mean?

Waldeedle:
I cannot reset an embedded ptr struct to nil because THAT would be a breaking change for gorm, we instead want to have only the values reset not the embedded `ptr` structs

Waldeedle:
@a631807682 I expanded the test to also have your example, thank you for sharing that!
Here is the output that we expect for the scan when db returns nulls for all the scanned columns:
```golang
{
    BoolField:false
    IntField:0
    Int8Field:0
    Int16Field:0
    Int32Field:0
    Int64Field:0
    UIntField:0
    UInt8Field:0
    UInt16Field:0
    UInt32Field:0
    UInt64Field:0
    Float32Field:0
    Float64Field:0
    StringField:""
    TimeField:0001-01-01 00:00:00 +0000 UTC
    TimePtrField:<nil>
    EmbeddedStruct:{EmbeddedIntField:0 NestedEmbeddedStruct:{NestedEmbeddedIntField:0 NestedEmbeddedIntFieldWithDefault:2}}
    EmbeddedPtrStruct:<nil>
}
```

a631807682:
> @a631807682
> 
> 1. the PR is because there is a breaking change in v1.24.6, for example if User is scanned 1 time, then the next scan if there is a Ptr field but the value from the db is null, the scan skips the field. Meaning User2 still has User1 data. This is because now for each row scan all values are not being reset.
> 
> ```go
> var user User
> for rows.Next() {
>   // ScanRows scan a row into user
>   db.ScanRows(rows, &user)
> 
>   // do something
> }
> ```
> 
> 2. I am not sure what you mean?

1. My question is why User2 is not empty time, but the default value?
2. Can we reset reflect value like ?
```
	...
	// reset if type struct ...
	db.Statement.ReflectValue.Set(reflect.Zero(reflectValue.Type()))
    ...
	db.scanIntoStruct(...)
```

Waldeedle:
> > @a631807682
> > 
> > 1. the PR is because there is a breaking change in v1.24.6, for example if User is scanned 1 time, then the next scan if there is a Ptr field but the value from the db is null, the scan skips the field. Meaning User2 still has User1 data. This is because now for each row scan all values are not being reset.
> > 
> > ```go
> > var user User
> > for rows.Next() {
> >   // ScanRows scan a row into user
> >   db.ScanRows(rows, &user)
> > 
> >   // do something
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 2. I am not sure what you mean?
> 
> 1. My question is why User2 is not empty time, but the default value?
> 2. Can we reset reflect value like this?
> 
> ```
> 	...
> 	// reset if type struct ...
> 	db.Statement.ReflectValue.Set(reflect.Zero(reflectValue.Type()))
>     ...
> 	db.scanIntoStruct(...)
> ```

1. Maybe it would be clear if I explained my usage. I query a table for all of my user data and then I iterate through to format and save into an excel report. Before v1.24.6,  I used the golang code from the gorm docs to allow me to have 1 instance of the var that is just reused instead of allocating memory per row scan. I noticed my tests breaking when updating gorm because of the behavior change. For each scan the struct should start as default and then scan in the value from the db if there is one. The User2 data was not reset for me for the time.Time fields due to the database value being `NULL` and the User1 having data. Please let me know if that is a bit clearer. Thanks

2. I will try that

a631807682:
> The User2 data was not reset for me for the time.Time fields due to the database value being NULL and the User1 having data

I understand that the User2 time field needs to be reset, I am sure this is a bug, my question is why do we need to reset to defalut value instead of zero value?

Waldeedle:
> > The User2 data was not reset for me for the time.Time fields due to the database value being NULL and the User1 having data
> 
> I understand that the User2 time field needs to be reset, I am sure this is a bug, my question is why do we need to reset to defalut value instead of zero value?

@a631807682 Ah just got what you meant now, updated the logic to reset to the zero values. I originally had that but was not aware of the default value annotation and thought the DefaultValueInterface was how gorm was referring to zero values. I applied your solution and my test passes locally I think the tests are broken in places outside my change?

a631807682 (commented):
It is necessary to filter `Ptr` to avoid setting the original type as `Interface`.
The `stmt.ReflectValue` of pointer type is usually unwraped, with one exception, that is Interface, which will be converted to a convertible value in `Scan`. At this time, it cannot be set directly through `reflect.Zero`.
This part of the logic can be difficult to understand at the moment because it is scattered in different places and lacks comments. I think we need to optimize it, or add necessary comments in another PR.
https://github.com/go-gorm/gorm/blob/master/callbacks.go#L117
https://github.com/go-gorm/gorm/blob/master/scan.go#L188
```
if reflectValue.Kind() == reflect.Struct {
		db.Statement.ReflectValue.Set(reflect.Zero(reflectValue.Type()))
}
```

Waldeedle:
@a631807682 @jinzhu I accidentally closed this PR, but wanted to grab your thoughts about the commit: f0f01b1 (before the last one). Am I correct in using: `mode == ScanInitialized` to ensure the update and conflict modes are not impacted? This isn't breaking any functionality according to the tests, but I would like to double check since you may be aware of more context that I am not here. The last commit was just for me to see if I broke the tests or they were already broken in the master branch for GORM. Appreciate your time!

Waldeedle:
@a631807682 just looking for your insight on the commit f0f01b1, is that solution sound? If it is I will reopen an MR, just want to have the full context in this MR. So you can refer to the history easily. Thanks

Waldeedle:
@a631807682 just looking for your insight on the commit https://github.com/go-gorm/gorm/commit/f0f01b115610a16b9ef55031eeb3909719c2559e, is that solution sound? If it is I will reopen an MR, just want to have the full context in this MR. So you can refer to the history easily. Thanks

a631807682:
> @a631807682 just looking for your insight on the commit [f0f01b1](https://github.com/go-gorm/gorm/commit/f0f01b115610a16b9ef55031eeb3909719c2559e), is that solution sound? If it is I will reopen an MR, just want to have the full context in this MR. So you can refer to the history easily. Thanks

LGTM

Issue #6956, opened 2024-04-11
`unsupported data` error with nested joins/preloads


N-Schaef:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/715

## Description

In our project we have deeply nested data that needs to be joined and some additional values preloaded.

In a previous version of gorm (1.25.5) this was working.
Now we get a `unsupported data` error stemming from [`gorm/callbacks/preload.go#L140`](https://github.com/go-gorm/gorm/blob/1e13fd7543826a5d341401387cb1aa22fdf4b9f6/callbacks/preload.go#L140)

The issue seems to be introduced in #6877.
The value of `db.Statement.ReflectValue.Kind().String()` is `ptr`.
My best guess is that preloads of sub-structs of optional nested structs/joins are the problem.

If I duplicate the whole `case reflect.Struct:` case (in  [`gorm/callbacks/preload.go#L133`](https://github.com/go-gorm/gorm/blob/1e13fd7543826a5d341401387cb1aa22fdf4b9f6/callbacks/preload.go#L133)) as an additional `case reflect.Pointer:` case, our code works again.

PR #6957, opened 2024-04-11
fix: `unsupported data` on nested joins with preloads

N-Schaef:
- [x] Do only one thing
- [x] Tested

This PR fixes issue #6956.

It adds `reflect.Pointer` as an valid reflect target for preloads.
This seems to be required for preloads of nested optional joins with preload targets.

I am not 100% sure of the broader implications of this change. But the implementation before #6877 also allowed this usecase.

### User Case Description

We have complex queries with nested joins and preloads that worked on version 1.25.5 but do not work with the current version.

### DIFF
--- a/callbacks/preload.go
+++ b/callbacks/preload.go
@@ -130,7 +130,7 @@ func preloadEntryPoint(db *gorm.DB, joins []string, relationships *schema.Relati
 							return err
 						}
 					}
-				case reflect.Struct:
+				case reflect.Struct, reflect.Pointer:
 					reflectValue := rel.Field.ReflectValueOf(db.Statement.Context, rv)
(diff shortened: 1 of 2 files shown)


jinzhu:
Hi @N-Schaef 

Can you add some test cases to make sure it works?

Thank you.

N-Schaef:
Hello @jinzhu,

sorry for the late reply. I've been on vacation.

I have added a test-case that results in the error on the main branch but works with the changes.

But somehow an (in my opinion) unrelated error now fails on the mysql:latest.
I'm not sure whats broken there.

N-Schaef:
Is there anything left to do?

a631807682 (approved):

Issue #6192, opened 2023-03-27
Empty primary key being inserted


chad-bekmezian-snap:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/587

## Description

From what I understand about gorm, struct fields that are equal to their zero values should be automatically omitted when inserting into a database. Specifically, if the primary key of a struct is equal to its zero value, gorm should not attempt to insert that value. This was the case prior to v1.20.4, which contains [this commit](https://github.com/go-gorm/gorm/commit/d834dd60b715422dc2a900fb2744f9c278a9830f), breaking the described functionality.

For example...
``` Go
import "github.com/google/uuid"

type Person struct {
   ID               uuid.UUID
   FirstName string
   LastName string
}

function CreatePerson(db *gorm.DB) (Person, error) {
  person := {
     FirstName: "John",
     LastName: "Doe",
   }

   tx := db.Debug().Create(&person) 
   // EXPECTED debug output: INSERT INTO "persons" ("first_name", "last_name") VALUES ('John', 'Doe')
   // ACTUAL debug output: INSERT INTO "persons" ("id", "first_name", "last_name") VALUES ('00000000-0000-0000-0000-000000000000', 'John', 'Doe')

   return person, tx.Error
}
```

TheDhejavu:
Hey @chad-bekmezian-snap 

**try this:**

```go
import "github.com/google/uuid"

type Person struct {
   ID               uuid.UUID
   FirstName string
   LastName string
}

func (p *Person) BeforeCreate(tx *gorm.DB) (err error) {
	p.ID = uuid.New()
	return 
}


func CreatePerson(db *gorm.DB) (Person, error) {
  person := {
     FirstName: "John",
     LastName: "Doe",
   }

   tx := db.Debug().Create(&person) 

   return person, tx.Error
}
```

chad-bekmezian-snap:
> Hey @chad-bekmezian-snap
> 
> **try this:**
> 
> ```go
> import "github.com/google/uuid"
> 
> type Person struct {
>    ID               uuid.UUID
>    FirstName string
>    LastName string
> }
> 
> func (p *Person) BeforeCreate(tx *gorm.DB) (err error) {
> 	p.ID = uuid.New()
> 	return 
> }
> 
> 
> func CreatePerson(db *gorm.DB) (Person, error) {
>   person := {
>      FirstName: "John",
>      LastName: "Doe",
>    }
> 
>    tx := db.Debug().Create(&person) 
[...]

PR #6970, opened 2024-04-16
Fix:  default string primary keys

martinmunillas:
# Default string primary keys

- [x] Do only one thing
- [x] Tested

#### Issue https://github.com/go-gorm/gorm/issues/6192

About a year ago there was a [commit](https://github.com/go-gorm/gorm/commit/d834dd60b715422dc2a900fb2744f9c278a9830f#diff-eafe62eae191e88faad4c223c4de9af268c175ee68695872e175b814d7c5baa5) "removing unnecessary code".

This change did not remove unnecessary code, but broke inserting into table that use strings as primary keys.

Now instead of relying in the default of the database it would insert a zero value instead, storing empty strings for varchars/text, storing 00000000-0000-0000-0000-000000000000 for UUIDs, etc.

### User Case Description

Take the following model

```go
type User struct {
  ID uuid.UUID
  Name string
}
```
With a table of the following shape
```sql
CREATE TABLE "users" (
    "id" UUID DEFAULT gen_random_uuid() PRIMARY KEY,
    "name" VARCHAR NOT NULL
);
```
If you tried to do 
```go
user :=  User{Name: "Martin Munilla"}
db.Save(user)
```
You would get the following sql executed
```sql
INSERT INTO "users" ("id", "name") VALUES ('00000000-0000-0000-0000-000000000000', 'Martin Munilla')
```
Instead of
```sql
INSERT INTO "users" ("name") VALUES ('Martin Munilla')
```
Thus not allowing the database to execute the default clause

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -299,6 +299,14 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 				field.HasDefaultValue = true
 				field.AutoIncrement = true
 			}
+		case String:
+			if _, ok := field.TagSettings["PRIMARYKEY"]; !ok {
+				if !field.HasDefaultValue || field.DefaultValueInterface != nil {
+					schema.FieldsWithDefaultDBValue = append(schema.FieldsWithDefaultDBValue, field)
+				}
+
+				field.HasDefaultValue = true
+			}
 		}
 	}
 
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -300,13 +300,11 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 				field.AutoIncrement = true
 			}
 		case String:
-			if _, ok := field.TagSettings["PRIMARYKEY"]; !ok {
-				if !field.HasDefaultValue || field.DefaultValueInterface != nil {
-					schema.FieldsWithDefaultDBValue = append(schema.FieldsWithDefaultDBValue, field)
-				}
-
-				field.HasDefaultValue = true
+			if !field.HasDefaultValue || field.DefaultValueInterface != nil {
+				schema.FieldsWithDefaultDBValue = append(schema.FieldsWithDefaultDBValue, field)
 			}
+
+			field.HasDefaultValue = true
 		}
 	}
 
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -300,11 +300,13 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 				field.AutoIncrement = true
 			}
 		case String:
-			if !field.HasDefaultValue || field.DefaultValueInterface != nil {
-				schema.FieldsWithDefaultDBValue = append(schema.FieldsWithDefaultDBValue, field)
-			}
+			if _, ok := field.TagSettings["PRIMARYKEY"]; !ok {
+				if !field.HasDefaultValue || field.DefaultValueInterface != nil {
+					schema.FieldsWithDefaultDBValue = append(schema.FieldsWithDefaultDBValue, field)
+				}
 
-			field.HasDefaultValue = true
+				field.HasDefaultValue = true
+			}
 		}
 	}
 
--- a/schema/schema_test.go
+++ b/schema/schema_test.go
@@ -341,11 +341,6 @@ func TestStringPrimaryKeyDefault(t *testing.T) {
 		Code string
 		Name string
 	}
-	type ProductWithNamedPrimaryKey struct {
-		ProductID string `gorm:"primaryKey"`
-		Code      string
-		Name      string
-	}
 
 	product, err := schema.Parse(&Product{}, &sync.Map{}, schema.NamingStrategy{})
 	if err != nil {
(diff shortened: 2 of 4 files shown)


efournival:
:+1:

jinzhu:
Hi @martinmunillas 

Thank you for your PR, can you add some tests for the function?

martinmunillas:
@jinzhu There was a test that I had to modify `upsert_test.go` which I think that makes sense that the sql query is now adding a RETURNING code clause as it  now taking the code as a string primary key, but I think you should confirm

martinmunillas:
@jinzhu also I'm not sure why some mysql and mariadb tests are failing but the others arent's

martinmunillas:
Okay, it was because I had changed the behaviour to also allow custom named primary keys to default, which would be nice but I'm not 100% what other places that would affect so I decided to roll that back.

schema/schema.go
```diff
@@ -299,6 +299,14 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 				field.HasDefaultValue = true
 				field.AutoIncrement = true
 			}
+		case String:
+			if _, ok := field.TagSettings["PRIMARYKEY"]; !ok {
+				if !field.HasDefaultValue || field.DefaultValueInterface != nil {
+					schema.FieldsWithDefaultDBValue = append(schema.FieldsWithDefaultDBValue, field)
+				}
+
+				field.HasDefaultValue = true
```
shedyfreak, line 308: I don't think we should allow default value on primarykeys unless there is a function to autoincrement in the database

shedyfreak (changes requested):

schema/schema.go
martinmunillas: why not? it's okay that an insert fails if you don't provide a primary key and didn't setup a default mechanism. It's not okay to pass a zero value instead, as it would only prevent it for failing the first time and failing after that for being non unique.

manicar2093: I support this change. After all this behavior existed before. And we should think that DB may generate id. Why Gorm should care of id generation?

manicar2093:
@jinzhu what do yo think of this change? Is it possible to add them? 😃

jinzhu:
From the tests, the behavior doesn't seems correct, why not just tell GORM the id will generated by database like this?

image

martinmunillas:
@jinzhu  I don't get what is the difference with 
```go
case Int, Uint:
	if _, ok := field.TagSettings["AUTOINCREMENT"]; !ok {
		if !field.HasDefaultValue || field.DefaultValueInterface != nil {
			schema.FieldsWithDefaultDBValue = append(schema.FieldsWithDefaultDBValue, field)
		}
		field.HasDefaultValue = true
		field.AutoIncrement = true
	}
```
why can we assume in ints (which could perfectly not be increment or have a default) that they do have a default but not for strings?

jinzhu:
> @jinzhu I don't get what is the difference with
> 
> ```go
> case Int, Uint:
> 	if _, ok := field.TagSettings["AUTOINCREMENT"]; !ok {
> 		if !field.HasDefaultValue || field.DefaultValueInterface != nil {
> 			schema.FieldsWithDefaultDBValue = append(schema.FieldsWithDefaultDBValue, field)
> 		}
> 		field.HasDefaultValue = true
> 		field.AutoIncrement = true
> 	}
> ```
> 
> why can we assume in ints (which could perfectly not be increment or have a default) that they do have a default but not for strings?

For integer primary key fields, we know the database will generate the IDs, but for string fields, it is not always true

martinmunillas:
@jinzhu that isn't true, you could perfectly have a table like the following
```ql
CREATE TABLE example_table (
    id INT PRIMARY KEY,
    name VARCHAR(255)
);
```
in this case the id is a integer, not incremental and doesn't have a default

jinzhu:
> @jinzhu that isn't true, you could perfectly have a table like the following
> 
> ```ql
> CREATE TABLE example_table (
>     id INT PRIMARY KEY,
>     name VARCHAR(255)
> );
> ```
> 
> in this case the id is a integer, not incremental and doesn't have a default

Our concern is whether the SQL generated by GORM is consistent with GORM's other logic. For integer fields, it is consistent now, but for string fields, the consistency will be broken after the PR

martinmunillas:
Okay I understand now, thank you for the support

manicar2093:
> From the tests, the behavior doesn't seems correct, why not just tell GORM the id will generated by database like this?
> 
> image

Hi, this does not work.

```
INSERT INTO "users" ("id") VALUES ('00000000-0000-0000-0000-000000000000')
```

```
type User struct {
	Id               uuid.UUID                   `json:"id" gorm:"primaryKey,default:gen_random_uuid()"`
}
```

In order to make it work primaryKey should be removed. is this the expected behavior?

jgermond:
> Hi, this does not work.
> 
> ```
> INSERT INTO "users" ("id") VALUES ('00000000-0000-0000-0000-000000000000')
> ```
> 
> ```
> type User struct {
> 	Id               uuid.UUID                   `json:"id" gorm:"primaryKey,default:gen_random_uuid()"`
> }
> ```
> 
> In order to make it work primaryKey should be removed. is this the expected behavior?

Use a semicolon, not a comma, in `primaryKey;default:gen_random_uuid()`.
That's work for me.

PR #6973, opened 2024-04-18
fix: int type field size should is int32

aveyuan:
- [x] Do only one thing
- [x] Tested

go type int,uint field size should is 32

### User Case Description
int type size is 32 in go . but field is 64.


a631807682 (commented):
https://go.dev/ref/spec#Constants

`int` always equals the processor bit-size. So on 64 bit systems, it's 64 bit, on 32bit systems, it's 32 bit.

aveyuan:
> https://go.dev/ref/spec#Constants
> 
> `int` always equals the processor bit-size. So on 64 bit systems, it's 64 bit, on 32bit systems, it's 32 bit.

Thinks for you.

PR #6999, opened 2024-04-29
fix: strings.Title -> cases.Title bcs strings.Title library is deprecated

ryuji-cre8ive:
- [x] Do only one thing
- [x] Tested

change using library strings.Title to cases.Title because of strings. Title is deprecated for Go 1.18
below is a warnings when I run golangci-cli:
```
strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.
``` 

I rewrite strings.Title -> cases.Title.
cases.TItle returns a Caser so it needed to parse to string because strings.Title returns string

cases.Title can be taken language, but I don't know which language supports for gorm so I tried to write `language.Und`.
this means you didn't decide which language.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,4 +5,5 @@ go 1.18
 require (
 	github.com/jinzhu/inflection v1.0.0
 	github.com/jinzhu/now v1.1.5
+	golang.org/x/text v0.14.0
 )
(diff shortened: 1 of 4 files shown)


daheige (approved):
Use `golang.org/x/text/cases` LGTM

jinzhu (approved):

PR #7003, opened 2024-04-30
fix: typo

hakusai22:
fix: typo


tests/query_test.go
```diff
@@ -1427,7 +1427,7 @@ func TestQueryScanToArray(t *testing.T) {
 		t.Fatal(err)
 	}
 	if users[0] == nil || users[0].Name != "testname1" {
-		t.Error("users[0] not covere")
+		t.Error("users[0] not cover")
```
greencoda, line 1430: `covered` was probably the intended text here:
```suggestion
		t.Error("users[0] not covered")
```

hakusai22: done

Issue #7002, opened 2024-04-30
Where Condition Build Issue


VarusHsu:
I have my sql query with multi `Where` conditions.
When using `||` instead of `OR`, there is diiferent rows fetched reason of brackets.
Is there a bug or designed like this ?

<img width="1393" alt="WeChatWorkScreenshot_bef48b6d-29d7-45bc-a0d1-fa70d830bf9d" src="https://github.com/go-gorm/gorm/assets/85171099/44072ec8-121d-402e-9b67-9ee9898e094e">

VarusHsu:
pr link see:
https://github.com/go-gorm/gorm/pull/7004

PR #7004, opened 2024-04-30
Fix (#7002) Where Condition Build Issue

VarusHsu:
- [x] Do only one thing
- [x] Tested

Add brackets wrap when sql expression contains `||`

### User Case Description
* single `Expr` without brackets wrapped.(follow previous rules)
* multiple `Expr` and contains `||` will brackets wrapped.(new feature)

### Playground Link
https://github.com/go-gorm/playground/pull/728

### Issue Link
https://github.com/go-gorm/gorm/issues/7002

### DIFF
--- a/clause/where.go
+++ b/clause/where.go
@@ -5,8 +5,10 @@ import (
 )
 
 const (
-	AndWithSpace = " AND "
-	OrWithSpace  = " OR "
+	AndWithSpace       = " AND "
+	OrWithSpace        = " OR "
+	AndSymbolWithSpace = " && "
+	OrSymbolWithSpace  = " || "
 )
 
 // Where where clause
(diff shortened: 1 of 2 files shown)


sprataa:
great find, but, I believe the reason why this is not covered in the first place is because || or && are not part of the ANSI SQL standard. 

as an example, || is the concatenate operator in postgresql and && is not supported. in postgresql `select 'this' || ' ' || 'is' = 'this is'` yields the same result as `select ('this' || ' ' || 'is') = 'this is'`. logically `select 1 = 1 || 2 = 2` yields a syntax error and `select (1 = 1)::text || (2 = 2)::text` yields the string `'truetrue'` as supposed.

while this change you are proposing looks harmless, I don't believe there's any good reason to have these extra comparisons overhead for something which is not a standard for the engines gorm supports.

my 2 cents are, you should be using `OR`, `AND` and `CONCAT` so your SQL code is ANSI compliant and works with every engine.

cheers

PR #7007, opened 2024-05-06
add DB level propagation for the Unscoped flag

sprataa:
- [x] Do only one thing
- [x] Tested

This PR adds `PropagateUnscoped` to `DB`'s `Config`. Its purpose is to allow propagation of the parent `Statement`'s `Unscoped` flag to nested statements. 

#### Should this become the default behavior?

Is there any scenario where `getInstance()` where we don't want this flag to propagate? For the record, all tests pass if we just persist the `Unscoped` state when `clone == 1`. 

I just didn't want to introduce what may be a breaking change in more complex production environments for which gorm's tests don't offer coverage therefore this is a disabled by default flag.

Regardless, this kind of change should not be introduced without a flag to panic control it.

### User Case Description

When one wants to create advanced hooks which leverage not just the usual Before/After but also `StatementModifier` hooks while reusing the `Unscoped` state of the parent statement, it can be expected that the flag propagates to all child hooks.

One can, and that's what we find ourselves doing more often than not, is to store the original tx/stmt's `Unscoped` state and adding repeated extra lines of code where one wants to keep the state after a nested `Delete` or `Model` calls done inside hooks - arguably, in every possibly nested hook.

Since for some, or even all, use cases, one just wants to have `Unscoped` propagated to all nested statements, this is a way cleaner and non intrusive solution to address such challenge.

### DIFF
--- a/tests/hooks_test.go
+++ b/tests/hooks_test.go
@@ -2,6 +2,8 @@ package tests_test
 
 import (
 	"errors"
+	"log"
+	"os"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


tests/hooks_test.go
```diff
@@ -566,3 +566,41 @@ func TestUpdateCallbacks(t *testing.T) {
...
+	DB.Migrator().DropTable(&Product6{}, &ProductItem2{})
+	DB.AutoMigrate(&Product6{}, &ProductItem2{})
+
+	p := Product6{
+		Name: "unique_code",
+		Item: &ProductItem2{},
+	}
+	DB.Model(&Product6{}).Create(&p)
+
+	DB.PropagateUnscoped = true
```
jinzhu, line 598: This is not thread-safe.

sprataa: indeed. pushed a fix and updated the test

jinzhu:
Can you add some doc to gorm.io, thank you

PR #7061, opened 2024-06-12
fix(scan): update Scan function to reset structs to zero values for each scan

Waldeedle:
- [x] Do only one thing
- [x] Tested

Accidentally closed #6938 so this is the same PR just with the final implemented solution
Addresses #6819
Adding a condition to reset the values of a struct when a scan is initialized to handle nil values from the database

### User Case Description
It fixes the example from the docs where you initialize the var once:
```golang
    var datasetRow T
    for rows.Next() {
        err := db.ScanRows(rows, &datasetRow)
```

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -331,6 +331,9 @@ func Scan(rows Rows, db *DB, mode ScanMode) {
 			}
 		case reflect.Struct, reflect.Ptr:
 			if initialized || rows.Next() {
+				if mode == ScanInitialized && reflectValue.Kind() == reflect.Struct {
+					db.Statement.ReflectValue.Set(reflect.Zero(reflectValue.Type()))
+				}
 				db.scanIntoStruct(rows, reflectValue, values, fields, joinFields)
 			}
 		default:
--- a/tests/scan_test.go
+++ b/tests/scan_test.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"strings"
 	"testing"
+	"time"
 
 	"gorm.io/gorm"
 	. "gorm.io/gorm/utils/tests"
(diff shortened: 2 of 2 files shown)


Waldeedle:
@a631807682 here's the new PR with the final solution. Appreciate all the help you've provided on this! 🚀 🎉

a631807682 (approved):

Waldeedle:
@a631807682 am I correct in assuming @jinzhu will be merging into master, after reviewing?

a631807682:
I usually wait for at least 2 members to review, and jinzhu will review all PRs and merge them when he is free.

Jcodelove:
OMG, I've been treating this as a feature the whole time.

Waldeedle:
@Jcodelove i hope that means you liked my changes and not that I caused you to look for this because of a regression!

ZGeomantic:
It doesn't seem like a "Non breaking API changes". Is this the expected behavior of this function itself?

tashaX:
I'm not sure if this is related, but there's a significant change in the `Scan` function's behaviour. The GORM version just got a minor update, but now `Scan` wipes clean the whole struct. That's a pretty big change.

We used to use `Scan` something like this:

```go
type SomeStruct struct {
    Name        string
    Date        null.Time
    Description string
    ID          uint64
}

data := SomeStruct{
    ID:   1,
    Date: null.TimeFrom(time.Now()),
}

db.Table("some_table").
    Select("name, description").
    Where("id = ?", data.ID).
    Scan(&data)
```

The intention was to fill in new data while keeping the already populated fields intact.  
But now, after the update, the struct is wiped clean and only the selected data is present after `Scan`.  
That seems like a major behavioural change.

PR #7086, opened 2024-06-25
feat: Modernize Docker Compose File

isso-719:
- [x] Do only one thing
- [x] Tested

This pull request did the following
- [x] Rename the file from docker-compose.yml to compose.yml .
- [x] Remove version top-level element on compose.yml .
- [x] ~~Changed command used from `docker-compose` to `docker compose`.~~
- [x] ~~Change related files.~~

#### Rename the file from docker-compose.yml to compose.yml .

The file name `compose.yml` is preferred.
https://docs.docker.com/compose/compose-application-model/#the-compose-file

#### Remove version top-level element on compose.yml .

Version top-level element is obsoleted.
https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-obsolete

When run test, system shows warning.

```sh
tests % ./tests_all.sh
WARN[0000] /Users/john/gorm/tests/docker-compose.yml: `version` is obsolete
...
```

#### Changed command used from `docker-compose` to `docker compose`.

The recommended command-line syntax is `docker compose`.
https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2

#### Change related files.

`docker-compose start` => `docker compose start`
`docker-compose up` => `docker compose up`

### DIFF
--- a/tests/compose.yml
+++ b/tests/compose.yml
@@ -1,5 +1,3 @@
-version: '3'
-
 services:
   mysql:
     image: 'mysql/mysql-server:latest'


jinzhu:
It seems that docker-compose is still an alias of docker compose in the new version. Let's keep it for better compatibility.

isso-719:
> It seems that docker-compose is still an alias of docker compose in the new version. Let's keep it for better compatibility.

@jinzhu 
Thank you for your reply.
For compatibility, I reverted from the `docker compose` command to the `docker-compose` command.

Please check Files changed.
Thank you.

PR #7093, opened 2024-07-02
`DB` provides `Close` function

demoManito:
- [ ] Do only one thing
- [ ] Tested


jinzhu:
We don't want to provide the `Close` method in GORM by default, as it has been misused frequently...

Issue #7100, opened 2024-07-09
AutoMigrate method creates incorrect index name


y-okuno:
## Description:
I encountered an issue with GORM's `AutoMigrate` function when upgrading to the latest versions.

**Successful migration with:**
- `gorm.io/driver/postgres v1.4.8`
- `gorm.io/gorm v1.25.5`

**Failed migration with:**
- `gorm.io/driver/postgres v1.5.9`
- `gorm.io/gorm v1.25.10`

In the latest version, it seems that the index name specified in the model is not being respected. The `AutoMigrate` function creates an incorrect index name which leads to a migration failure.

### Steps to Reproduce

1. Use the following versions:
   - `gorm.io/driver/postgres v1.5.9`
   - `gorm.io/gorm v1.25.10`

2. Define the model as follows:
   ```go
   type Master struct {
       ID         uint32 `gorm:"primaryKey"`
       IPv4Prefix string `gorm:"uniqueIndex"`
   }

3. Create the table in PostgreSQL with the following SQL:
```sql
CREATE TABLE public.masters (
    id bigserial NOT NULL,
    ipv4_prefix text NULL,
    CONSTRAINT idx_masters_ipv4_prefix UNIQUE (ipv4_prefix)
);
```

4. Execute AutoMigrate method with the above model.

5. Also tried with the following model definition, but the result was the same:
```go
type Master struct {
    ID         uint32 `gorm:"primaryKey"`
    IPv4Prefix string `gorm:"uniqueIndex:idx_masters_ipv4_prefix"`
}
```

### Expected Behavior
The migration should succeed without errors.

### Actual Behavior
The migration fails with the following error:
```
ERROR: constraint "uni_masters_ipv4_prefix" of relation "masters" does not exist (SQLSTATE 42704)
```

It appears that AutoMigrate is attempting to create a constraint named `uni_masters_ipv4_prefix` instead of respecting the specified `idx_masters_ipv4_prefix` .

### Environment
- GORM version: v1.25.10
- GORM Postgres driver version: v1.5.9
- PostgreSQL version: 13.4
- Go version: 1.21.1

### Additional Context
This issue does not occur with the following versions:

- gorm.io/driver/postgres v1.4.8
- gorm.io/gorm v1.25.5

It seems that the index name specified in the struct tag is not being applied correctly in the latest version.

Thank you for your attention to this issue. Please let me know if you need any further information or assistance in reproducing the issue.

dashrews78:
[...]

PR #6640, opened 2023-10-13
feat: add MigrateColumnUnique

black-06:
- [X] Do only one thing
- [ ] Non breaking API changes
- [ ] Tested

### What did this pull request do?

It's a part of https://github.com/go-gorm/gorm/pull/6386.

Defining new methods for use in drivers

Issue #6888, opened 2024-03-10
1.2.7 breaks migrations when updating 1.2.5


ChristianSch:
Playground link doesn't make any sense, as I'd need to use multiple versions at once.

Say you have the following model:

```
type User struct {
	BaseModel

	// The user's email address
	Email                    string        `gorm:"not null;uniqueIndex:idx_user_email" json:"email"`
}
```

Migrating with:

`tx.AutoMigrate(&domain.User{})`

Causes:

```
2024/03/10 22:09:04  ERROR: constraint "uni_users_email" of relation "users" does not exist (SQLSTATE 42704)
[0.697ms] [rows:0] ALTER TABLE "users" DROP CONSTRAINT "uni_users_email"
```

I'd appreciate, if releases would be properly tested beforehand. I'll be pinning 1.2.5 for now.

link for not being marked as stale: https://github.com/go-gorm/playground/issues/0

ChristianSch:
v1.25.7-0.20240204074919-46816ad31dde works

klipach:
Have the same issue while using  v1.25.9 and PostgreSQL. Downgrading to v1.25.5 resolves the issue.

zetaab:
we are also seeing this

```
2024/04/09 05:52:14  pq: constraint "uni_image_statistics_image_id" of relation "image_statistics" does not exist
[0.955ms] [rows:0] ALTER TABLE "image_statistics" DROP CONSTRAINT "uni_image_statistics_image_id"
```

`,"error":"pq: constraint \"uni_image_statistics_image_id\" of relation \"image_statistics\" does not exist"`

UninspiredNickname:
I can confirm that - this error happens because in https://github.com/go-gorm/gorm/blob/v1.25.9/migrator/migrator.go#L569 `field.Unique` returns `false` when `UniqueIndex` is used on the models - simply swapping it  to `unique;index` fixes the problem for now, but `field.Unique` should be set to true if `uniqueIndex` is used

varfrog:
I wish this library:
1. used proper versioning - semver
2. had a changelog
3. test their changes
big asks fsho... I don't expect any of these to be implemented though. Who versions their libraries anyways... changelog? c'mon...

PaulSonOfLars:
+1 on this issue. Also noticed this upon upgrading from 1.25.5 to 1.25.9.

This may be helpful for debugging this issue -
[...]

PR #7135, opened 2024-08-01
fix: auto migrate tries to delete non-existent unique constraint

dashrews78:
- [ ] Do only one thing
- [ ] Tested

This change simply updates the migrator to use `IF EXISTS` when trying to drop constraint.  The reason for this is based on issue #7100 .  With the PR #6640 the naming for unique constraints was changed.  Previously the migrator was creating unique indexes with a naming prefix of `idx_`.  When #6640 went it, the name changed to be `uni_`.  But if your previous upgrade created the constraint with a name of `idx_` the migrator would see you had a constraint and try to delete it.  But it was using the updated naming pattern in this statement.  So that would result in a command like `ALTER TABLE "compliance_integrations" DROP CONSTRAINT "uni_compliance_integrations_clusterid"`.  Since the name of the constraint is `idx_compliance_integrations_clusterid` the call to `DROP CONSTRAINT` fails.  The point of this change is to simply make it a no-op if the constraint doesn't exist.  That will prevent failure.  The old constraint may linger in these cases but that is not the scope of this change.  The scope of this change is to have a migrator that does not cause unnecessary failures for users.

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -762,7 +762,8 @@ func (m Migrator) DropConstraint(value interface{}, name string) error {
 		if constraint != nil {
 			name = constraint.GetName()
 		}
-		return m.DB.Exec("ALTER TABLE ? DROP CONSTRAINT ?", clause.Table{Name: table}, clause.Column{Name: name}).Error
+		// using IF EXISTS here as we are "guessing" the constraint name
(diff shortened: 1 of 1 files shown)


varfrog:
Thanks for the fix. Hopefully, it ends up in a tag!
A related issue also: https://github.com/go-gorm/gorm/issues/6888

rayshihUI:
This PR will help. please do more progress.

PR #7141, opened 2024-08-06
Add parse methods with central log instance

nettnikl:
- [x] Do only one thing
- [x] Tested (works as far as i can tell)

Added additional methods that include a logger as parameter, so internal methods can pass a the logger from the config though.

### User Case Description
When i use gorm, i get logs like this:
```
2024/08/06 20:08:59 /project/internal/http_server/http.go:34
[error] invalid field found for struct wishes/internal/model.Wish's field Owner: define a valid foreign key for relations or implement the Valuer/Scanner interface
```

Currently, i need to do something like this.
```golang
import "io/gorm/logger"
var g *gorm.DB
logger.Default = g.Logger
```
To get gorm to use the central logger (this is the same (unrelated) error, but in the format i specified for my logs):
```
19:32:47 ERR ../../go/pkg/mod/gorm.io/gorm@v1.25.11/schema/schema.go:334 > invalid field found for struct wishes/internal/model.Wish's field Owner: define a valid foreign key for relations or implement the Valuer/Scanner interface type=gorm
```

### DIFF
--- a/association.go
+++ b/association.go
@@ -22,7 +22,7 @@ func (db *DB) Association(column string) *Association {
 	association := &Association{DB: db}
 	table := db.Statement.Table
 
-	if err := db.Statement.Parse(db.Statement.Model); err == nil {
+	if err := db.Statement.ParseWithLogger(db.Statement.Model, db.Logger); err == nil {
 		db.Statement.Table = table
 		association.Relationship = db.Statement.Schema.Relationships.Relations[column]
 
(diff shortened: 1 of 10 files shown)


jinzhu:
It seems that we should define a schema.Parser object and initialize the necessary components—such as the Logger—within the db instance. Then we can use this Parser to parse models later on. 

I will go ahead and close this PR for now, but you’re welcome to submit a new one with the proposed changes.

Thank you.

Issue #7137, opened 2024-08-02
BUG：`PrepareStmt：true` 在 SQL 执行时错误 `PreparedSQL` 会内存泄露


FreeMasonGalaxy:
#### BUG 原因
>**开启预编译后，gorm `(db *PreparedStmtDB) prepare` 会拿预编译语句先到 stmts 查找是否存在预编译 (stmt) 对象，有则直接返回 stmt 对象，无则缓存预编译对象，再缓存预编译语句，`(db *PreparedStmtDB) ExecContext` 有报错则 `delete(db.Stmts, query)` 删除预编译对象，但未删除缓存预编译语句，会造成 gorm 发送预编译通过但 SQL 执行时错误(如不可重复) PreparedSQL 一直增长未释放数据**

#### 版本
+ go：1.22
+ gorm：1.25.8

#### 复现
1. gorm 配置
```go
DB, _= gorm.Open(mysql.New(conf), &gorm.Config{
		PrepareStmt: true,
	})
```
2. 调用创建方法
```go
database.DB.Create(&m)
```

3. 报错
```go
 Duplicate entry
```

#### Gorm 预编译源码

```go
type Stmt struct {
	*sql.Stmt
	Transaction bool
	prepared    chan struct{}
	prepareErr  error
}

type PreparedStmtDB struct {
	Stmts       map[string]*Stmt
	PreparedSQL []string
	Mux         *sync.RWMutex
	ConnPool
}

func NewPreparedStmtDB(connPool ConnPool) *PreparedStmtDB {
	return &PreparedStmtDB{
		ConnPool:    connPool,
		Stmts:       make(map[string]*Stmt),
		Mux:         &sync.RWMutex{},
		PreparedSQL: make([]string, 0, 100),
	}
}

func (db *PreparedStmtDB) GetDBConn() (*sql.DB, error) {
	if sqldb, ok := db.ConnPool.(*sql.DB); ok {
		return sqldb, nil
	}

	if dbConnector, ok := db.ConnPool.(GetDBConnector); ok && dbConnector != nil {
		return dbConnector.GetDBConn()
	}

	return nil, ErrInvalidDB
}

func (db *PreparedStmtDB) Close() {
	db.Mux.Lock()
	defer db.Mux.Unlock()

	for _, query := range db.PreparedSQL {
		if stmt, ok := db.Stmts[query]; ok {
			delete(db.Stmts, query)
			go stmt.Close()
		}
	}
}

[...]

PR #7142, opened 2024-08-07
fix memory leaks in PrepareStatementDB

ivila:
- [x] Do only one thing
- [x] Tested

Fix the issue #7137 

### User Case Description

Just fix the memory leaks.

### DIFF
--- a/prepare_stmt.go
+++ b/prepare_stmt.go
@@ -55,9 +55,8 @@ func (db *PreparedStmtDB) Close() {
 			}
 		}(stmt)
 	}
-	// I think setting it to nil should be better, as there should be no one
-	// writing into it after calling Close, but leave it for compatibility and safety
-	db.Stmts = make(map[string]*Stmt)
+	// setting db.Stmts to nil to avoid further using
+	db.Stmts = nil
 }
 
 func (sdb *PreparedStmtDB) Reset() {
@@ -65,7 +64,13 @@ func (sdb *PreparedStmtDB) Reset() {
 	defer sdb.Mux.Unlock()
 
 	for _, stmt := range sdb.Stmts {
-		go stmt.Close()
+		go func(s *Stmt) {
+			// make sure the stmt must finish preparation first
+			<-s.prepared
+			if s.Stmt != nil {
+				_ = s.Close()
+			}
+		}(stmt)
 	}
 	sdb.Stmts = make(map[string]*Stmt)
 }
@@ -96,7 +101,12 @@ func (db *PreparedStmtDB) prepare(ctx context.Context, conn ConnPool, isTransact
 
 		return *stmt, nil
 	}
-
+	// check db.Stmts first to avoid Segmentation Fault(setting value to nil map)
+	// which cause by calling Close and executing SQL concurrently
+	if db.Stmts == nil {
+		db.Mux.Unlock()
+		return Stmt{}, ErrInvalidDB
+	}
 	// cache preparing stmt first
 	cacheStmt := Stmt{Transaction: isTransaction, prepared: make(chan struct{})}
 	db.Stmts[query] = &cacheStmt
--- a/prepare_stmt.go
+++ b/prepare_stmt.go
@@ -17,18 +17,16 @@ type Stmt struct {
 }
 
 type PreparedStmtDB struct {
-	Stmts       map[string]*Stmt
-	PreparedSQL []string
-	Mux         *sync.RWMutex
+	Stmts map[string]*Stmt
+	Mux   *sync.RWMutex
 	ConnPool
 }
 
 func NewPreparedStmtDB(connPool ConnPool) *PreparedStmtDB {
 	return &PreparedStmtDB{
-		ConnPool:    connPool,
-		Stmts:       make(map[string]*Stmt),
-		Mux:         &sync.RWMutex{},
-		PreparedSQL: make([]string, 0, 100),
+		ConnPool: connPool,
+		Stmts:    make(map[string]*Stmt),
+		Mux:      &sync.RWMutex{},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


prepare_stmt.go
```diff
@@ -48,12 +46,18 @@ func (db *PreparedStmtDB) Close() {
...
+			// make sure the stmt must finish preparation first
+			<-s.prepared
+			if s.Stmt != nil {
+				_ = s.Close()
+			}
+		}(stmt)
 	}
+	// I think setting it to nil should be better, as there should be no one
+	// writing into it after calling Close, but leave it for compatibility and safety
+	db.Stmts = make(map[string]*Stmt)
```
a631807682, line 60: Setting db.Stmts to nil is ok fo me, because `PreparedStmtDB` should not be allowed to be used after `Close`

prepare_stmt.go
```diff
@@ -48,12 +46,18 @@ func (db *PreparedStmtDB) Close() {
...
 
-	for _, query := range db.PreparedSQL {
-		if stmt, ok := db.Stmts[query]; ok {
-			delete(db.Stmts, query)
-			go stmt.Close()
-		}
+	for _, stmt := range db.Stmts {
+		go func(s *Stmt) {
+			// make sure the stmt must finish preparation first
+			<-s.prepared
```
a631807682, line 52: Can you describe what happens if we don't wait for `s.prepared`?

prepare_stmt.go
```diff
@@ -63,7 +67,6 @@ func (sdb *PreparedStmtDB) Reset() {
 	for _, stmt := range sdb.Stmts {
 		go stmt.Close()
 	}
```
a631807682, line 69: `Reset` will have similar problems as `Close`. Can we changed it?

prepare_stmt.go
ivila: while we prepare a new statement, the new Stmt register itself to the db.Stmt map (with sql.Stmt* field value nil), and unlock the mutex at line 100:
https://github.com/go-gorm/gorm/blob/4a50b36f638c6899089e6e3457425528ce693933/prepare_stmt.go#L98-L124

If we call Close function at this time(after line 101 and before line 119), don't wait for `s.prepared`, the condition of s.Stmt(the *sql.Stmt field) is unknown, it could be:

### 1) a nil pointer, and the preparation step not done yet(line 110), and successfully prepared after we call Close

in this scenario, we must not call Close of the Stmt as it is a null pointer, which result in a Stmt leak(the prepared Stmt generated by line 110 never called Close after that)

### 2) a nil pointer, and the preparation step not done yet(line 110), but failed to prepare after we call Close

in this scenario, we must not call Close of the Stmt as it is a null pointer too, but it is ok

So waiting for `s.prepared` to avoid Stmt leak.

prepare_stmt.go
ivila: > `Reset` will have similar problems as `Close`. Can we changed it?

OK, will do it.

prepare_stmt.go
```diff
@@ -48,22 +46,32 @@ func (db *PreparedStmtDB) Close() {
 	db.Mux.Lock()
 	defer db.Mux.Unlock()
```
a631807682, line 47: Dose write lock necessary? It seems that only `db.Stmts = nil` writes?

a631807682:
Can you add some unit tests in tests/prepared_stmt_test.go?

prepare_stmt.go
ivila: if we don't use write lock, then there would be data race condition, it can be easily proved by running following codes(run by `go run -race main.go`):
```golang
package main

import (
        //"fmt"
        "sync"
)

func main() {
        var m map[string]int
        var lock sync.RWMutex
        var wg sync.WaitGroup
        loopCount := 100000

        wg.Add(2)
        go func() {
                defer wg.Done()
                var val int
                for i := 0; i < loopCount; i++ {
                        lock.RLock()
                        val += m["not exist"]
                        lock.RUnlock()
                }
        }()
        go func() {
                defer wg.Done()
                for i := 0; i < loopCount; i++ {
                        lock.RLock()
                        m = make(map[string]int, 0)
                        lock.RUnlock()
                }
        }()

        wg.Wait()
}
```

a631807682: > if we don't use write lock, then there would be data race condition, it can be easily proved by running following codes(run by `go run -race main.go`):
> 
> ```go
> package main
> 
> import (
>         //"fmt"
>         "sync"
> )
> 
> func main() {
>         var m map[string]int
>         var lock sync.RWMutex
>         var wg sync.WaitGroup
>         loopCount := 100000
> 
>         wg.Add(2)
>         go func() {
>                 defer wg.Done()
>                 var val int
>                 for i := 0; i < loopCount; i++ {
>                         lock.RLock()
>                         val += m["not exist"]
>                         lock.RUnlock()
>                 }
>         }()
>         go func() {
>                 defer wg.Done()
>                 for i := 0; i < loopCount; i++ {
>                         lock.RLock()
>                         m = make(map[string]int, 0)
>                         lock.RUnlock()
>                 }
>         }()
> 
>         wg.Wait()
> }
> ```

I mean, can we use read lock in `for loop` and write lock only when assigning value.

ivila: No, I don't think we should, chances that some other goroutines could get a closed stmt and ran into the time gap between RUnlock and Lock (if we change codes to the following).
```golang
func (db *PreparedStmtDB) Close() {
	db.Mux.RLock()
	for _, stmt := range db.Stmts {
		go func(s *Stmt) {
			// make sure the stmt must finish preparation first
			<-s.prepared
			if s.Stmt != nil {
				_ = s.Close()
			}
		}(stmt)
	}
        db.Mux.RUnlock()

        db.Mux.Lock()
        defer db.Mux.Unlock()
	// setting db.Stmts to nil to avoid further using
	db.Stmts = nil
}
```
There will be some concurrency problems:
1）the prepare function use RLock too , some other goroutines can run into it while we are closing old Stmt, and they will get a closed/closing/valid Stmt(depends on the range progess)
image

2）there is a time gap between `db.Mux.RUnlock()` and `db.Mux.Lock()`, and chances that prepare function can get the write lock between this time gap(for example, if there are a goroutine waiting at Line 84), and they could get the WriteLock ahead of the Close function, then trying to insert a new Stmt, which result in Stmt leak:
image

ivila:
> Can you add some unit tests in tests/prepared_stmt_test.go?

Done, anything else?

a631807682 (approved):

Issue #7025, opened 2024-05-16
Executing count will clear Statement.Joins


liov:
https://github.com/go-gorm/playground/pull/736

```go
type AB struct {
	Id  int `json:"id" gorm:"primaryKey"`
	BId int `json:"aId" gorm:"index"`
	AId int `json:"bId"`
	B   B   `json:"b"`
	A   A   `json:"a"`
}

type A struct {
	Id   int    `json:"id" gorm:"primaryKey"`
	Name string `json:"name" gorm:"uniqueIndex"`
}

type B struct {
	Id   int    `json:"id" gorm:"primaryKey"`
	Name string `json:"name" gorm:"uniqueIndex"`
}

func main() {
	db, _ := gorm.Open(tests.DummyDialector{}, &gorm.Config{
		NamingStrategy: schema.NamingStrategy{
			SingularTable: true,
		},
	})

	var (
		count    int64
		appNodes = []*AB{}
	)
	db:= db.Model(&AB{}).Joins("A").Joins("B").Where(`id = ?`, 1)

	db.Count(&count)
	db.Limit(1).Offset(0).Order("id DESC").Find(&appNodes)
}
```
```
db := db.Model(&AB{}).Preload("A").Preload("B").Where(`id = ?`, 1)
```
The two pieces of code have different results, preload will get the correct result, because Count will call AfterQuery,it will "clear the joins after query because preload need it" ,I don't know why did it, this means need to write Joins twice.

ivila:
@liov What version are you using? 
``` golang
        var (
                count    int64
                appNodes = []*AB{}
        )
        db1 := db.Model(&AB{}).Joins("A").Joins("B").Where("abs.`id` = ?", 1)

        // [0.572ms] [rows:1] SELECT count(*) FROM `abs` LEFT JOIN `as` `A` ON `abs`.`a_id` = `A`.`id` LEFT JOIN `bs` `B` ON `abs`.`b_id` = `B`.`id` WHERE abs.`id` = 1
        db1.Debug().Count(&count)
[...]

PR #7150, opened 2024-08-09
fix: AfterQuery using safer right trim to clear FROM clause's joins

abhijeet45:
As a fix to [this issue](https://github.com/go-gorm/gorm/issues/7025) , clearing of db.Statement.Clauses["FROM"] was in introduced in [this pull-request](https://github.com/go-gorm/gorm/pull/7027)

At https://github.com/go-gorm/gorm/blob/master/callbacks/query.go#L291 , it can cause a panic if `db.Statement.Clauses["FROM"].Joins` slice is being sliced using negative upper bound index. Please observe below the case reproduced in current source code

image

In the case above, wrong column name `pets.names` (instead of pets.name) is passed as query.
When `Count()` is called after `Find()` on the the same `gorm.DB` instance with a wrong query, it calls AfterQuery twice which in turn tries to trim the slice twice which now is empty after first trim. Hence instead of obtaining actual SQL error, it causes a panic due to index out of bound error.

- [x] Do only one thing
- [x] Tested

This Pull Request:

- Safely trims the `db.Statement.Clauses["FROM"].Joins`
- Unit test for newly added functions for Safe right trimming of slice
- Unit Test for wrong query case

### User Case Description

User Case:

In one of my our projects, we are using gorm (just fantastic) as ORM support for Postgres. After upgrading gorm version from `v1.9.16` -> `v1.25.11`, we identified panic in API due to broken migration which changes a column name causing the query to be wrong.

Of course this is a minor fix requirement but much needed to prevent API server error and handle SQL error gracefully.

### DIFF
--- a/callbacks/query.go
+++ b/callbacks/query.go
@@ -288,7 +288,7 @@ func AfterQuery(db *gorm.DB) {
 	// clear the joins after query because preload need it
 	if v, ok := db.Statement.Clauses["FROM"].Expression.(clause.From); ok {
 		fromClause := db.Statement.Clauses["FROM"]
-		fromClause.Expression = clause.From{Tables: v.Tables, Joins: v.Joins[:len(v.Joins)-len(db.Statement.Joins)]} // keep the original From Joins
+		fromClause.Expression = clause.From{Tables: v.Tables, Joins: utils.RTrimSlice(v.Joins, len(v.Joins)-len(db.Statement.Joins))} // keep the original From Joins
 		db.Statement.Clauses["FROM"] = fromClause
 	}
 	if db.Error == nil && db.Statement.Schema != nil && !db.Statement.SkipHooks && db.Statement.Schema.AfterFind && db.RowsAffected > 0 {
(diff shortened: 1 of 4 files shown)


abhijeet45:
@jinzhu  Hello. Could you please review this PR and give further suggestions?

abhijeet45:
@liov Please kindly review. Please do let me know any suggestions. Thanks.

callbacks/query.go
```diff
@@ -288,7 +288,7 @@ func AfterQuery(db *gorm.DB) {
 	// clear the joins after query because preload need it
 	if v, ok := db.Statement.Clauses["FROM"].Expression.(clause.From); ok {
 		fromClause := db.Statement.Clauses["FROM"]
-		fromClause.Expression = clause.From{Tables: v.Tables, Joins: v.Joins[:len(v.Joins)-len(db.Statement.Joins)]} // keep the original From Joins
+		fromClause.Expression = clause.From{Tables: v.Tables, Joins: utils.RTrimSlice(v.Joins, len(v.Joins)-len(db.Statement.Joins))} // keep the original From Joins
```
abhijeet45, line 291: utils.RTrimSlice makes sure no panic when `v.Joins` is empty while `db.Statement.Joins` is not

PR #7154, opened 2024-08-12
Fix join in delete

mtsoltan:
- [x] Do only one thing
- [x] Tested

In MySQL, delete supports inner joins (but not left joins).
This PR does two things to make it possible for us to use `InnerJoin` to build a clause into our delete SQL statements.
1. Abstract the join logic from `query.go` into `join.go` to allow it to be re-used in delete (now) and possibly in update (in future work).
2. Use the join logic in delete, and add table name after the `DELETE ` keyword.

The problems with this PR:
1. It does not check if the join is an `InnerJoins` or a `Joins` - so this will build `DELETE t FROM t LEFT JOIN r WHERE ...;` statements if it is passed a normal left join - when this is invalid SQL. The logic should be upgraded to silently discard the join, or emit an error that says `"You cannot left join in a delete statement. Maybe you meant to inner join instead?"`.
2. Tries to make the least amount of changes possible. This is why the introduced `HandleJoins` function takes two callback arguments - to make it very faithful to what initially happened in `query.go` and ensure no regressions.
3. It does not introduce a test. I tried it out and it works, but it should introduce a test.

I'm reluctant to solve those three problems until a contributor goes through the logic and verifies that the work done is meaningful. If that's the case, we can have a discussion about the organization of the code, and error handling / silent discarding of clause for left joins, and once the logic is finalized, I'll introduce tests for it.

### User Case Description

```
// Delete a ModelB based on the value of a column in TableA
db.InnerJoins("TableA").Where("TableA.columnA in ?", values).Delete(&ModelB { ID: someID });
```

### DIFF
--- a/callbacks/delete.go
+++ b/callbacks/delete.go
@@ -126,7 +126,19 @@ func Delete(config *Config) func(db *gorm.DB) {
 
 		if db.Statement.SQL.Len() == 0 {
 			db.Statement.SQL.Grow(100)
-			db.Statement.AddClauseIfNotExists(clause.Delete{})
+
+			deleteClause := clause.Delete{}
+
+			HandleJoins(
+				db,
+				func(db *gorm.DB) {
+					deleteClause.Modifier = db.Statement.Table
+				},
+				func(db *gorm.DB, tableAliasName string, join gorm.Join, relation *schema.Relationship) {
+				},
+			)
+
+			db.Statement.AddClauseIfNotExists(deleteClause)
 
 			if db.Statement.Schema != nil {
 				_, queryValues := schema.GetIdentityFieldValuesMap(db.Statement.Context, db.Statement.ReflectValue, db.Statement.Schema.PrimaryFields)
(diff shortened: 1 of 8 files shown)


a631807682 (commented):
`Delete/Update Join` does not seem to be part of the sql standard, but an implementation of mysql. As far as I know, sqlite/postgres do not support it (they associate other tables through `From`).
I don't recommend changing it for specific driver, I think we can add `clause` to support it, and if the new clause requires internal functions similar to `HandleJoins`, then exporting these functions is also possible.
What do you think?

mtsoltan:
Thank you for the very swift reply! 🙇 

If it's only supported on one dialect, I agree with you, it makes the most sense to add a clause to support it. Can you point me in the right direction of what I need to take care of to make sure that my work in adding a clause is of high enough quality to be merged?

I'm also interested in what we'd call the clause. Since it's also a delete clause (just one that has an inner join). So maybe `DELETE_JOIN`? - added by the function `Delete()` on a statement that already has an inner join?

I can look into this myself, but if you have the extra time to spare, I'd appreciate being pointed to where I can make this dialect-specific to just MySQL, if such a place does exist.

a631807682:
> So maybe DELETE_JOIN? - added by the function Delete() on a statement that already has an inner join?

I think yes, we can change the existing behavior through `StatementModifier` just like `SoftDeleteQueryClause`. I am not sure if it is possible or if it requires support from internal functions.

PR #7166, opened 2024-08-21
feat: `BeforeUpdate` hook supports update using struct field value

demoManito:
- [x] Do only one thing
- [x] Tested

In the `BeforeUpdate` hook function, if the structure field value is updated and the field has the `gorm:"column:field_name"` set, the field is updated in SQL

### User Case Description
- Code:
```go
type User struct {
	ID      int64 `gorm:"column:id;primary_key"`
	Version int64 `gorm:"column:version"`
	UserID  int64 `gorm:"column:user_id"`
}

// BeforeUpdate gorm hook
func (u *User) BeforeUpdate(*gorm.DB) error {
	u.Version = time.Now().Unix()
	return nil
}

func Update() error {
	return DB().Model(&User{}).Where("id = ?", 1).Update("user_id", 312123).Error
}
```
- Expected Output:
```sql
UPDATE `user` SET `version`=1724247677,`user_id`=312123 WHERE id = 1
```

### DIFF
--- a/callbacks/update.go
+++ b/callbacks/update.go
@@ -32,22 +32,50 @@ func SetupUpdateReflectValue(db *gorm.DB) {
 // BeforeUpdate before update hooks
 func BeforeUpdate(db *gorm.DB) {
 	if db.Error == nil && db.Statement.Schema != nil && !db.Statement.SkipHooks && (db.Statement.Schema.BeforeSave || db.Statement.Schema.BeforeUpdate) {
-		callMethod(db, func(value interface{}, tx *gorm.DB) (called bool) {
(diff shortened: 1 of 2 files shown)


jinzhu:
It's too tricky; it's not recommended to use Hooks for updating values. Instead, it's better to use SQL expressions.

demoManito:
> It's too tricky; it's not recommended to use Hooks for updating values. Instead, it's better to use SQL expressions.

ok thanks

demoManito:
> It's too tricky; it's not recommended to use Hooks for updating values. Instead, it's better to use SQL expressions.

ok thanks

Issue #7173, opened 2024-08-28
Duplicate savepoint names resulting in partial rollbacks in nested transactions


phroggyy:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/757

## Description

When you wrap `db.Transaction`, go will only put that closure in memory _once_:
```
func Transaction(ctx context.Context, db *gorm.DB, apply func(ctx context.Context, db *gorm.DB) error) error {
	return db.
		WithContext(ctx).
		Transaction(func(db *gorm.DB) error {
			return apply(ctx, db)
		})
}
```

This means that the inner function passed to `Transaction` always has the same memory address.

This causes issues in https://github.com/go-gorm/gorm/blob/master/finisher_api.go#L626, where we use the memory address to decide the savepoint name.

With the current implementation, the following statements execute:

```
2024/08/28 11:32:41 /Users/leosjoberg/Code/open-source/gorm-playground/main_test.go:17 record not found
[0.155ms] [rows:0] SELECT * FROM `users` WHERE name = "jinzhu" AND `users`.`deleted_at` IS NULL ORDER BY `users`.`id` LIMIT 1

2024/08/28 11:32:41 /Users/leosjoberg/Code/Go/pkg/mod/gorm.io/driver/sqlite@v1.5.6/sqlite.go:237
[0.021ms] [rows:0] SAVEPOINT sp0x10050efe0

2024/08/28 11:32:41 /Users/leosjoberg/Code/open-source/gorm-playground/main_test.go:31
[1.206ms] [rows:1] INSERT INTO `users` (`created_at`,`updated_at`,`deleted_at`,`name`,`age`,`birthday`,`company_id`,`manager_id`,`active`) VALUES ("2024-08-28 11:32:41.299","2024-08-28 11:32:41.299",NULL,"jinzhu",0,NULL,NULL,NULL,false) RETURNING `id`
User created: 1
2024/08/28 11:32:41 /Users/leosjoberg/Code/Go/pkg/mod/gorm.io/driver/sqlite@v1.5.6/sqlite.go:237
[0.011ms] [rows:0] SAVEPOINT sp0x10050efe0

2024/08/28 11:32:41 /Users/leosjoberg/Code/open-source/gorm-playground/main_test.go:45 record not found
[0.143ms] [rows:0] SELECT * FROM `accounts` WHERE `accounts`.`id` = 1 AND `accounts`.`deleted_at` IS NULL ORDER BY `accounts`.`id` LIMIT 1

[...]

PR #7174, opened 2024-08-28
Generate unique savepoint names for nested transactions

phroggyy:
- [x] Do only one thing
- [x] Tested

This test changes `SAVEPOINT` names from relying on the memory address of the callback (which can cause unexpected issues with partial rollbacks), to using a random integer.

The integer is generated using `maphash.Hash`, which is used as that internally relies on `runtime.fastrand`, which is the most performant way to get a random integer.

I've ensured the test fails without my code changes, and passes with them applied.

### Use Case Description

Covered in https://github.com/go-gorm/gorm/issues/7173

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -4,6 +4,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"hash/maphash"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


a631807682 (approved):

daheige (approved):
LGTM

demoManito (approved):

finisher_api.go
```diff
@@ -623,14 +624,15 @@ func (db *DB) Transaction(fc func(tx *DB) error, opts ...*sql.TxOptions) (err er
 	if committer, ok := db.Statement.ConnPool.(TxCommitter); ok && committer != nil {
 		// nested transaction
 		if !db.DisableNestedTransaction {
-			err = db.SavePoint(fmt.Sprintf("sp%p", fc)).Error
+			spID := new(maphash.Hash).Sum64()
```
loeffel-io, line 627: instead of sprintf two times for the same id we could just do: 

spID := fmt.Sprintf("sp%d", new(maphash.Hash).Sum64())

PR #7207, opened 2024-09-25
Fixed "*DB.Raw()" bug where single or double quotes containing '@' were incorrectly identified as named variables

Rankgice:
- [*] Do only one thing
- [*] Tested

Fixed "*DB.Raw()"  has a bug where single or double quotes containing '@' are incorrectly recognized as named variables.
When I call the Raw method, if the SQL contains a quote wrapped ` @ `, then this ` @ ` will not be treated as part of the string, but as a named parameter. I have increased the judgment of quotation marks to ensure that the content inside the quotation marks will not be recognized as distinctive symbols.

### User Case Description
Values
``` go
values := map[string]interface{}{
	"email": `%qq.com`,
}
```
Origin
``` sql
select * from users where email like @email and name = "@email"
```
Result
``` shell
D:/go/src/runtime/proc.go:271 sql: expected 1 arguments, got 2
[21.008ms] [rows:-] select * from users where email like '%qq.com' and name = "'%qq.com'"
```

#### After repair

Result
``` shell
D:/go/src/runtime/proc.go:271
[29.642ms] [rows:0] select * from users where email like '%qq.com' and name = "@email"
```

### DIFF
--- a/clause/expression.go
+++ b/clause/expression.go
@@ -88,6 +88,7 @@ func (expr NamedExpr) Build(builder Builder) {
 		inName           bool
 		afterParenthesis bool
 		namedMap         = make(map[string]interface{}, len(expr.Vars))
+		quotationMarks   byte
 	)
 
 	for _, v := range expr.Vars {
@@ -124,6 +125,14 @@ func (expr NamedExpr) Build(builder Builder) {
 	name := make([]byte, 0, 10)
 
 	for _, v := range []byte(expr.SQL) {
+		if quotationMarks == v && (v == '"' || v == '\'') {
+			quotationMarks = 0
+		} else if quotationMarks == 0 && (v == '"' || v == '\'') {
+			quotationMarks = v
+		} else if quotationMarks == '"' || quotationMarks == '\'' {
+			builder.WriteByte(v)
+			continue
+		}
 		if v == '@' && !inName {
 			inName = true
 			name = name[:0]
--- a/tests/raw_test.go
+++ b/tests/raw_test.go
@@ -1,0 +1,110 @@
+package tests_test
+
+import (
+	. "gorm.io/gorm/utils/tests"
+	"testing"
+)
+
+func TestRawSelect(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Rankgice:
The failure of this test seems to be due to an issue with the test image container. I noticed that other PRs have also encountered the same problem as me.

jinzhu:
Hi, can you add some tests?

Rankgice:
> Hi, can you add some tests?

Hello, of course not a problem. I will add some tests later

Rankgice:
> Hi, can you add some tests?

I have added some test cases. May I proceed with the next review

clause/expression.go
```diff
@@ -124,6 +125,14 @@ func (expr NamedExpr) Build(builder Builder) {
...
 	for _, v := range []byte(expr.SQL) {
+		if quotationMarks == v && (v == '"' || v == '\'') {
+			quotationMarks = 0
+		} else if quotationMarks == 0 && (v == '"' || v == '\'') {
+			quotationMarks = v
+		} else if quotationMarks == '"' || quotationMarks == '\'' {
+			builder.WriteByte(v)
+			continue
+		}
 		if v == '@' && !inName {
```
jinzhu, line 136: I think it would be better to add a check here to ensure that the previous character is not a " or '

Rankgice: The logic of this code is
**The second if:**
When we first encounter '' or '', quotationMarks will be marked as' 'or' 'accordingly`
**The first if:**
When we have already been marked and encountered that mark (end mark) again, we will clear the mark
**The third if:**
When we have been marked and have not encountered the end mark, we will output it as it is

I didn't understand where you wanted to add a judgment

daheige (approved):
LGTM!

Messhias (approved):

clause/expression.go
daheige: > I think it would be better to add a check here to ensure that the previous character is not a " or '

good!

PR #7208, opened 2024-09-27
fix: deterministic index ordering when migrating

bamo:
- [x] Do only one thing
- [x] Tested

Change ParseIndexes to return a list of indices rather than a map, so the callers will iterate in deterministic order.  This means when re-creating a database from the same schema, indexes will be created in a consistent order.

### User Case Description

Issue: We observed that, when creating a database based on the same gORM schema multiple times, indexes could appear in different orders, hurting determinism for use-cases like schema comparison.

### DIFF
--- a/schema/index.go
+++ b/schema/index.go
@@ -27,8 +27,9 @@ type IndexOption struct {
 }
 
 // ParseIndexes parse schema indexes
-func (schema *Schema) ParseIndexes() map[string]Index {
-	indexes := map[string]Index{}
+func (schema *Schema) ParseIndexes() []*Index {
+	indexesByName := map[string]*Index{}
+	indexes := []*Index{}
 
 	for _, field := range schema.Fields {
 		if field.TagSettings["INDEX"] != "" || field.TagSettings["UNIQUEINDEX"] != "" {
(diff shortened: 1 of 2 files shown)


bamo:
If the maintainers are happy with this, I think it makes sense as well to do this for ParseUniqueConstraints and ParseCheckConstraints as well.

I'm happy to do those as follow-on PRs as well. Just let me know.

schema/index.go
```diff
@@ -38,7 +39,12 @@ func (schema *Schema) ParseIndexes() map[string]Index {
 				break
 			}
 			for _, index := range fieldIndexes {
-				idx := indexes[index.Name]
+				idx := indexesByName[index.Name]
```
ChenJesse, line 42: hey i think this is a little complex, mind moving it into a helper

bamo:
The test failure seems like a flake or something (container failed to start).  Looks like I don't have permissions to re-run it.

bamo:
@jinzhu What do you think about this?

aashishkapur:
I have the same issue - would be good to merge this fix.

jinzhu:
Hi @bamo 

Can you rebase the master branch, it might fix the tests.

bamo:
@jinzhu done!

bamo:
Hey @jinzhu, I don't mean to bother you here, just wondering if there's anything else you'd like me to do for this or if you think we can land this soon.  Thanks!

wookie0 (approved):
LGTM

PR #7216, opened 2024-09-30
feature(finisher_api): add Exists method

demoManito:
- [x] Do only one thing
- [x] Tested

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -502,6 +502,12 @@ func (db *DB) Count(count *int64) (tx *DB) {
 	return
 }
 
+// Exists checks if there is any record matching the given conditions
+func (db *DB) Exists() (bool, error) {
+	var exists bool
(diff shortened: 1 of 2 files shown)


demoManito:
Hi, can I add this method? If so, is it reasonable to add it this way, or should it be placed in the driver?

jinzhu:
Maybe we should place it in a helper package, keeping GORM itself clean and avoiding adding too much API directly.

demoManito:
> Maybe we should place it in a helper package, keeping GORM itself clean and avoiding adding too much API directly.

Do you mean the helper package under the go-gorm organization? @jinzhu

jinzhu:
For now, we probably won’t maintain an additional helper package. It could be placed in a separate third-party library for the time being.

demoManito:
> For now, we probably won’t maintain an additional helper package. It could be placed in a separate third-party library for the time being.

ok

PR #7220, opened 2024-10-03
Allow session config naming override

AdamDrewsTR:
- [x] Do only one thing
- [x] Tested

Allows naming strategy to change on the fly. This is helpful when you may want to dynamically alter the naming strategy for individual struct without changing the behavior everywhere.

### User Case Description

Change naming strategy for individual structs/db transactions/sessions.


jinzhu:
Hi @AdamDrewsTR 

Since we cache the generated values for each model, this may not work in some cases.

PR #7242, opened 2024-10-16
use map look-up for indexes

abbyssoul:
- [ X] Do only one thing
- [ X] Non breaking API changes
- [ X] Tested

Changing the implementation of `schema.LookIndex` to use map's two-value assignment tests for the existence of a key when looking for an index before switching to field lookup for performance.

### User Case Description

The existing implementation of `Schema.LookIndex` uses linear search (in the best case) instead of using map test using const-time before falling back to a linear scan.
The new implementation has a performance advantage for cases with many indexes.


a631807682 (approved):

daheige (approved):
LGTM!

PR #7263, opened 2024-11-02
fix: update incorrect default name when parse dest model field.

zyb-992:
- [x] Do only one thing
- [ ] Tested

it correctly update Field.HasDefaultValue field value when using Schema.ParseField method to init the field.


zyb-992:
oh! it's my fault, this method will rewrite thie field value after this operation, i did not read this until now...

Issue #6913, opened 2024-03-19
Multiple calls to the Model method update the query table name only on the first call


ZiRunHua:
### playground
https://github.com/go-gorm/playground/pull/704
### problem
When a Model method is called multiple times, only the first call queries the table, and subsequent calls to the Model method do not update the query table.
Probably not many people use it that way. The reason we use query criteria to query multiple tables is that, in some cases, we set the same fields in multiple tables by embedding `struct`, which also makes the same query criteria available in all tables. We do this to reduce duplication of logic and simplify the code.

ZiRunHua:
It seems that no one else has encountered this problem, but I think this problem needs to be fixed, because the Model method is closely related to the factory method, such as the factory method actually used by the following business:
```go
// ForeignKeyCondition 交易外键查询条件 用于交易记录和统计的查询
type ForeignKeyCondition struct {
    AccountId   uint
    UserIds     *[]uint
    CategoryIds *[]uint
}

func (f *ForeignKeyCondition) addConditionToQuery(db *gorm.DB) *gorm.DB {
    query := db.Where("account_id = ?", f.AccountId)
    if f.UserIds != nil {
        query = query.Where("user_id IN (?)", *f.UserIds)
    }
    if f.CategoryIds != nil {
        query = query.Where("category_id IN (?)", *f.CategoryIds)
    }
    return query
}

type statisticModel interface {
    GetUpdatesValue(amount, count int) map[string]interface{}
    GetDate(tradeTime time.Time) time.Time
}

func (f *ForeignKeyCondition) GetStatistic(ie constant.IncomeExpense) statisticModel {
    if ie == constant.Income {
        return f.getIncomeStatisticModel()
    } else {
        return f.getExpendStatisticModel()
    }
}

func (f *ForeignKeyCondition) getIncomeStatisticModel() statisticModel {
    if f.CategoryIds == nil {
        if f.UserIds == nil {
            return &IncomeAccountStatistic{}
        } else {
            return &IncomeAccountUserStatistic{}
        }
    } else if f.UserIds == nil {
        return &IncomeCategoryStatistic{}
    }
    return nil
}

func (f *ForeignKeyCondition) getExpendStatisticModel() statisticModel {
    if f.CategoryIds == nil {
        if f.UserIds == nil {
            return &ExpenseAccountStatistic{}
        } else {
            return &ExpenseAccountUserStatistic{}
        }
    } else if f.UserIds == nil {
[...]

PR #7266, opened 2024-11-05
fix: table not change (#6913)

ZiRunHua:
- [x] Do only one thing
- [x] Tested

__Update Condition for Table Selection:__ This change allows `ParseWithSpecialTableName` to re-evaluate and update the table name based on the value passed, not just when the table name is initially empty.

### User Case Description
When you use embedded structures, you might write queries like this
```go
type StatisticBase struct {
    Date          time.Time
    Amount, Count int
}

type IncomeStatistic struct {
    StatisticBase
    Tas    int
    UserId uint
}

type ExpendStatistic struct {
    StatisticBase
    UserId uint
}

type Total struct{ Amount, Count int }

type Statistic struct{ Income, Expend Total }

func queryStatistic(db *gorm.DB, start, end time.Time) (result Statistic, err error) {
    query := db.Where("date BETWEEN ? AND ?", start, end)
    query = query.Select("SUM(amount) as Amount,SUM(count) as Count")
    err = query.Model(&IncomeStatistic{}).Scan(&result.Income).Error
    if err != nil {
        return
    }
    err = query.Model(&ExpendStatistic{}).Scan(&result.Expend).Error
    return
}
```
But the table for the second query did not change.

```mysql
SELECT SUM(amount) as Amount,SUM(count) as Count FROM `income_statistic` WHERE date BETWEEN '2024-11-03 09:38:53.615' AND '2024-11-08 09:38:53.615';
SELECT SUM(amount) as Amount,SUM(count) as Count FROM `income_statistic` WHERE date BETWEEN '2024-11-03 09:38:53.615' AND '2024-11-08 09:38:53.615';
```
Since the table is only updated when empty, this is what the "pull request" wants to change, which will give the user more flexibility to manipulate the data as needed.

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -490,7 +490,7 @@ func (stmt *Statement) Parse(value interface{}) (err error) {
 }
 
 func (stmt *Statement) ParseWithSpecialTableName(value interface{}, specialTableName string) (err error) {
-	if stmt.Schema, err = schema.ParseWithSpecialTableName(value, stmt.DB.cacheStore, stmt.DB.NamingStrategy, specialTableName); err == nil && stmt.Table == "" {
+	if stmt.Schema, err = schema.ParseWithSpecialTableName(value, stmt.DB.cacheStore, stmt.DB.NamingStrategy, specialTableName); err == nil {
 		if tables := strings.Split(stmt.Schema.Table, "."); len(tables) == 2 {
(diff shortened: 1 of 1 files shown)


ZiRunHua:
Oh, the test didn't pass. Maybe it shouldn't be used that way.

PR #7276, opened 2024-11-12
feat(expression): support case-when expression

minhtri1396:
- [X] Do only one thing
- [X] Tested

Implement `case-when` expression to use `case-when` in SQL `query`.

### User Case Description
We can use this expression to update batches instead of single update.
```Golang
userIDs := make([]string, len(inputUsers))
userNameCases := make([]*clause.ExprCaseCondition, len(inputUsers))
for idx, user := range inputUsers {
    userIDs[idx] = user.ID
    userNameCases[idx] = &clause.ExprCaseCondition{
        When: "user_id=?",
        Then: "?",
        Vars: []any{
            user.ID,
            user.Name,
        },
    }
}

db.
    Table("users").
    Where("user_id IN (?)", userIDs).
    UpdateColumns(map[string]any{
        "user_name": clause.ExprCase{
            Cases: userNameCases,
            Else: &clause.ExprCaseElse{
            Then: "user_name",
                Vars: nil,
            },
        },
    })
```

The generated SQL will be:
```
UPDATE `users`
SET `user_name`=CASE
    WHEN user_id="user-001" THEN "user-name-001"
    WHEN user_id="user-002" THEN "user-name-002"
    ELSE user_name
END
WHERE user_id IN ("user-001","user-002")
```

### DIFF
--- a/clause/expr_case.go
+++ b/clause/expr_case.go
@@ -1,0 +1,54 @@
+package clause
+
+type ExprCaseCondition struct {
+	When string
+	Then string
+	Vars []any
+}
+
+type ExprCaseElse struct {
+	Then string
+	Vars []any
+}
+
+type ExprCase struct {
+	Cases []*ExprCaseCondition
+	Else  *ExprCaseElse
+}
+
+func (expr ExprCase) Name() string {
+	return "CASE"
+}
+
+func (expr ExprCase) Build(builder Builder) {
+	var vars []any
+	for idx, condition := range expr.Cases {
+		if idx > 0 {
+			_ = builder.WriteByte(' ')
+		}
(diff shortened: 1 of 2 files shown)


jinzhu:
Could we design the API to look something like this and place it in the datatypes package?

casewhen := datatypes.When("name = ?", "123").Then("hello").
            When("name = ?", "234").Then("hello2").
            Else("hello3")

db.Model(&user).Where("id = ?", 1).Update(map[string]any{"name": casewhen})

Messhias (approved):

jinzhu:
Closing this PR for now. Let me know once it’s published to go-gorm/datatypes!

Issue #7297, opened 2024-12-02
Сoncurrent map writes while parsing table


golandprog:
While running the code below, there is a chance of running into an error "`fatal error: concurrent map writes`"
It happens suddenly and need to use loop running until it fails

```bash
while go test -bench=. -benchtime=0.1s; do :; done
```
```
go 1.22.6

require (
	github.com/google/uuid v1.6.0
	gorm.io/driver/sqlite v1.5.6
	gorm.io/gorm v1.25.12
)

require (
	github.com/jinzhu/inflection v1.0.0 // indirect
	github.com/jinzhu/now v1.1.5 // indirect
	github.com/mattn/go-sqlite3 v1.14.22 // indirect
	golang.org/x/text v0.20.0 // indirect
)

```
```go
package main

import (
	"github.com/google/uuid"
	"gorm.io/driver/sqlite"
	"gorm.io/gorm"
	"gorm.io/gorm/schema"
	"math/rand/v2"
	"sync"
	"testing"
)

type Data1 struct {
	UUID uuid.UUID `gorm:"primaryKey;type:uuid"`
}

type Data2 struct {
	UUID uuid.UUID `gorm:"primaryKey;type:uuid"`
}

type Data3 struct {
	UUID uuid.UUID `gorm:"primaryKey;type:uuid"`
}

func BenchmarkParse(b *testing.B) {
	cacheStore := &sync.Map{}
	db, err := gorm.Open(sqlite.Open(":memory:"), &gorm.Config{})
	if err != nil {
		b.Fatalf("failed to connect database")
	}

	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			switch rand.IntN(2) {
			case 0:
				_, _ = schema.ParseWithSpecialTableName([]*struct {
					Data1
					Data2 Data2 `gorm:"foreignKey:UUID;references:UUID"`
				}{}, cacheStore, db.NamingStrategy, "")
			case 1:
				_, _ = schema.ParseWithSpecialTableName([]*struct {
					Data1
					Data2 Data2 `gorm:"foreignKey:UUID;references:UUID"`
					Data3 Data3 `gorm:"foreignKey:UUID;references:UUID"`
				}{}, cacheStore, db.NamingStrategy, "")
			}
		}
	})
}
```
```
fatal error: concurrent map writes

goroutine 87 [running]:
gorm.io/gorm/schema.(*Schema).parseRelation(0xc00041e1e0, 0xc000420600)
[...]

PR #7298, opened 2024-12-02
Fix concurrent map writes #7297

Ponywka:
- [x] Do only one thing
- [ ] Tested

Fixes `fatal error: concurrent map writes` error (https://github.com/go-gorm/gorm/issues/7297)

### User Case Description
Using normalized tables

### DIFF
--- a/schema/relationship.go
+++ b/schema/relationship.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync"
 
 	"github.com/jinzhu/inflection"
 	"golang.org/x/text/cases"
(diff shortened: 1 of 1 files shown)


schema/relationship.go
```diff
@@ -32,6 +33,8 @@ type Relationships struct {
 	Relations map[string]*Relationship
 
 	EmbeddedRelations map[string]*Relationships
+
+	Mux *sync.RWMutex
```
jinzhu, line 37: thank you for your PR, let’s use a non-pointer value to avoid potential nil panics.

golandprog: Done!

PR #7333, opened 2025-01-04
fix(limit): set default limit to math.MaxInt when only offset is provided

### LINKED ISSUE #7302: Using only offset instead of limit
## Your Question

When I use offset but not limit, the SQL executed by Gorm only includes offset and not limit. Unfortunately, not all databases support using only offset without limit. In MySQL, limit and offset need to appear in pairs, and MySQL will report an error at this time. So I think this may be a problem, and I think setting the limit to "math. MaxInt" when the user only specifies offset can solve this problem. 

## The document you expected this should be explained

I want to know if this error was tacitly approved by the developer

## Expected answer

If it is an unexpected bug for the developer, I can submit a PR.

aydinomer00:
## Description
This PR fixes an issue where MySQL requires both LIMIT and OFFSET to be provided when using OFFSET. When only OFFSET is specified, the LIMIT is now automatically set to math.MaxInt.

## Changes
- Updated Limit clause to use math.MaxInt when only offset is provided
- Updated existing tests to reflect new behavior
- Added new test cases to verify the behavior

## Related Issue
Fixes #7302

## Testing
- Added new test case for verifying behavior when only offset is provided
- All existing tests pass
- Manually tested with MySQL

### DIFF
--- a/clause/limit.go
+++ b/clause/limit.go
@@ -1,18 +1,28 @@
 package clause
 
-// Limit limit clause
+import (
+	"math"
+)
+
+// Limit represents a limit clause
 type Limit struct {
 	Limit  *int
 	Offset int
 }
 
-// Name where clause name
+// Name returns the name of the clause ("LIMIT")
 func (limit Limit) Name() string {
 	return "LIMIT"
 }
 
-// Build build where clause
+// Build constructs the LIMIT clause
 func (limit Limit) Build(builder Builder) {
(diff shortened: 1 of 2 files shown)


jinzhu:
I suggest keeping it this way; returning an error might be more reasonable. We shouldn’t set an unreasonable limit value as it could mask and lead to a more serious issue.

PR #7361, opened 2025-01-26
fix: prevent interface type array from causing runtime errors

KEHyeon:
- [ ] Do only one thing
- [v] Tested

fix: prevent interface type array from causing runtime errors
It has been modified to find out the value when it is an interface array.

### User Case Description

``` go
type UserInterface interface {
	GetName() string
}

db.Table("users").Create([]UserInterface{UserToUserInterface(&userData)})
```
In this case, create is allowed.

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -19,10 +19,7 @@ func (db *DB) Create(value interface{}) (tx *DB) {
 	if db.CreateBatchSize > 0 {
 		return db.CreateInBatches(value, db.CreateBatchSize)
 	}
-
-	tx = db.getInstance()
-	tx.Statement.Dest = value
-	return tx.callbacks.Create().Execute(tx)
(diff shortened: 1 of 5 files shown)


daheige (approved):
lgtm

KEHyeon (commented):
thankyou

KEHyeon:
> Can you add some tests?

Can I just write test code?

a631807682:
> > Can you add some tests?
> 
> Can I just write test code?

You can add unit tests to show what this PR fixes. https://github.com/go-gorm/gorm/blob/master/tests/create_test.go

KEHyeon:
I guess I was wrong. If I change it one by one like this, the sideEffect seems to be too big. I thought of another way, so I'll upload the pr again.

PR #7370, opened 2025-02-16
feat: add BeforeFind hook

### LINKED ISSUE #7357: BeforeFind
## Describe the feature

Create a BeforeFind hook along the same lines as the other Before hooks, but for Find.

## Motivation

We are implementing a permissions system and using hooks to gate access. Some users will not have even read access to some data and therefore, even a find query should return an error. We can do this in AfterFind, but then we've already created load and overhead that is completely unnecessary. It would be much more efficient to reject before the query even happens.

MhmdGol:
- [x] Do only one thing
- [x] Tested

Added BeforeFind hook capability to the queries. Closes #7357 

### User Case Description

From now on Find, First, Take and so on can have before hook.
It will be executed before executing query itself.

```go
type Product struct {
    Code string
    Price float64
}

func (s *Product) BeforeFind() error {
    // do something before query
}

### DIFF
--- a/callbacks/callbacks.go
+++ b/callbacks/callbacks.go
@@ -48,6 +48,7 @@ func RegisterDefaultCallbacks(db *gorm.DB, config *Config) {
 	createCallback.Clauses = config.CreateClauses
 
 	queryCallback := db.Callback().Query()
+	queryCallback.Register("gorm:before_query", BeforeQuery)
 	queryCallback.Register("gorm:query", Query)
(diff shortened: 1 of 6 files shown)


MhmdGol:
@a631807682 and @daheige 
Can you please review?

mrubiottec:
Looking forward to getting this added in GORM. Is there an update on this?

jinzhu:
We won’t be accepting this proposal, as it could lead to some misuse.

Thank you for your PR.

Issue #7294, opened 2024-11-26
Using clause.Not and clause.And result in erroneous simplification


ezk84:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/767

## Description

When using clause.Not and clause.And to create query Where conditions and faulty simplification is applied:

NOT(col1 = val1 AND col2 = val2) gets turned into (col1 <> val1 AND col2 <> val2), which is wrong.

The right simplification is to turn the AND into an OR, by [De Morgan's Laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws), so (col1 <> OR val1 AND col2 <> val2)

This is a regression that was introduced into gorm in 1.25.6. Testing against 1.25.5 returns the correct condition and expected rows in the included tests (see gorm playground PR).

Test for completeness

```go
func TestGORM(t *testing.T) {
	DB.Create(&User{Name: "jinzhu", Active: true})
	DB.Create(&User{Name: "ezequiel", Active: true})
	DB.Create(&User{Name: "jinzhu", Active: false})
	DB.Create(&User{Name: "ezequiel", Active: false})

	var results []User
	conds := clause.Not(clause.And(
		clause.Eq{
			Column: clause.Column{Name: "name"},
			Value:  "jinzhu",
		},
		clause.Eq{
			Column: clause.Column{Name: "active"},
			Value:  true,
		},
	))
	// Query with `NOT(name = 'jinzhu' AND active = true)` should return last three records
	// By De Morgan's Law this is equivalent to `name != 'jinzhu' OR active != true`
	// but current version (since 1.25.6) transforms this to
	// `name != 'jinzhu' AND active != true` which returns only the last record

	err := DB.
		Model(&User{}).
		Clauses(clause.Where{Exprs: []clause.Expression{conds}}).
		Find(&results).
		Error
	if err != nil {
		t.Errorf("Failed, got error: %v", err)
	}
	if len(results) != 3 {
		t.Errorf("Failed, expected 3 records, got %v", len(results))
	}
	for _, r := range results {
		if r.Name == "jinzhu" && r.Active {
			t.Errorf("Failed, unexpected record: %v", r)
		}
	}

	conds = clause.Not(clause.Or(
		clause.Eq{
			Column: clause.Column{Name: "name"},
			Value:  "jinzhu",
		},
		clause.Eq{
			Column: clause.Column{Name: "active"},
			Value:  true,
		},
	))
[...]

PR #7371, opened 2025-02-16
fix: make NotConditions obey De Morgan's laws

molon:
- [x] Do only one thing
- [x] Tested

Resolve this issue: https://github.com/go-gorm/gorm/issues/7294

This should be a breaking change, but I think the previous logic is a bug. 
Or is it because I don't understand the role of this condition?

### DIFF
--- a/clause/where.go
+++ b/clause/where.go
@@ -181,7 +181,7 @@ func (not NotConditions) Build(builder Builder) {
 
 		for idx, c := range not.Exprs {
 			if idx > 0 {
-				builder.WriteString(AndWithSpace)
+				builder.WriteString(OrWithSpace)
 			}
 
 			if negationBuilder, ok := c.(NegationExpressionBuilder); ok {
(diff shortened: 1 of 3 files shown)


jinzhu:
This change does make sense, but since it is a long-standing behavior, we’d prefer to keep things as they are to avoid breaking existing user systems. Thank you for your PR!

ezk84:
@jinzhu this was a bug introduced in gorm in 1.25.6 (released Jan 12, 2024). 

Testing against 1.25.5 (released Oct 10, 2023) returns the correct condition and expected rows in the included tests (see https://github.com/go-gorm/playground/pull/767).

This doesn't seem like designed behaviour, and it's only been one year and a half that it's been in this faulty state.

I've had to stay at 1.25.6 in my system because I rely on the previous correct logic for negating expressions. I just don't see how the comment about not breaking existing systems makes any sense.

PR #7375, opened 2025-02-19
Support multiple expressions in OrderBy clause

Orbsynated:
- [x] Do only one thing
- [x] Tested

Currently, it's only possible to include one expression in an OrderBy clause. And there is no way to include both OrderBy expression and column name.
This PR changes the OrderBy clause to allow the usage of multiple expressions along with multiple column names.

### User Case Description

I want to order by an expression and multiple column names.


jinzhu:
Sorry we don't want to introduce break changes for this.

With `Columns{Raw:true}` you want add multi expressions to order by.

PR #7379, opened 2025-03-04
fix: return init dialector error

codingplz:
- [ ] Do only one thing
- [ ] Tested


a631807682 (commented):
Can you add unit tests that describe what this PR is fixing?

PR #7414, opened 2025-04-10
Log levels change

hertzcodes:
- [ ] Do only one thing
- [ ] Tested

Almost every logger's log levels is descending so i decided to change Gorm's levels too since I had problems configuring slog with it.


logger/logger.go
```diff
@@ -36,13 +36,13 @@ type LogLevel int
 
 const (
 	// Silent silent log level
```
jinzhu, line 38: Why did you change it to a fixed number instead of just reordering them?

jinzhu:
Closing this PR as it introduces breaking changes to third-party loggers.

PR #7418, opened 2025-04-13
fix: prevent interface type array from causing runtime errors

KEHyeon:
- [ ] Do only one thing
- [v] Tested

fix: prevent interface type array from causing runtime errors
It has been modified to find out the value when it is an interface array.
### User Case Description
```go
type UserInterface interface {
	GetName() string
}

db.Table("users").Create([]UserInterface{UserToUserInterface(&userData)})
```
In this case, create is allowed.

### DIFF
--- a/tests/create_test.go
+++ b/tests/create_test.go
@@ -791,3 +791,15 @@ func TestCreateFromMapWithTable(t *testing.T) {
 		t.Errorf("failed to create data from map with table, @id != id")
 	}
 }
+
+func TestCreateWithInterfaceArrayType(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tests/create_test.go
```diff
@@ -791,3 +791,15 @@ func TestCreateFromMapWithTable(t *testing.T) {
 		t.Errorf("failed to create data from map with table, @id != id")
 	}
 }
+
+func TestCreateWithInterfaceArrayType(t *testing.T) {
```
jinzhu, line 795: Does it work for queries with an interface array as result?

PR #7420, opened 2025-04-14
fixed memory leak in func (stmt *Statement) ParseWithSpecialTableName

ManyakRus:
- [ ] Do only one thing
- [ ] Tested

*fixed memory leak in func (stmt *Statement) ParseWithSpecialTableName*

need fill nil value after using, for prevent memory leak
		tables = nil

### User Case Description

Gorm function WithContext() have big memory leak.
In golang version 1.2.21 my request solve all problems with memory leak.
in golang version higher gorm have another new one memory leak.

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -495,10 +495,10 @@ func (stmt *Statement) ParseWithSpecialTableName(value interface{}, specialTable
 		if len(tables) == 2 {
 			stmt.TableExpr = &clause.Expr{SQL: stmt.Quote(stmt.Schema.Table)}
 			stmt.Table = tables[1]
-			tables = nil
+			//tables = nil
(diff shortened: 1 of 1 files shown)


last-warrior:
how is this fixing memory leak ? You have commented out `tables=nil` . Now it's just expanded if condition.

ManyakRus:
my fix not fix memeory leak i think :-( 
i am sorry
it worked at first, but not now.

however Gorm function WithContext() have big memory leak.
I deleted WithContext() everywhere and problem fixed for me, but it is not good solve

jinzhu:
WithContext doesn't leak memory, maybe the ctx you created leaks

PR #7424, opened 2025-04-17
(WIP) Implement Generics API

jinzhu:
feat(gorm): add generic CRUD & query interfaces

This commit introduces a type-safe, generic API layer on top of GORM that
allows you to perform Create, Read, Update, Delete and raw SQL operations

Helper constructor:

```go
  func G[T any](db *DB, opts ...clause.Expression) Interface[T]
```

Basic usage:

```go
// Create a user
u := User{Name: "Alice"}
if err := gorm.G[User](DB).Create(ctx, &u); err != nil {
  // handle error
}

// Query by name and count
cnt, err := gorm.G[User](DB).
  Where("name = ?", "Alice").
  Count(ctx, "*")

// Raw + Find
users, err := gorm.G[User](DB).
  Raw("SELECT * FROM users WHERE created_at > ?", since).
  Find(ctx)

// Update one field
rows, err := gorm.G[User](DB).
  Where("id = ?", u.ID).
  Update(ctx, "email", "alice@example.com")

// Delete
deleted, err := gorm.G[User](DB).
  Where("id = ?", u.ID).
  Delete(ctx)
```

### DIFF
--- a/generics.go
+++ b/generics.go
@@ -1,0 +1,254 @@
+package gorm
+
+import (
+	"context"
+	"database/sql"
+
+	"gorm.io/gorm/clause"
+)
+
+type Interface[T any] interface {
+	Raw(sql string, values ...interface{}) ExecInterface[T]
+	Exec(ctx context.Context, sql string, values ...interface{}) error
+	CreateInterface[T]
+}
+
+type CreateInterface[T any] interface {
+	ChainInterface[T]
+	Table(name string, args ...interface{}) CreateInterface[T]
+	Create(ctx context.Context, r *T) error
+	CreateInBatches(ctx context.Context, r *[]T, batchSize int) error
+}
+
+type ChainInterface[T any] interface {
+	ExecInterface[T]
+	Scopes(scopes ...func(db *Statement)) ChainInterface[T]
+	Where(query interface{}, args ...interface{}) ChainInterface[T]
+	Not(query interface{}, args ...interface{}) ChainInterface[T]
+	Or(query interface{}, args ...interface{}) ChainInterface[T]
+	Limit(offset int) ChainInterface[T]
+	Offset(offset int) ChainInterface[T]
+	Joins(query string, args ...interface{}) ChainInterface[T]
+	InnerJoins(query string, args ...interface{}) ChainInterface[T]
+	Select(query string, args ...interface{}) ChainInterface[T]
+	Omit(columns ...string) ChainInterface[T]
+	MapColumns(m map[string]string) ChainInterface[T]
+	Distinct(args ...interface{}) ChainInterface[T]
+	Group(name string) ChainInterface[T]
+	Having(query interface{}, args ...interface{}) ChainInterface[T]
+	Order(value interface{}) ChainInterface[T]
+	Preload(query string, args ...interface{}) ChainInterface[T]
+
+	Delete(ctx context.Context) (rowsAffected int, err error)
+	Update(ctx context.Context, name string, value any) (rowsAffected int, err error)
+	Updates(ctx context.Context, t T) (rowsAffected int, err error)
+	Count(ctx context.Context, column string) (result int64, err error)
+}
+
+type ExecInterface[T any] interface {
+	Scan(ctx context.Context, r interface{}) error
+	First(context.Context) (T, error)
+	Last(ctx context.Context) (T, error)
+	Find(ctx context.Context) ([]T, error)
+	FindInBatches(ctx context.Context, batchSize int, fc func(data []T, batch int) error) error
+	Row(ctx context.Context) *sql.Row
+	Rows(ctx context.Context) (*sql.Rows, error)
+}
+
+func G[T any](db *DB, opts ...clause.Expression) Interface[T] {
+	v := &g[T]{
+		db:   db.Session(&Session{NewDB: true}).Clauses(opts...),
+		opts: opts,
+	}
+
+	v.createG = &createG[T]{
+		chainG: chainG[T]{
+			execG: execG[T]{g: v},
+		},
+	}
+	return v
+}
+
+type g[T any] struct {
+	*createG[T]
+	db   *DB
+	opts []clause.Expression
+}
+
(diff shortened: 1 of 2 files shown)


callbacks/query.go
```diff
@@ -146,9 +146,13 @@ func BuildQuerySQL(db *gorm.DB) {
 
 					if isRelations {
 						genJoinClause := func(joinType clause.JoinType, parentTableName string, relation *schema.Relationship) clause.Join {
-							tableAliasName := relation.Name
-							if parentTableName != clause.CurrentTable {
-								tableAliasName = utils.NestedRelationName(parentTableName, tableAliasName)
+							tableAliasName := join.Alias
```
a631807682, line 149: Nested joins will use the same alias. Can we use a new API for generics? Maybe something like
```diff
type JoinTarget struct {
	Type JoinType
	Association string
	Subquery Expression
	Table string
+ 	NestedJoin JoinTarget
}
```

a631807682 (commented):
It may be necessary to add unit tests for nested join/preload.
For generics API, how should we reuse it? (Similar to the original Session)

callbacks/query.go
jinzhu: Fixed the same alias issue for nested Joins.

jinzhu:
> It may be necessary to add unit tests for nested join/preload. For generics API, how should we reuse it? (Similar to the original Session)

There are two ways to reuse db:

1.  Just share the db var with condition, as shown in this test https://github.com/go-gorm/gorm/blob/generics/tests/generics_test.go#L758-L771 , which no longer causes any DB pollution issues.

2. Using `Scopes` to reuse code — though I'm still thinking about whether there's a better way to implement this.

Do you have any suggestions?

a631807682:
> There are two ways to reuse db:
> 
> 1. Just share the db var with condition, as shown in this test [`generics`/tests/generics_test.go#L758-L771](https://github.com/go-gorm/gorm/blob/generics/tests/generics_test.go#L758-L771) , which no longer causes any DB pollution issues.
> 2. Using `Scopes` to reuse code — though I'm still thinking about whether there's a better way to implement this.
> 
> Do you have any suggestions?

II think the generics API is simple enough without polluting the db.

a631807682 (approved):
LGTM

NeariX67:
I am currently adopting to the generics branch and noticed that there is no .ToSQL() method. Would be nice to see that added as well, to basicly stay method compatible with the old api.

jinzhu:
> I am currently adopting to the generics branch and noticed that there is no .ToSQL() method. Would be nice to see that added as well, to basicly stay method compatible with the old api. Edit: .Save() and .FirstOrCreate() are also missing

@NeariX67  you can write it like this https://github.com/go-gorm/gorm/blob/generics/tests/generics_test.go#L842-L847

NeariX67:
@jinzhu Are you going to implement `FirstOrCreate`, `Pluck`, `Create` (for multiple objects, without batches) and `Save` for this generics API?

jinzhu:
> @jinzhu Are you going to implement `FirstOrCreate`, `Pluck`, `Create` (for multiple objects, without batches) and `Save` for this generics API?

We are not planning to implement those methods. `FirstOrCreate` and `Save` are convenience APIs, but they can be misleading — for example, they don’t handle concurrency issues well. It's often better to write more explicit and meaningful code instead.

Also, for `Pluck`, you can use `Select` with `Scan`; and for batch creation, you can use `CreateInBatches` based on the length of your data to control the batch size manually.

With the generics API, our goal is to take a more restrained approach to API design to help avoid potential misuse.

jinzhu:
> @jinzhu A small bug that prevents the correct execution of `Updates`;
> 
> `Updates(ctx context.Context, t T) (rowsAffected int, err error)`
> 
> should be a pointer like so;
> 
> `Updates(ctx context.Context, t *T) (rowsAffected int, err error)`
> 
> This doesn't show up as a mistake because the internal `Updates` receives an `interface{}`

What issue might occur if a pointer is not passed in?

jinzhu:
> Transaction would logically also be a feature for Generics? So that the `tx` is referring to the generic Interface object instead of the whole db object. Or am I looking at this incorrectly?

Transactions are used like this: https://github.com/go-gorm/gorm/blob/master/tests/generics_test.go#L813-L819

or 

```go
db.Transaction(func(tx *DB) error {
    users, err gorm.G[User](tx).Find(ctx)
    return err
})
```

Allendar:
> > @jinzhu A small bug that prevents the correct execution of `Updates`;
> > `Updates(ctx context.Context, t T) (rowsAffected int, err error)`
> > should be a pointer like so;
> > `Updates(ctx context.Context, t *T) (rowsAffected int, err error)`
> > This doesn't show up as a mistake because the internal `Updates` receives an `interface{}`
> 
> What issue might occur if a pointer is not passed in?

The error that the library throws is explicitly that it expects to receive `a pointer to a struct` in the `Updates` method internals. For example my User Store's Upsert method;

```
func (s *Store) Upsert(data *model.User, userID uint) error {
	ctx := context.WithValue(context.Background(), model.UserIDKey, userID)

	if data.ID == 0 {
		return errors.Trace(s.creatorDB.Create(ctx, data))
	}

	rowsAffected, err := s.updaterDB.Updates(ctx, data)
	if err != nil {
		return errors.Trace(err)
	}

	if rowsAffected == 0 {
		return errors.Trace(ErrUpsertDidntModifyData)
	}

	return nil
}
```

If I call `s.updaterDB.Updates(ctx, data)` as `s.updaterDB.Updates(ctx, *data)` then it will throw that error.

jinzhu:
> Also the `Association` is not available on purpose in Generics? Like so in the normal style;
> 
> ```go
> 	if err := s.osDeleterDB.WithContext(ctx).Model(&data).Association("GameGameOSs").
> 		Delete(operatingSystem); err != nil {
> 		return errors.Trace(err)
> 	}
> ```

Yes, it’s not included yet—I’m still considering if there’s a better way to handle associations.

PR #7429, opened 2025-04-17
tests: Added tests for associations

gdhameeja:
Added tests for relationship, parsing, replace, delete

- [ x  ] Do only one thing
- [ x ] Non breaking API changes
- [ x ] Tested

Adds tests for associations. Bumped test coverage for the file `associations.go` from 0 percent to 48%.

### User Case Description

I added these test cases as a way to learn more about gorm and its codebase. I would love to contribute more tests going forward if they're welcome.

### DIFF
--- a/association_test.go
+++ b/association_test.go
@@ -1,0 +1,471 @@
+package gorm
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+	"sync"
+	"testing"
+
+	"gorm.io/gorm/clause"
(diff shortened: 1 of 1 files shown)


jinzhu:
You can find some existing association-related tests in the `tests` directory of the GORM repository — the files typically start with `associations_`, like:

👉 https://github.com/go-gorm/gorm/tree/master/tests

It might be a good idea to add your tests there to keep things organized and consistent with the current structure.

PR #7435, opened 2025-04-24
Preparestmt use LRU Map instead default map

Yaccc:
PrepareStmt support use lrumap instead default Map
PreparedStmtDB stmt change to interface to defined maps
```golang 
type StmtStore interface {
	Get(key string) (*Stmt, bool)
	Set(key string, value *Stmt)
	Delete(key string)
	AllMap() map[string]*Stmt
}
```

```golang 
func NewPreparedStmtDB(connPool ConnPool, prepareStmtLruConfig *PrepareStmtLruConfig) *PreparedStmtDB {
	return &PreparedStmtDB{
		ConnPool: connPool,
		Stmts: func() StmtStore {
			var stmts StmtStore
			if prepareStmtLruConfig != nil && prepareStmtLruConfig.Open {
				lru := &LruStmtStore{}
				lru.NewLru(prepareStmtLruConfig.Size, prepareStmtLruConfig.TTL)
				stmts = lru
			} else {
				stmts = &DefaultStmtStore{}
			}
			return stmts
		}(),
		Mux: &sync.RWMutex{},
	}
}
```
1. original maps use defaultStmtStore,lur maps use LruStmtStore
2. PrepareStmtLruConfig init lru size and expired time
3. size must lt zero ande default ttl is 
```golang 
const noEvictionTTL = time.Hour * 24 * 365 * 10
```
```golang 
const DEFAULT_MAX_SIZE = (1 << 63) - 1
const DEFAULT_TTL = time.Hour * 24
	// PrepareStmt cache support LRU expired
	PrepareStmtMaxSize int
	PrepareStmtTTL     time.Duration
```

### User Case Description

TestPreparedStmtLruFromTransaction check lru active
```golang 
conn, ok := tx.ConnPool.(*gorm.PreparedStmtDB)
	lens := len(conn.Stmts.AllMap())
	if lens == 0 {
		t.Fatalf("lru should not be empty")
	}
	time.Sleep(time.Second * 40)
	AssertEqual(t, ok, true)
	AssertEqual(t, len(conn.Stmts.AllMap()), 0)
```


gorm.go
```diff
@@ -65,6 +69,11 @@ type Config struct {
 	callbacks  *callbacks
 	cacheStore *sync.Map
 }
+type PrepareStmtLruConfig struct {
```
jinzhu, line 72: This is no longer needed?

Yaccc: deleted

gorm.go
```diff
@@ -268,7 +277,7 @@ func (db *DB) Session(config *Session) *DB {
 		if v, ok := db.cacheStore.Load(preparedStmtDBKey); ok {
 			preparedStmt = v.(*PreparedStmtDB)
 		} else {
-			preparedStmt = NewPreparedStmtDB(db.ConnPool)
+			preparedStmt = NewPreparedStmtDB(db.ConnPool, db.Config.PrepareStmtMaxSize, db.Config.PrepareStmtTTL)
```
jinzhu, line 275: Do we need to allow session-level configuration? For example, configure it here: https://github.com/go-gorm/gorm/blob/master/gorm.go#L107

prepare_stmt.go
```diff
@@ -17,16 +20,52 @@ type Stmt struct {
...
 
 type PreparedStmtDB struct {
-	Stmts map[string]*Stmt
+	Stmts StmtStore
 	Mux   *sync.RWMutex
 	ConnPool
 }
 
-func NewPreparedStmtDB(connPool ConnPool) *PreparedStmtDB {
+const DEFAULT_MAX_SIZE = (1 << 63) - 1
```
jinzhu, line 28: Could we relocate the majority of the following code to an internal package, such as internal/stmt_store, to prevent the exposure of additional structs or methods?

Yaccc: cause circular dependency

gorm.go
Yaccc:
```
if config.PrepareStmt {
		preparedStmt := NewPreparedStmtDB(db.ConnPool, config.PrepareStmtMaxSize, config.PrepareStmtTTL)
		db.cacheStore.Store(preparedStmtDBKey, preparedStmt)
		db.ConnPool = preparedStmt
	}

```
session default use static config

iTanken:
Build failed in 32bit environment

```shell
GOARCH=386 go test -timeout 20m -v ./...
  shell: /usr/bin/bash -e {0}
# gorm.io/gorm/internal/stmt_store
Error: ../../../go/pkg/mod/gorm.io/gorm@v1.26.0/internal/stmt_store/stmt_store.go:99:10: cannot use defaultMaxSize (untyped int constant 9223372036854775807) as int value in assignment (overflows)
```

andig:
We've just hit this as well. A .1 release would be highly appreciated.

Issue #7411, opened 2025-04-09
Add a `Only` API to queries


AlonMenczerLema:
## Describe the feature

Add a `Only` Api to the current `First`, `Last` and `Take` - which will fetch a single record and asserts that it was the only record that matched the query

## Motivation

Enables validation in a nicer method - is available in many ORMs

Examples:
EntGO - https://github.com/ent/ent/blob/b1786895ce976a38722c6aeb1e3ba679d23786a9/ent.go#L505 has an `Only` operation which will return a `ent.IsNotSingular` if there is more than one result
SQLAlchemy - https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.one

aditya14as:
If this issue is still open, can I take this?

PR #7444, opened 2025-05-05
feat: add Only

mo1ein:
- [x] Do only one thing
- [x] Tested

Introduces a new `Only` query method (see issue #7411). This method retrieves exactly one record ordered by primary key and returns an error if zero or multiple records match.

### User Case Description
GORM previously lacked a built‑in `Only` operation, forcing users to manually combine `First` and `Count` to enforce uniqueness. Many other ORMs provide this feature natively.


jinzhu:
thank you for your PR! However, we’d prefer not to add too many “sugar” APIs.

And a better implementation might be using LIMIT 2 to check if there’s more than one record, rather than triggering an additional COUNT query.

thank you

mo1ein:
So shouldn't the issue be closed too?

PR #7450, opened 2025-05-12
fix decimal migrate error.

Chise1:
- [x] Do only one thing
- [x] Tested

修复automigrate无法正确识别decimal精度是否变化的问题。

### User Case Description

[参考issues 7449](https://github.com/go-gorm/gorm/issues/7449)

如果有一个对象如下：
```
type Change struct {
	RecID int64  `gorm:"column:recid;type:decimal(9,0);not null" json:"recid"`
}

```

每次自动同步都会触发
```
 ALTER TABLE `changes` MODIFY COLUMN `recid` decimal(9,0) NOT NULL    
```

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -474,7 +474,6 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 	// found, smart migrate
 	fullDataType := strings.TrimSpace(strings.ToLower(m.DB.Migrator().FullDataTypeOf(field).SQL))
 	realDataType := strings.ToLower(columnType.DatabaseTypeName())
-
 	var (
 		alterColumn bool
 		isSameType  = fullDataType == realDataType
@@ -513,8 +512,19 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 				}
 			}
 		}
+	}
 
-		// check precision
+	// check precision
+	if realDataType == "decimal" || realDataType == "numeric" &&
+		regexp.MustCompile(realDataType+`\(.*\)`).FindString(fullDataType) != "" { // if realDataType has no precision,ignore
+		precision, scale, ok := columnType.DecimalSize()
+		if ok {
+			if !strings.HasPrefix(fullDataType, fmt.Sprintf("%s(%d,%d)", realDataType, precision, scale)) &&
+				!strings.HasPrefix(fullDataType, fmt.Sprintf("%s(%d)", realDataType, precision)) {
+				alterColumn = true
+			}
+		}
+	} else {
 		if precision, _, ok := columnType.DecimalSize(); ok && int64(field.Precision) != precision {
 			if regexp.MustCompile(fmt.Sprintf("[^0-9]%d[^0-9]", field.Precision)).MatchString(m.DataTypeOf(field)) {
 				alterColumn = true
--- a/tests/migrate_test.go
+++ b/tests/migrate_test.go
@@ -1987,3 +1987,114 @@ func TestMigrateWithUniqueIndexAndUnique(t *testing.T) {
 		}
 	}
 }
+
+func testAutoMigrateDecimal(t *testing.T, model1, model2 any) []string {
+	tracer := Tracer{
+		Logger: DB.Config.Logger,
+		Test: func(ctx context.Context, begin time.Time, fc func() (sql string, rowsAffected int64), err error) {
+			sql, _ := fc()
+			if strings.HasPrefix(sql, "ALTER TABLE ") {
+				t.Fatalf("shouldn't execute ALTER COLUMN TYPE if decimal is not change: sql: %s", sql)
+			}
+		},
+	}
+	session := DB.Session(&gorm.Session{Logger: tracer})
+
+	DB.Migrator().DropTable(model1)
+	var modifySql []string
+	if err := session.AutoMigrate(model1); err != nil {
+		t.Fatalf("failed to auto migrate, got error: %v", err)
+	}
+	if err := session.AutoMigrate(model1); err != nil {
+		t.Fatalf("failed to auto migrate, got error: %v", err)
+	}
+	tracer2 := Tracer{
+		Logger: DB.Config.Logger,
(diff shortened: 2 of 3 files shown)


Chise1:
我在代码里面没有看到automigrate相关的测试内容。
如果需要补充测试代码，请留言。

a631807682 (commented):
Please add some tests in https://github.com/go-gorm/gorm/blob/master/tests/migrate_test.go

migrator/migrator.go
```diff
@@ -515,13 +515,22 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 		}
 
 		// check precision
-		if precision, _, ok := columnType.DecimalSize(); ok && int64(field.Precision) != precision {
-			if regexp.MustCompile(fmt.Sprintf("[^0-9]%d[^0-9]", field.Precision)).MatchString(m.DataTypeOf(field)) {
-				alterColumn = true
+		if realDataType == "decimal" {
```
a631807682, line 518: The types `numeric` and `decimal` are similar in `postgrel/sqlserver`.

migrator/migrator.go
```diff
@@ -515,13 +515,22 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
...
+			precision, scale, ok := columnType.DecimalSize()
+			if ok {
+				if !strings.HasPrefix(fullDataType, fmt.Sprintf("decimal(%d,%d)", precision, scale)) &&
+					!strings.HasPrefix(fullDataType, fmt.Sprintf("decimal(%d)", precision)) {
```
a631807682, line 522: Should we check for prefix or full string match here?

a631807682, line 522: It seems that only when scale is 0, `decimal(%d)` needs to be checked.

a631807682 (commented):
Can you add unit tests for my question?

migrator/migrator.go
Chise1: fullDataType maybe like: ``` decimal(10,1) not null```

Chise1: mysql fails to correctly assign the value of decimal(10,0) to precision and scale. So the following configuration methods are not supported:
```go
type MigrateDecimalColumn2 struct {
	RecID1 int64 `gorm:"column:recid1;type:decimal;not null;precision:10;scale:1" json:"recid1"`
}
```

migrator/migrator.go
Chise1: add sqlserver decimal support.
postgres only support numeric. When aliases occur, postgres fails to correctly recognize the change in precision.（code in go-gorm/postgres/migrator.go line 336）

a631807682: https://www.postgresql.org/docs/17/datatype-numeric.html

It seems unnecessary to recheck the same types in postgres, this is already checked in https://github.com/go-gorm/gorm/blob/master/migrator/migrator.go#L480, are you interested in investigating and fixing them in another PR?

migrator/migrator.go
a631807682: We need to handle the case where the type `decimal(10,2)` changes to `decimal(10)`, which is actually an abbreviation of `decimal(10,0)`.

Chise1: Ideally, decimal(10,0) needs to be parsed, and values should be assigned to precision and scale, followed by comparison. However, such an operation requires modifying the parsing parts of all drivers. I'm not sure whether too big changes can be accepted.

migrator/migrator.go
Chise1: I can fix this problem in new PR. Is there anything that still needs to be modified in the current PR?

migrator/migrator.go
Chise1: > We need to handle the case where the type changes to , which is actually an abbreviation of .`decimal(10,2)``decimal(10)``decimal(10,0)`

The purpose of this PR is to solve the problem that decimal is modified each time automigrate is used (decimal remains unchanged). For a change caused by the structural difference of decimal, the database driver needs to support decimal (just like supporting VARCHAR).

Chise1: > We need to handle the case where the type `decimal(10,2)` changes to `decimal(10)`, which is actually an abbreviation of `decimal(10,0)`.

When migrating, the aliases of data types should not be checked. No one changes back and forth between `numeric` and `decimal`. I suggest that the check for variable aliases should be removed.

Chise1: > > > We need to handle the case where the type `decimal(10,2)` changes to `decimal(10)`, which is actually an abbreviation of `decimal(10,0)`.
> > 
> > 
> > When migrating, the aliases of data types should not be checked. No one changes back and forth between `numeric` and `decimal`. I suggest that the check for variable aliases should be removed.
> 
> It makes sense to check the alias. Although we declare to use type A, the actual type in the database may be an alias of A. Of course, I don't think it's necessary to check again in postgres driver. I think it's only necessary to check the precision when the types are the same (including aliases). If different types are used, an update will definitely be triggered. If you think the changes are too big we can create multiple PRs, or if you think such use cases are rare we can also support it later.

I have been thinking about a new solution to aliases recently. Strengthen the functionality of the DataTypeOf function. For example, when the input parameter is numeric or decimal, decimal is returned as the result. This can also solve the problem that decimal(10) and decimal(10,0) are regarded as two types.

Chise1: I hope to merge this pr first. For other issues, we can communicate via email or qq.

tests/migrate_test.go
```diff
@@ -1987,3 +1987,112 @@ func TestMigrateWithUniqueIndexAndUnique(t *testing.T) {
...
+		},
+	}
+	session2 := DB.Session(&gorm.Session{Logger: tracer2})
+	err := session2.Table("migrate_decimal_columns").Migrator().AutoMigrate(model2)
+	if err != nil {
+		t.Fatalf("failed to get column types, got error: %v", err)
+	}
+	return modifySql
+}
+func decimalColumnsTest[T, T2 any](t *testing.T, expectedSql []string) {
```
demoManito, line 2025: Add a line here?

demoManito (approved):

PR #7453, opened 2025-05-12
Composite pk handling

keif888:
- [X] Do only one thing
- [X] Tested

Ensure that Composite Primary Keys are handled correctly.
Fix issue https://github.com/go-gorm/playground/issues/800
Add new test for FindInBatch with Composite Primary key

### User Case Description

When using a struct with a composite primary key, all fields in the primary key should be used in the order by's to ensure that all databases will guarantee returning the same record(s) in the same order every time.
This is applicable to First, Last, FirstOrCreate, FirstOrInit and FindInBatch.

When getting subsequent batches in FindInBatch, the where clause has to take into account all parts of the composite primary key.

#### First, FirstOrCreate, FirstOrInit Use Case

1. create struct with multiple columns tagged with primaryKey
2. load struct with many rows of data, with the 1st column in the primary key repeated many times
3. issue a First/FirstOrCreate/FirstOrInit command

**Expected:**
The same record is always returned.
**Actual:**
At a random, usually under heavy load, the dbms may return a different 1st record, although it has the same value for the 1st column in the primary key.
**This Pull:**
Guaranteed to return the same record every time as all columns in the primary key are included in the order by statement.

#### Last Use Case

1. create struct with multiple columns tagged with primaryKey
2. load struct with many rows of data, with the 1st column in the primary key repeated many times
3. issue a Last command

**Expected:**
The same record is always returned.
**Actual:**
At a random, usually under heavy load, the dbms may return a different record, although it has the same 1st column in the primary key.
The record returned may not be the last record when taking all columns in the composite primary key into account, as the order by does not specify all the columns.
**This Pull:**
Guaranteed to return the same record every time as all columns in the primary key are included in the order by statement.

#### FindInBatch Use Case

1. create struct with multiple columns tagged with primaryKey
2. load struct with many rows of data, with the 1st column in the primary key repeated many times
3. issue a FindInBatch command with a batch smaller than the number of records
4. process the batches until all records consumed

**Expected:**
All the records are returned in batches
**Actual:**
A no primary key error is returned
**This Pull:**
1st issuance for FindInBatch has all columns in primary key in the order by.
Subsequent batches have where clause to ensure that the already consumed batches are excluded, and remaining records are returned up to the batch limit.

### DIFF
--- a/clause/clause.go
+++ b/clause/clause.go
@@ -64,6 +64,7 @@ func (c Clause) Build(builder Builder) {
 
 const (
 	PrimaryKey   string = "~~~py~~~" // primary key
+	PrimaryKeys  string = "~~~ps~~~" // primary keys
 	CurrentTable string = "~~~ct~~~" // current table
 	Associations string = "~~~as~~~" // associations
 )
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -228,7 +228,7 @@ func (db *DB) FindInBatches(dest interface{}, batchSize int, fc func(tx *DB, bat
 
 		// Optimize for-break
 		resultsValue := reflect.Indirect(reflect.ValueOf(dest))
-		if result.Statement.Schema.PrioritizedPrimaryField == nil {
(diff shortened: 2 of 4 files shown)


keif888:
FYI:
The test ExampleLRU appears to be unstable.
It is not utilising any of the code that this Pull changes.

PR #7457, opened 2025-05-15
feat: Add optional case-insensitive matching for schema field lookup

joboon:
- [x] Do only one thing
- [x] Tested

Adds a new configuration option to ignore case sensitivity when associating table and column names with the target schema. Requires opting in. The default behavior is unchanged.

### User Case Description

As a developer supporting an application which supports multiple databases, all with different default casing and case sensitivity implementations, I want to use structs with consistent naming from a Go perspective while keeping table and column names consistent with the casing for the target database. This avoids the need to quote all future interactions with the database to match a specific casing. In order to preserve the current behavior, matching schema fields without case sensitivity should be optional.

### DIFF
--- a/finisher_api.go
+++ b/finisher_api.go
@@ -270,7 +270,7 @@ func (db *DB) assignInterfacesToValue(values ...interface{}) {
 				db.assignInterfacesToValue(exprs)
 			}
 		default:
-			if s, err := schema.Parse(value, db.cacheStore, db.NamingStrategy); err == nil {
+			if s, err := schema.ParseWithCaseInsensitivity(value, db.cacheStore, db.NamingStrategy, db.CaseInsensitiveSchemaFields); err == nil {
 				reflectValue := reflect.Indirect(reflect.ValueOf(value))
 				switch reflectValue.Kind() {
 				case reflect.Struct:
--- a/go.mod
+++ b/go.mod
@@ -5,5 +5,12 @@ go 1.18
 require (
 	github.com/jinzhu/inflection v1.0.0
 	github.com/jinzhu/now v1.1.5
+	github.com/stretchr/testify v1.10.0
 	golang.org/x/text v0.20.0
+)
+
+require (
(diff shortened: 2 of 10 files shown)


jinzhu:
Hi @joboon 

How about setting up a different NamingStrategy for each database?

joboon:
> Hi @joboon
> 
> How about setting up a different NamingStrategy for each database?

Hi, @jinzhu!

Thanks for taking a look and reaching out. Admittedly, I'm not super familiar with all the customizability within gorm. However, from what I can tell, the NamingStrategy controls the Go -> SQL conversions. I'm less worried about that, since the database (Oracle, for example) will automatically adjust the casing for me when interpreting the SQL. However, in the SQL -> Go conversion in gorm, Oracle will return results with uppercased table/column names and they won't be mapped to the schema fields based on a case-agnostic struct tag. This PR is to add optional case insensitivity when attempting to match those names with the names defined by the structs (without having to tie us to a specific DB's casing implementation for all queries). 

Am I missing something in your recommendation? Have I misunderstood how NamingStrategy objects work? Thanks again for your time!

joboon:
Hi, @jinzhu. Any more ideas for me here?

KeenBash:
I also need this feature. We have many fields with irregular naming conventions, and we hope that when defining the structure, we only need to define lowercase field names to match the database fields.

KeenBash:
> I also need this feature. We have many fields with irregular naming conventions, and we hope that when defining the structure, we only need to define lowercase field names to match the database fields.

There may be tables with the same name and structure in PG and MySQL databases, but the field names may be all uppercase in PG and all lowercase in MySQL. We hope to use the same structure to mapping the fields

joboon:
@jinzhu, any more thoughts on this one? Thanks!

joboon:
Bumping this PR. Thanks.

jinzhu:
Closing this PR in favor of [#7619](https://github.com/go-gorm/gorm/pull/7619)

PR #7458, opened 2025-05-17
Create modelp.go

snowdream:
- [x] Do only one thing
- [x] Tested

Create modelp.go


a631807682 (commented):
Why do we need it? https://github.com/go-gorm/gorm/blob/master/model.go

PR #7462, opened 2025-05-21
Feature: add RightJoin Function

NeariX67:
- [X] Do only one thing
- [X] Tested

Add RightJoin Function

### User Case Description

Get correct results from right join, that inner join and left join wont deliver


jinzhu:
We will recommend users to use the generics-based approach in the future: https://github.com/go-gorm/gorm/pull/7424

Therefore, this method won’t be necessary anymore — the same logic can be written like this instead:

image

Thank you for your PR.

PR #7468, opened 2025-05-29
fix(schema): Check the callback function parameter type

demoManito:
- [ ] Do only one thing
- [ ] Tested

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -313,8 +313,12 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 	for _, cbName := range callbackTypes {
 		if methodValue := callBackToMethodValue(modelValue, cbName); methodValue.IsValid() {
 			switch methodValue.Type().String() {
-			case "func(*gorm.DB) error": // TODO hack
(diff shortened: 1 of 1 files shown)


schema/schema.go
```diff
@@ -313,8 +313,12 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 	for _, cbName := range callbackTypes {
 		if methodValue := callBackToMethodValue(modelValue, cbName); methodValue.IsValid() {
 			switch methodValue.Type().String() {
-			case "func(*gorm.DB) error": // TODO hack
-				reflect.Indirect(reflect.ValueOf(schema)).FieldByName(string(cbName)).SetBool(true)
+			case "func(*gorm.DB) error":
```
jinzhu, line 316: It should check whether the package matches the currently used one, rather than matching against gorm.io/gorm, to avoid issues with forked codebase

demoManito: My idea is to judge directly according to the type `func(*gorm.DB) error`, and it seems that they will be referenced in a circular manner. Do you have a better solution for implementation?

However, the callback function will also be checked interface during actual callbacks. Do you think the judgment here is still necessary to modify?

demoManito: Modified, thanks, it's really a good solution

Issue #6343, opened 2023-05-22
Is it possible to turn off getting the last inserted id?


ppanphper:
## Your Question

If returning id is turned off, and the model contains an ID field, lastInsertId will be read by default;
eg: driver postgres 
WithoutReturning: true

gorm.Model contain ID field

## Expected answer

Is it possible to turn off this feature?
![image](https://github.com/go-gorm/gorm/assets/6057923/60e48945-8e1f-48fe-b7ec-cb79d581e0e3)

a631807682:
We no longer support this feature, refer to https://github.com/go-gorm/postgres/issues/177#issuecomment-1522815541
Does anyone else need to support this feature? I hope to get more feedback.
To support this feat, a configuration can be added to disable LastInsertId for drivers.

jinzhu:
Maybe we can use field permission to disable the `LastInsertId` feature?

a631807682:
What are the scenarios? Can you describe it in detail? @ppanphper 
My understanding is 
1. Some databases do not support either `Returning` syntax or `LastInsertID` queries, in which case we need to disable them at the driver level
2. `LastInsertID` causes other problem in some cases, so we need to disable it, in which case we can provide field permission disable or driver level disable. 
	https://github.com/go-gorm/gorm/issues/6047

ppanphper:
@a631807682 

eg: distributed pgsql database dws;

For distributed database dws, if you want to return id, sql cannot be pushed down to DN, which affects performance.

If possible, it is best to support control at the driver and field levels; in fact, it is also possible to restrict the characteristics of the default primary key of id.

yyystation:
I am facing the same problem, I try to operate aws redshift, which can't support returning id, when I disable it , it return an error: "LastInsertId is not supported by this driver".

e1esm:
May I take this one?

a631807682:
> May I take this one?

Assigned to you, thanks.

Ajaysinh1290:
is there any update on this issue @e1esm ?
 I'm also facing same problem for redshift. 

> I try to operate aws redshift, which can't support returning id, when I disable it , it return an error: "LastInsertId is not supported by this driver".

krisdiano:
If this issue is not fixed, I would do it in 7 days. @e1esm @a631807682

jinzhu:
Fixed by https://github.com/go-gorm/gorm/commit/c63374f5d154709f99de902cb4d90a1aa8279103

[...]

PR #7469, opened 2025-05-29
feat: Ignore LastInsertID failed

krisdiano:
LastInsertID not yet supported by some database drivers. Users could avoid errors through configuration.

Close https://github.com/go-gorm/gorm/issues/6343

### DIFF
--- a/callbacks/create.go
+++ b/callbacks/create.go
@@ -2,9 +2,7 @@ package callbacks
 
 import (
 	"fmt"
-	"os"
 	"reflect"
-	"strconv"
 	"strings"
 
 	"gorm.io/gorm"
@@ -38,19 +36,10 @@ func BeforeCreate(db *gorm.DB) {
 // Create create hook
 func Create(config *Config) func(db *gorm.DB) {
 	supportReturning := utils.Contains(config.CreateClauses, "RETURNING")
-	rawSupportReturning := supportReturning
 
 	return func(db *gorm.DB) {
 		if db.Error != nil {
 			return
-		}
-
-		mock := os.Getenv("GORM_E2E_TEST_MOCK_CREATE_RETURNING")
-		mockSupportReturning, err := strconv.ParseBool(mock)
-		if err == nil {
-			supportReturning = mockSupportReturning
-		} else {
-			supportReturning = rawSupportReturning
 		}
 
 		if db.Statement.Schema != nil {
--- a/tests/create_test.go
+++ b/tests/create_test.go
@@ -6,7 +6,6 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
-	"os"
 	"regexp"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


callbacks/create.go
```diff
@@ -126,6 +127,22 @@ func Create(config *Config) func(db *gorm.DB) {
 		insertOk := err == nil && insertID > 0
 
 		if !insertOk {
+			if db.IgnoreLastInsertIDWhenNotSupport {
```
a631807682, line 130: I think this feat is to allow users to skip updating the model's primary key. We don't need to pay attention to the internal logic of the driver. Whether to skip `LastInsertID` can be config by the user or the driver itself.

gorm.go
```diff
@@ -24,6 +24,12 @@ type Config struct {
 	SkipDefaultTransaction    bool
 	DefaultTransactionTimeout time.Duration
 
+	// Not all database support LastInsertId, you can set `IgnoreLastInsertIDWhenNotSupport` to true in those cases
+	IgnoreLastInsertIDWhenNotSupport bool
```
a631807682, line 28: I think users can config it themselves, even if the driver supports it, it is not necessary to return `LastInsertID`

a631807682 (commented):
Can you add some tests?

tests/create_test.go
```diff
@@ -535,6 +538,57 @@ func TestCreateNilPointer(t *testing.T) {
...
+}
+
+func TestCreateWithDisableLastInsertID(t *testing.T) {
+	if isSQLServer() {
+		t.Skip("SQLServer driver doesn't use default create hook in gorm")
+	}
+
+	user := GetUser("TestCreateWithDisableLastInsertID0", Config{})
+	err := DB.Create(user).Error
+	if DB.RowsAffected > 0 && err == nil {
```
a631807682, line 562: If you need to check RowsAffected, you should change it to
```go
rs := DB.Create(user)
rs.RowsAffected // 1
rs.Error // nil
```

tests/create_test.go
```diff
@@ -535,6 +538,57 @@ func TestCreateNilPointer(t *testing.T) {
...
+		rawPool := db.ConnPool
+		db.ConnPool = ConnPoolLastInsertIDMock{rawPool}
+		rawStatementPool := db.Statement.ConnPool
+		db.Statement.ConnPool = ConnPoolLastInsertIDMock{rawStatementPool}
+		return func() {
+			db.ConnPool = rawPool
+			db.Statement.ConnPool = rawStatementPool
+		}
+	}
+	defer mockConnPoolExec()()
```
a631807682, line 583: Why do we need a mock conn pool? Removing it does not affect the test results

tests/create_test.go
```diff
@@ -535,6 +538,57 @@ func TestCreateNilPointer(t *testing.T) {
...
+
+	user = GetUser("TestCreateWithDisableLastInsertID1", Config{})
+	err = db.Create(user).Error
+	if err != nil {
```
a631807682, line 587: To test the impact of this config, we should check the result returned by user.ID

PR #7492, opened 2025-06-19
fixes #7486

### LINKED ISSUE #7486: Statement.Changed can't work on a new custom struct
## GORM Playground Link

https://github.com/go-gorm/playground/pull/808

## Description

When hook `tx.Statement.Changed` on func `BeforeUpdate` catch `"age"`

create a struct, and method.

```go
func (m *Man) update(data interface{}) error {
	return DB.Set("data", data).Model(m).Where("id = ?", m.Id).Updates(data).Error
}

func (m *Man) BeforeUpdate(tx *gorm.DB) error {
	if !tx.Statement.Changed("age") {
		fmt.Println("no")
		return nil
	}
	fmt.Println("yes")
	return nil
}
```

this struct have `Age`, so `tx.Statement.Changed("age")` should got `yes`, but got `panic: reflect: Field index out of range`
```go
	var change3 = struct {
		Age int
	}{Age: 30}

	// change3 -> panic: reflect: Field index out of range
	err := idx.update(change3)
	if err != nil {
		return
	}
```

at **playground**, just `change1` and `change2` got yes.

Eshan-Jogwar:
- [x] Do only one thing
- [x] Tested

it fixes #7486. a subset of struct of the main model when passed through to update would create an error in the tx.Statement.Changed() function. this PR fixed this problem

below is a example code which demonstrates that condition
image

struct used for the reference
type Man struct {
	Id string
	Age string
	Name string
	Detail string
}

The problem was that if a complete man struct gorm can predict which the column by column index but if a shorter subset of the struct is passed for quick changes then the index of the column lets say Age will not equate to the index in the struct.

### User Case Description

fixed issue of #7486

### DIFF
--- a/schema/field.go
+++ b/schema/field.go
@@ -457,7 +457,7 @@ func (field *Field) setupValuerAndSetter() {
 	switch {
 	case len(field.StructField.Index) == 1 && fieldIndex > 0:
 		field.ValueOf = func(ctx context.Context, value reflect.Value) (interface{}, bool) {
-			fieldValue := reflect.Indirect(value).Field(fieldIndex)
+			fieldValue := reflect.Indirect(value).FieldByName(field.Name)
 			return fieldValue.Interface(), fieldValue.IsZero()
 		}
 	default:
--- a/tests/check_subset_model_change_test.go
+++ b/tests/check_subset_model_change_test.go
@@ -2,42 +2,87 @@ package tests_test
 
 import (
 	"fmt"
+	"strings"
 	"testing"
+
+	"gorm.io/gorm"
 )
 
 type Man struct {
-	ID  int
-	Age int
-	Name string
+	ID     int
+	Age    int
+	Name   string
+	Detail string
+}
+
+// Panic-safe BeforeUpdate hook that checks for Changed("age")
+func (m *Man) BeforeUpdate(tx *gorm.DB) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("panic in BeforeUpdate: %v", r)
+		}
+	}()
+
+	if !tx.Statement.Changed("age") {
+		return nil
+	}
+	return nil
+}
+
+func (m *Man) update(data interface{}) error {
+	return DB.Set("data", data).Model(m).Where("id = ?", m.ID).Updates(data).Error
 }
 
 func TestBeforeUpdateStatementChanged(t *testing.T) {
+	DB.AutoMigrate(&Man{})
 	type TestCase struct {
 		BaseObjects Man
-		change interface{}
+		change      interface{}
+		expectError bool
 	}
-	fmt.Println("Running Eshan Jogwar Test")
+
 	testCases := []TestCase{
 		{
-			BaseObjects: Man{ID: 12231234, Age: 18, Name: "random-name"},
+			BaseObjects: Man{ID: 1, Age: 18, Name: "random-name"},
 			change: struct {
 				Age int
 			}{Age: 20},
+			expectError: false,
 		},
 		{
-			BaseObjects: Man{ID: 12231234, Age: 18, Name: "random-name"},
+			BaseObjects: Man{ID: 2, Age: 18, Name: "random-name"},
 			change: struct {
-				Age int
 				Name string
-			}{Age: 20, Name: "another-random-name"},
+			}{Name: "name-only"},
+			expectError: true,
+		},
+		{
+			BaseObjects: Man{ID: 2, Age: 18, Name: "random-name"},
+			change: struct {
+				Name string
+				Age int
+			}{Name: "name-only", Age: 20},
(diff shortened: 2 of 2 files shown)


jinzhu:
Can you add some tests?

Eshan-Jogwar:
yes i can add tests. working on it

Eshan-Jogwar:
sir i added the test case file for the bug fix. Thank You for such a great project and a good community.

Eshan-Jogwar:
I want to contribute more to the repository. Is there any list as top priority issues or anything which needs to be done first. Asking because I did not found such information anywhere.

jinzhu:
> I want to contribute more to the repository. Is there any list as top priority issues or anything which needs to be done first. Asking because I did not found such information anywhere.

Thank you for your contribution! If you'd like to proceed, I believe the best approach would be to start from one of the open issues labeled with "type:with reproduction steps":
https://github.com/go-gorm/gorm/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22type%3Awith%20reproduction%20steps%22

liov:
I don't think there should be support for usages that aren't in the documentation, this change could slow down performance to some extent.

liov:
This is a question that arises from this
```go
type Age struct {
	Age int
}

func TestGORM(t *testing.T) {
	man := Man{Id: 1, Name: "Man1", Age: 1}
	DB.Create(&man)

	// update data
	idx := Man{Id: 1}

	change1 := Man{
		Age: 10,
	}
	change2 := map[string]interface{}{
		"age": 10,
	}
	change3 := struct {
		Age
	}{Age{Age: 10}}

	_, _, _ = change1, change2, change3
	var err error

	// change1 -> yes
	err = idx.update(change1)
	if err != nil {
		return
	}

	// change2 -> yes
	err = idx.update(change2)
	if err != nil {
		return
	}

	// change3 -> panic: reflect: Field index out of range
	err = idx.update(change3)
	if err != nil {
		return
	}
}
```
```

2025/06/27 18:10:08 D:/code/playground/main_test.go:17
[2.614ms] [rows:1] INSERT INTO `men` (`name`,`age`,`id`) VALUES ("Man1",1,1) RETURNING `id`
yes

2025/06/27 18:10:09 D:/code/playground/main.go:13 SLOW SQL >= 200ms
[1662.197ms] [rows:1] UPDATE `men` SET `age`=10 WHERE id = 1 AND `id` = 1
no

2025/06/27 18:10:09 D:/code/playground/main.go:13
[0.000ms] [rows:1] UPDATE `men` SET `age`=10 WHERE id = 1 AND `id` = 1
yes

2025/06/27 18:10:11 D:/code/playground/main.go:13 SLOW SQL >= 200ms
[776.255ms] [rows:1] UPDATE `men` SET `age`=10 WHERE id = 1 AND `id` = 1
```
@jinzhu

liov:
After a lot of testing and consideration, I still don't think this PR should be merged.

Firstly, FieldByName performs poorly with large structs, which makes caching fields ineffective and unnecessary.

Secondly, field.ValueOf only addresses one case — the default implementation remains problematic. Additionally, field.ReflectValueOf hasn't been modified at all.
```go
	switch {
	case len(field.StructField.Index) == 1 && fieldIndex >= 0:
		field.ValueOf = func(ctx context.Context, value reflect.Value) (interface{}, bool) {
			fieldValue := reflect.Indirect(value).FieldByName(field.Name)
			return fieldValue.Interface(), fieldValue.IsZero()
		}
	default:
			v = reflect.Indirect(v)
			for _, fieldIdx := range field.StructField.Index {
				if fieldIdx >= 0 {
					v = v.Field(fieldIdx)
...
}
```
And filed. FindByName does not solve problems similar to the following
```go
type Age struct {
	Age int
}
db.Updates(struct {
		Age
	}{Age{Age: 30}})
```

If we need to support other types of structs, I believe similar issues will arise in many other places. 
```go
	changed := func(field *schema.Field) bool {
		fieldValue, _ := field.ValueOf(stmt.Context, modelValue)
		if v, ok := selectColumns[field.DBName]; (ok && v) || (!ok && !restricted) {
			if mv, mok := stmt.Dest.(map[string]interface{}); mok {
				...
			} else {
				destValue := reflect.ValueOf(stmt.Dest)
				for destValue.Kind() == reflect.Ptr {
					destValue = destValue.Elem()
				}
				var changedValue interface{}
				var zero bool
				if destValue.Type() == modelValue.Type() {
					changedValue, zero = field.ValueOf(stmt.Context, destValue)
				} else {
					destField := reflect.ValueOf(destValue).FieldByName(field.Name)
					changedValue, zero = destField.Interface(), destField.IsZero()
				}
				if v {
					return !utils.AssertEqual(changedValue, fieldValue)
				}
				return !zero && !utils.AssertEqual(changedValue, fieldValue)
			}
		}
		return false
	}
```
@jinzhu

PR #7510, opened 2025-07-09
feat: make errors more detailed when setting value to fields

iseki0:
- [x] Do only one thing
- [x] Tested

Provide a detailed error message during `Create`/`Save`/`Update`...

### User Case Description

Currently the error message too simple to diagnostic what's happened.

### DIFF
--- a/callbacks/create.go
+++ b/callbacks/create.go
@@ -183,7 +183,7 @@ func Create(config *Config) func(db *gorm.DB) {
 
 						_, isZero := pkField.ValueOf(db.Statement.Context, rv)
 						if isZero {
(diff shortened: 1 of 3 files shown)


iseki0:
@jinzhu  Maybe you have to change the linter rule or reformat the code(then rebase the PR)?

jinzhu:
Let’s enhance the error message directly within the Set function, rather than introducing an additional wrapper.

Thank you.

iseki0:
@jinzhu Enhance it in the `Set` function is great. But why we don't make a wrapper? If we use a wrapper, we have a chance let the caller knows which field failed, in the future(Consider whether expose a error type is a difficult thing.)

Issue #7513, opened 2025-07-11
Long column identifiers in deeply nested joins cause fields to be omitted


maxwey:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/815

## Context 

In Postgres, query column identifiers are subject to Postgres' 63 character limit. While Postgres will accept queries with identifiers that are longer than this limit, the query response will truncate the identifier to the 63 character limit.

This has an impact on Gorm when it attempts to map a query response back into a struct.

### Failing Use Case -- As Demonstrated in Playground Test Case

I have added a test case with a doubly nested entity. The test case expects the `international_code` field to be populated following the join query, but it is `""`. This is a silent error.

This happens because the generated SQL query to retrieve this column is:
```SQL
`CurrentEmployerInformation__PreferredCompanyLanguage`.`international_code` AS `CurrentEmployerInformation__PreferredCompanyLanguage__international_code`
```

`CurrentEmployerInformation__PreferredCompanyLanguage__international_code` is larger than 63 characters long. 

Postgres accepts this in the query correctly, but returns a truncated identifier for this column that is 63 characters long. Because of the truncation, Gorm cannot map the data to its intended column, leaving it empty.

Note that this happens silently; other columns are still populated. Using the `NamingStrategy` with max length = 63 does not affect the identifier length in query building, and thus does not address this issue.

maxwey:
I tried to find prior art/related fixes, and found one relevant issue: [PR 6337](https://github.com/go-gorm/gorm/pull/6337)
Though it appears that it was meant to fix schema migration names rather than query generation.

PR #7519, opened 2025-07-16
Fix omission of fields in deeply nested joins with long identifiers (#7513)

IgorStepanov:
- [x] Do only one thing
- [x] Tested

Fixes https://github.com/go-gorm/gorm/issues/7513

### User Case Description

PostgreSQL has 63-character limit for identifiers (including select field aliases).
In a long chain of joins we may have a troubles:
```
err = db.Model(&Table1{}).Debug().
	Joins("Looooooooooooooooooooooooooooooooooooooooooooooooooooooooong").
        Joins("Looooooooooooooooooooooooooooooooooooooooooooooooooooooooong.Table2").
        Joins("Looooooooooooooooooooooooooooooooooooooooooooooooooooooooong.Table2.Table3").Find(&objList).Error
```

This code truncates long aliases similar `Namer.UniqueName` and stores a truncated to original map in `db.Statement` that allow to not touch old fileld mapping strategy.

### DIFF
--- a/callbacks/query.go
+++ b/callbacks/query.go
@@ -40,6 +40,12 @@ func BuildQuerySQL(db *gorm.DB) {
 		}
 	}
 
+	truncatedTableAliases := make(map[string]string)
+
+	if db.Statement.ColumnMapping == nil {
+		db.Statement.ColumnMapping = make(map[string]string)
+	}
+
 	if db.Statement.SQL.Len() == 0 {
 		db.Statement.SQL.Grow(100)
 		clauseSelect := clause.Select{Distinct: db.Statement.Distinct}
@@ -158,11 +164,17 @@ func BuildQuerySQL(db *gorm.DB) {
 							selectColumns, restricted := columnStmt.SelectAndOmitColumns(false, false)
 							for _, s := range relation.FieldSchema.DBNames {
 								if v, ok := selectColumns[s]; (ok && v) || (!ok && !restricted) {
+									aliasName := db.NamingStrategy.JoinNestedRelationNames([]string{tableAliasName, s})
 									clauseSelect.Columns = append(clauseSelect.Columns, clause.Column{
 										Table: tableAliasName,
 										Name:  s,
-										Alias: utils.NestedRelationName(tableAliasName, s),
+										Alias: aliasName,
 									})
+									origTableAliasName := tableAliasName
+									if alias, ok := truncatedTableAliases[tableAliasName]; ok {
+										origTableAliasName = alias
+									}
+									db.Statement.ColumnMapping[aliasName] = utils.NestedRelationName(origTableAliasName, s)
 								}
 							}
 
@@ -232,12 +244,23 @@ func BuildQuerySQL(db *gorm.DB) {
 						}
 
 						parentTableName := clause.CurrentTable
+						parentFullTableName := clause.CurrentTable
 						for idx, rel := range relations {
 							// joins table alias like "Manager, Company, Manager__Company"
 							curAliasName := rel.Name
+
+							var nameParts []string
+							var fullName string
 							if parentTableName != clause.CurrentTable {
-								curAliasName = utils.NestedRelationName(parentTableName, curAliasName)
+								nameParts = []string{parentFullTableName, curAliasName}
+								fullName = utils.NestedRelationName(parentFullTableName, curAliasName)
+							} else {
+								nameParts = []string{curAliasName}
+								fullName = curAliasName
 							}
+							aliasName := db.NamingStrategy.JoinNestedRelationNames(nameParts)
+							truncatedTableAliases[aliasName] = fullName
+							curAliasName = aliasName
 
 							if _, ok := specifiedRelationsName[curAliasName]; !ok {
 								aliasName := curAliasName
@@ -250,6 +273,7 @@ func BuildQuerySQL(db *gorm.DB) {
 							}
 
 							parentTableName = curAliasName
+							parentFullTableName = fullName
 						}
 					} else {
 						fromClause.Joins = append(fromClause.Joins, clause.Join{
--- a/schema/naming.go
+++ b/schema/naming.go
@@ -1,11 +1,13 @@
 package schema
 
 import (
-	"crypto/sha1"
+	"crypto/sha256"
 	"encoding/hex"
 	"regexp"
 	"strings"
 	"unicode/utf8"
+
+	"gorm.io/gorm/utils"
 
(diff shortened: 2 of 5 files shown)


maxwey (commented):
Thank you for taking a crack at this bug! 

I left a couple comments, but am not a maintainer (or even very familiar with Gorm internals)

schema/naming.go
```diff
@@ -95,6 +99,29 @@ func (ns NamingStrategy) UniqueName(table, column string) string {
...
+}
+
+// TruncatedName generate truncated name
+func (ns NamingStrategy) truncateName(ident string) string {
+	formattedName := ident
+	if ns.IdentifierMaxLength == 0 {
+		ns.IdentifierMaxLength = 64
+	}
+
+	if utf8.RuneCountInString(formattedName) > ns.IdentifierMaxLength {
```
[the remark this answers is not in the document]
IgorStepanov: Hmm. Thanks. I'll fix it today.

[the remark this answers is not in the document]
IgorStepanov: Fixed

callbacks/query.go
```diff
@@ -236,7 +245,9 @@ func BuildQuerySQL(db *gorm.DB) {
 							// joins table alias like "Manager, Company, Manager__Company"
 							curAliasName := rel.Name
 							if parentTableName != clause.CurrentTable {
-								curAliasName = utils.NestedRelationName(parentTableName, curAliasName)
+								aliasName := db.NamingStrategy.JoinNestedRelationNames([]string{parentTableName, curAliasName})
+								db.Statement.TruncatedFields[aliasName] = utils.NestedRelationName(parentTableName, curAliasName)
```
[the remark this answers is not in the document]
IgorStepanov: Done

jinzhu:
Would it be possible to reimplement this feature using ColumnMapping in combination with NamingStrategy? Also, the current truncateName implementation seems to share a lot of duplicated logic with formatName.

IgorStepanov:
@jinzhu 

Hello. Sorry for the delay. 
I added using of ColumnMapping (temoorary `truncatedTableAliases` still used for table aliases, but as local variable).
Also I reworked `formatName` to use my function truncateName. However I changed hash function (as linter advised), thus I must have changed tests for it.

Looks like current linter issue is not affected by me.

IgorStepanov:
@jinzhu any comments?

IgorStepanov:
@jinzhu ping

schema/naming.go
```diff
@@ -95,25 +98,66 @@ func (ns NamingStrategy) UniqueName(table, column string) string {
...
+			break
+		}
+		res++
+	}
+
+	if utf8.Valid([]byte(s[len(s)-res:])) {
+		res = 0
+	}
+	return res
+}
```
maxwey: I the logic here is a bit harder to follow; and without comments to explain the intention behind the bitwise operations, it makes might make this code harder to maintain. 

You can combine `truncate` and `brokenTailSize` into:
```golang
// truncateSafely returns a string that is at most the provided maxLength number of bytes.
// If the requested maxLength truncates the string midway in a multi-byte UTF-8 code point,
// it will find the code point's starting byte and truncate the string to remove the entire
// code point to avoid breaking the multi-byte sequence.
func truncateSafely(s string, maxLength int) string {
	if len(s) <= maxLength {
		return s
	}
	if maxLength <= 0 {
		return ""
	}

	lastRuneStartIdx := maxLength
	for lastRuneStartIdx > 0 && !utf8.RuneStart(s[lastRuneStartIdx]) {
		lastRuneStartIdx--
	}

	return s[:lastRuneStartIdx]
}
```


It looks like `utf8.RuneStart()` does effectively the same as your bitwise check, but is self-documenting.

schema/naming.go
```diff
@@ -95,25 +98,66 @@ func (ns NamingStrategy) UniqueName(table, column string) string {
...
+func (ns NamingStrategy) truncateName(ident string) string {
+	formattedName := ident
 	if ns.IdentifierMaxLength == 0 {
 		ns.IdentifierMaxLength = 64
 	}
 
-	if utf8.RuneCountInString(formattedName) > ns.IdentifierMaxLength {
-		h := sha1.New()
+	if len(formattedName) > ns.IdentifierMaxLength {
+		h := sha256.New224()
```
maxwey: I think changing this to SHA-1 could introduce breaking changes for users, since the generated identifier would differ from prior GORM versions for the same input. Unless there’s a strong reason to change the hash function, we should likely preserve the existing behavior here.

As far as I can tell, this hash does not need to be cryptographically secure, so moving to sha256 is likely not necessary. The identifier is used purely to map query results back to the correct fields. The probability that two _valid_ column identifiers both exceed the length limit _and_ collide on the truncated hash is extremely low.

A collision would only be more concerning if arbitrary column identifiers could be supplied via untrusted user input. Even then, it would still be unlikely that any colliding identifiers would correspond to valid columns. That said, allowing arbitrary column selection would be a much larger issue on its own.

If we choose to be extra defensive, one possible mitigation would be to detect and error (or panic) if a column identifier collision ever occurs by checking `truncatedTableAliases` before insertion into the map.

callbacks/query.go
```diff
@@ -158,11 +164,17 @@ func BuildQuerySQL(db *gorm.DB) {
 							selectColumns, restricted := columnStmt.SelectAndOmitColumns(false, false)
 							for _, s := range relation.FieldSchema.DBNames {
 								if v, ok := selectColumns[s]; (ok && v) || (!ok && !restricted) {
+									aliasName := db.NamingStrategy.JoinNestedRelationNames([]string{tableAliasName, s})
```
maxwey: Should this be forming nested relation name based on the possibly truncated `tableAliasName` or the full, un-truncated `origTableAliasName`?

callbacks/query.go
```diff
@@ -158,11 +164,17 @@ func BuildQuerySQL(db *gorm.DB) {
...
 										Table: tableAliasName,
 										Name:  s,
-										Alias: utils.NestedRelationName(tableAliasName, s),
+										Alias: aliasName,
 									})
+									origTableAliasName := tableAliasName
+									if alias, ok := truncatedTableAliases[tableAliasName]; ok {
+										origTableAliasName = alias
+									}
+									db.Statement.ColumnMapping[aliasName] = utils.NestedRelationName(origTableAliasName, s)
```
maxwey: nit: would `origTableAliasName` be better named as "fullTableName"? It's perhaps not clear what "orig" means as we are not visibly modifying the `tableAliasName` here.

PR #7526, opened 2025-07-22
optimize: performance optimization

liov:
- [x] Do only one thing
- [x] Tested

This PR is for: https://github.com/go-gorm/gorm/pull/7492, doesn't completely solve the problem,, and only optimizes the performance for FieldByName to avoid every for loop

The implementation of https://github.com/go-gorm/gorm/pull/7492 does not completely solve the problem and introduces performance issues

The following is a reprinted comment
```go
type Age struct {
	Age int
}

func TestGORM(t *testing.T) {
	man := Man{Id: 1, Name: "Man1", Age: 1}
	DB.Create(&man)

	// update data
	idx := Man{Id: 1}

	change1 := Man{
		Age: 10,
	}
	change2 := map[string]interface{}{
		"age": 10,
	}
	change3 := struct {
		Age
	}{Age{Age: 10}}

	_, _, _ = change1, change2, change3
	var err error

	// change1 -> yes
	err = idx.update(change1)
	if err != nil {
		return
	}

	// change2 -> yes
	err = idx.update(change2)
	if err != nil {
		return
	}

	// change3 -> yes
	err = idx.update(change3)
	if err != nil {
		return
	}
}
```
After a lot of testing and consideration, I still don't think this PR should be merged.

Firstly, FieldByName performs poorly with large structs, which makes caching fields ineffective and unnecessary.

Secondly, field.ValueOf only addresses one case — the default implementation remains problematic. Additionally, field.ReflectValueOf hasn't been modified at all.
```go
	switch {
	case len(field.StructField.Index) == 1 && fieldIndex >= 0:
		field.ValueOf = func(ctx context.Context, value reflect.Value) (interface{}, bool) {
			fieldValue := reflect.Indirect(value).FieldByName(field.Name)
			return fieldValue.Interface(), fieldValue.IsZero()
		}
	default:
			v = reflect.Indirect(v)
			for _, fieldIdx := range field.StructField.Index {
				if fieldIdx >= 0 {
					v = v.Field(fieldIdx)
...
}
```
And filed. FindByName does not solve problems similar to the following
```go
type Age struct {
	Age int
}
db.Updates(struct {
		Age
	}{Age{Age: 30}})
```

### DIFF
--- a/schema/field.go
+++ b/schema/field.go
@@ -448,7 +448,7 @@ func (schema *Schema) ParseField(fieldStruct reflect.StructField) *Field {
 }
 
 // create valuer, setter when parse struct
-func (field *Field) setupValuerAndSetter() {
+func (field *Field) setupValuerAndSetter(modelType reflect.Type) {
 	// Setup NewValuePool
 	field.setupNewValuePool()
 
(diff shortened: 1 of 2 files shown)


jinzhu:
Maybe we should introduce another field `DestSchema`, and use DestSchema specifically to set field values.

PR #7529, opened 2025-07-22
feat: Provide detailed error messages for field.Set failures

iseki0:
- [x] Do only one thing
- [x] Tested

Provide a detailed error message during Create/Save/Update...

### User Case Description
Currently the error message too simple to diagnostic what's happened.

It's relative to #7510. I don't think the code is good. But to achieve the goal, we have to change our minds on error handling.
My opinion is create a customized error type, even it's internal. But looks like you don't like it? (Of course I understand it's difficult to make decision about make the error type public or not)


jinzhu:
Hi @iseki0 

I means we can rewrite more details error in each `Set` func like this one https://github.com/go-gorm/gorm/blob/d3b49c3d235b670ba7fb2d5558dfa48033f7b9d8/schema/field.go#L852

PR #7569, opened 2025-08-27
fix: Updates change from T to interface{}

smalls0098:
fix Updates() to use map

https://gorm.io/zh_CN/docs/update.html

// Update attributes with `map`
err := gorm.G[User](db).Where("id = ?", 111).Updates(ctx, map[string]interface{}{"name": "hello", "age": 18, "active": false})
// UPDATE users SET name='hello', age=18, active=false, updated_at='2013-11-17 21:34:10' WHERE id=111;


dungmv:
what happen with "Update attributes with map"

PR #7574, opened 2025-08-31
fix slogLogger to support ParameterizedQueries Config

EricWvi:
- [x] Do only one thing
- [x] Tested

Add ParamsFilter method to slogLogger

### User Case Description
After update, `logger.NewSlogLogger` supports `ParameterizedQueries`. Now, it has no effect.

### DIFF
--- a/logger/slog.go
+++ b/logger/slog.go
@@ -88,3 +88,11 @@ func (l *slogLogger) Trace(ctx context.Context, begin time.Time, fc func() (sql 
 		})
 	}
 }
+
+// ParamsFilter filter params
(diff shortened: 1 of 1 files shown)


EricWvi:
Hi, @jinzhu @rezamokaram. Please take a look.

EricWvi:
ref.

https://github.com/go-gorm/gorm/blob/cace4a6eb85715c8fd648f76728ec554ecafb8e5/logger/logger.go#L193-L198

rezamokaram:
Thanks for your attention and contribution.
@EricWvi

Issue #7192, opened 2024-09-09
Support go1.23 iterators


ilya-hontarau:
## Describe the feature

Allow to pass iterators in methods:

```
db.Delete(&User, "name  IN (?)", maps.Keys(namesToIDs))

users := []User{{Name: "jinzhu1"}, {Name: "jinzhu2"}, {Name: "jinzhu3"}}
err = db.Create(slices.Values(users)).Error
```

Now it returns errors:
```
panic: failed to encode args[0]: unable to encode (iter.Seq[string])(0x102f20cf0) into text format for text (OID 25): cannot find encode plan

or 

panic: unsupported data type: iter.Seq[main.Test]: Table not set, please set it like: db.Model(&user) or db.Table("users")

```

## Motivation

Support new go standard library features.

abdyek:
Hi! @jinzhu I can do it. You can assign me this issue?

peterdeme:
Fix: https://github.com/go-gorm/gorm/pull/7577

PR #7577, opened 2025-09-05
Add support for Go 1.23 iter.Seq as SQL parameter values

peterdeme:
- [x] Do only one thing
- [x] Tested

Fixes: https://github.com/go-gorm/gorm/issues/7192

```go
// Note that maps.Keys() returns an `iter.Seq` element in Go 1.23+
db.Delete(&User, "name  IN (?)", maps.Keys(namesToIDs))

panic: failed to encode args[0]: unable to encode (iter.Seq[string])(0x102f20cf0) into text format for text (OID 25): cannot find encode plan
```

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -243,7 +243,19 @@ func (stmt *Statement) AddVar(writer clause.Writer, vars ...interface{}) {
 			writer.WriteString(subdb.Statement.SQL.String())
 			stmt.Vars = subdb.Statement.Vars
 		default:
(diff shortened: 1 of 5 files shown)


peterdeme:
@jinzhu mind taking a look?

jinzhu:
Thanks for the PR, but I think this change may not be necessary — it’s probably better to keep the core lookup simple.
Users can handle this externally with a small helper to convert names before passing them to GORM, which avoids extra runtime checks

Thank you.

PR #7612, opened 2025-10-12
fix #7602: copy-paste typo

### LINKED ISSUE #7602: `DB.ScanRows()` into `int` variants skips all rows and returns last row
https://github.com/go-gorm/gorm/blob/b88148363a954f69fa680b152dfd96a94ffea1e1/scan.go#L179-L189
Here must be `if` rather then `for`. So now `DB.ScanRows()` into `int` variants silently eats all the rows and returns the last one.

Example:
```go
func list[T any](tx *gorm.DB) ([]T, error) {
	result := make([]T, 0)
	rows, err := tx.Raw(
		`(SELECT 10 AS x UNION SELECT 20 UNION SELECT 30) ORDER BY x`,
	).Rows()
	if err != nil {
		return nil, err
	}
	defer rows.Close()
	for rows.Next() {
		var item T
		err = tx.ScanRows(rows, &item)
		if err != nil {
			return nil, err
		}
		result = append(result, item)
	}
	return result, nil
}

func main() {
	db, err := gorm.Open(...)

	type Ref struct{ X int }
	structs, err := list[Ref](tx)
	if err != nil {...}
	fmt.Printf("structs: %#v\n", structs)

	ints, err := list[int](tx)
	if err != nil {...}
	fmt.Printf("ints: %#v\n", ints)
}
```
output:
```
structs: []main.Ref{main.Ref{X:10}, main.Ref{X:20}, main.Ref{X:30}}
ints: []int{30}
```

- go 1.24
- gorm 1.30 , 1.31

Vovan-VE:
- [X] Do only one thing
- [X] Tested

Unexpected `for` loop consumes all the rows from the result set. Looks like it's a copy-paste mistake from similar slice `case`. Here must be `if` instead.

### User Case Description

```go
for rows.Next() {
	var id int64
	if err := tx.ScanRows(rows, &id); err != nil {
		return err
	}
	// ... id ...
}
```

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -182,7 +182,7 @@ func Scan(rows Rows, db *DB, mode ScanMode) {
 		*bool, *string, *time.Time,
 		*sql.NullInt32, *sql.NullInt64, *sql.NullFloat64,
 		*sql.NullBool, *sql.NullString, *sql.NullTime:
-		for initialized || rows.Next() {
+		if initialized || rows.Next() {
 			initialized = false
 			db.RowsAffected++
 			db.AddError(rows.Scan(dest))


jinzhu:
thank you for your pr, that’s intentional, it ensures all returned rows are read so the last one can be selected.

Vovan-VE:
@jinzhu , are you sure? Destination is a single scalar value like `int`, not `[]int` to take them all. Since `query.Take(&model)` and `query.Take(&id)` works as it looks, why `for rows.Next()` works only for `&model`, but not for `&id`? It's confusing and counterintuitive.
```go
for rows.Next() {
	var id int64
	if err := tx.ScanRows(rows, &id); err != nil {
		return err
	}
	// ... id ...
}
```

Issue #7620, opened 2025-10-20
`Tabler` interface does not support dynamic table name


lvan100:
## GORM Playground Link

I create a PR [https://github.com/go-gorm/gorm/pull/7624](https://github.com/go-gorm/gorm/pull/7624) for this issuse.

## Description

When I was using **GORM v1**, it supported **dynamic table names** through the `Tabler` interface:

```go
type UserWithTable struct {
	ID   int64
	Name string
}

func (u *UserWithTable) TableName() string {
	return fmt.Sprintf("gorm.user_%d", u.ID%1000)
}
```

However, after upgrading to **GORM v2**, I found that it no longer works. This is because GORM v2 uses a **zero-value model object** instead of the original instance, so the dynamic information is lost. The relevant code is shown below:

`schema/schema.go`
```go
func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Namer, specialTableName string) (*Schema, error) {
	...

	var tableName string
	modelValue := reflect.New(modelType) // not source object
	if specialTableName != "" {
		tableName = specialTableName
	} else if en, ok := namer.(embeddedNamer); ok {
		tableName = en.Table
	} else if tabler, ok := modelValue.Interface().(Tabler); ok {
		tableName = tabler.TableName()
	} else if tabler, ok := modelValue.Interface().(TablerWithNamer); ok {
		tableName = tabler.TableName(namer)
	} else {
		tableName = namer.TableName(modelType.Name())
	}

	...
}
```

jinzhu:
This is by design in GORM v2 — we only allow immutable table names to avoid the misuses that happened in v1.
You can use **Scopes** to set the correct table name instead.
See: [https://gorm.io/docs/v2_release_note.html#Table-Name](https://gorm.io/docs/v2_release_note.html#Table-Name)

PR #7624, opened 2025-10-21
fix(schema): using `dest` to determine the `Tabler` or `TablerWithNamer` first

lvan100:
Using `dest` to determine the `Tabler` or `TablerWithNamer` first, in order to support dynamic table names.

- [x] Do only one thing
- [x] Tested

This PR aims to solve this problem, [https://github.com/go-gorm/gorm/issues/7620](https://github.com/go-gorm/gorm/issues/7620). The code before repair cann't use field `ID`.

```go
type UserWithTable struct {
	ID   int64
	Name string
}

func (u *UserWithTable) TableName() string {
	return fmt.Sprintf("gorm.user_%d", u.ID%1000)
}
```

### User Case Description

As above.

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -174,9 +174,21 @@ func ParseWithSpecialTableName(dest interface{}, cacheStore *sync.Map, namer Nam
 	} else if en, ok := namer.(embeddedNamer); ok {
 		tableName = en.Table
 	} else if tabler, ok := modelValue.Interface().(Tabler); ok {
-		tableName = tabler.TableName()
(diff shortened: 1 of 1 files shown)


jinzhu:
This is by design in GORM v2 — we only allow immutable table names to avoid the misuses that happened in v1.
You can use **Scopes** to set the correct table name instead.
See: [https://gorm.io/docs/v2_release_note.html#Table-Name](https://gorm.io/docs/v2_release_note.html#Table-Name)

PR #7627, opened 2025-10-22
Fix Count: honor custom SELECT count expressions, avoid extra columns

StounhandJ:
- [x] Do only one thing
- [x] Tested

When creating a Count request, if a Select was passed with count, substitute it in Clause to avoid unexpected addition of the request.

### User Case Description

InnerJoins is used here as adding an internal verification condition.
```go
r.db.Model(&models.KubeRBAC{}).
	Select("count(distinct(subj->>'kind', subj->>'apiGroup', subj->>'name', subj->>'namespace', \"ClusterRef\".id))").
	InnerJoins("ClusterRef").
	Count(&total)
```

Extra fields appear in the final sql, which violates the correctness of the work.
```sql
SELECT
  count(distinct(
    subj->>'kind',
    subj->>'apiGroup',
    subj->>'name',
    subj->>'namespace',
    "ClusterRef".id
  )),
  "ClusterRef"."id"         AS "ClusterRef__id",
  "ClusterRef"."name"       AS "ClusterRef__name",
  "ClusterRef"."updated_at" AS "ClusterRef__updated_at",
  "ClusterRef"."created_at" AS "ClusterRef__created_at"
FROM
  "kube_rbac"
INNER JOIN
  "kube_clusters" "ClusterRef"
    ON "kube_rbac"."kube_cluster_id" = "ClusterRef"."id"
    AND "ClusterRef"."updated_at" > now() - '1 hours'::interval;
```

The result after correction
```sql
SELECT
  count(distinct(
    subj->>'kind',
    subj->>'apiGroup',
    subj->>'name',
    subj->>'namespace',
    "ClusterRef".id
  ))
FROM
  "kube_rbac"
INNER JOIN
  "kube_clusters" "ClusterRef"
    ON "kube_rbac"."kube_cluster_id" = "ClusterRef"."id"
    AND "ClusterRef"."updated_at" > now() - '1 hours'::interval;
```

### DIFF
--- a/tests/count_test.go
+++ b/tests/count_test.go
@@ -95,6 +95,16 @@ func TestCount(t *testing.T) {
 		t.Fatalf("Build count with select, but got %v", result.Statement.SQL.String())
 	}
 
+	result = dryDB.Model(&User{}).Distinct("name").Joins("Team").Count(&count)
+	if !regexp.MustCompile(`SELECT COUNT\(DISTINCT\(.name.\)\) FROM .*users.* JOIN .*users.*`).MatchString(result.Statement.SQL.String()) {
+		t.Fatalf("Build count with select, but got %v", result.Statement.SQL.String())
+	}
+
+	result = dryDB.Model(&User{}).Select("COUNT(DISTINCT(`name`))").Joins("Team").Count(&count)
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @StounhandJ 

Can you add some tests?

StounhandJ:
Hi @jinzhu 

I missed the message, but here are two Count tests along with the Join

StounhandJ:
Hi @jinzhu

This change is necessary to remove the existing workarounds. Without it, we are forced to keep temporary patches in place, which is not sustainable in the long term.

Issue #7600, opened 2025-09-29
Support escaping double quotes in struct tags


trungnt2910:
## Describe the feature

Support escaping double quotes in struct tags.

## Motivation

I have the following object and want to do a [PostgreSQL full text search index](https://www.postgresql.org/docs/current/textsearch-tables.html) on the column `Name`:

```
type Object struct {
	Key uuid.UUID `gorm:"primaryKey;type:uuid;default:gen_random_uuid()"`
	Name string `gorm:"not null"`
}
```

In SQL this would be:
```sql
CREATE INDEX idx_Object_Name ON "Object" USING GIN (to_tsvector('english', "Name"));
```

To do this in a GORM struct tag, I need to be able to put `to_tsvector('english', "Name")` with double quotes as the `expression` parameter. However, this doesn't seem to be supported or at least documented.

PR #7631, opened 2025-10-27
fix: Allow escaped double quotes in struct tag parser

kankankanp:
- [x] Do only one thing  
- [x] Tested  

This pull request adds support for escaping double quotes (\" → ") in struct tag parsing.  
Now you can write struct tags like `gorm:"expression:to_tsvector('english', \"Name\")"` and the parser will correctly interpret the escaped double quote.

This change addresses [#7600](https://github.com/go-gorm/gorm/issues/7600).

### User Case Description

When using PostgreSQL full text search, it is often necessary to refer to a column name with double quotes (e.g., `"Name"`).  
Previously, GORM's struct tag parser did not support writing double quotes inside the tag expression.  
With this change, you can escape double quotes using a backslash, and the intended SQL expression will be parsed and used correctly.

**Example:**
```go
type Object struct {
    Name string `gorm:"expression:to_tsvector('english', \"Name\")"`
}
```
This struct tag will now be correctly interpreted as:
`to_tsvector('english', "Name")`


kankankanp:
@jinzhu 
If possible, I would be grateful if you could review this PR.

Issue #7274, opened 2024-11-11
Panic saving a struct with nil value field and json serializer


mytlogos:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/770

## Description

When i save a struct with a json serializer field and nil value, gorm panics with this stacktrace:

--- FAIL: TestGORM_failure (0.01s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x8c5295]

goroutine 53 [running]:
testing.tRunner.func1.2({0xf0dd80, 0x15d6180})
        /usr/local/go/src/testing/testing.go:1631 +0x3f7
testing.tRunner.func1()
        /usr/local/go/src/testing/testing.go:1634 +0x6b6
panic({0xf0dd80?, 0x15d6180?})
        /usr/local/go/src/runtime/panic.go:770 +0x132
gorm.io/gorm/schema.(*Field).setupValuerAndSetter.func17({0x1109750, 0x25cda20}, {0xf5ab00?, 0xc00034b280?, 0x0?}, {0xf15fa0, 0xc000350000})
        /media/MintData/clones/gorm-playground/gorm/schema/field.go:964 +0x1d5
gorm.io/gorm/callbacks.ConvertToAssignments.func2(0xc0000fd400, {0xf15fa0, 0xc000350000})
        /media/MintData/clones/gorm-playground/gorm/callbacks/update.go:149 +0x126
gorm.io/gorm/callbacks.ConvertToAssignments(0xc000450700)
        /media/MintData/clones/gorm-playground/gorm/callbacks/update.go:288 +0x187d
gorm.io/gorm/callbacks.RegisterDefaultCallbacks.Update.func7(0xc000296a80)
        /media/MintData/clones/gorm-playground/gorm/callbacks/update.go:74 +0x326
gorm.io/gorm.(*processor).Execute(0xc000348a00, 0xc000296a80)
        /media/MintData/clones/gorm-playground/gorm/callbacks.go:130 +0xa78
gorm.io/gorm.(*DB).Save(0xc00030dd70, {0xec8c00, 0xc00034b280})
        /media/MintData/clones/gorm-playground/gorm/finisher_api.go:106 +0x9b8
gorm.io/playground.TestGORM_failure(0xc00044eb60?)
[...]

PR #7634, opened 2025-10-29
Fix nil pointer panic in serializer for nil any-type fields

doraemonkeys:
- [x] Do only one thing
- [x] Tested

Fix: https://github.com/go-gorm/gorm/issues/7274

Added a nil check before calling `s.Serializer.Scan()`. When both `s.fieldValue` and `s.Serializer` are nil, the field is now set to its zero value and returns early, preventing the panic.

#### Problem
When using `any` type fields, if both `s.fieldValue` and `s.Serializer` are nil, the code would directly call `s.Serializer.Scan()` at line 974, causing a nil pointer panic.

### DIFF
--- a/tests/serializer_test.go
+++ b/tests/serializer_test.go
@@ -227,3 +227,59 @@ func TestSerializerAssignFirstOrCreate(t *testing.T) {
 	AssertEqual(t, result.Roles, data.Roles)
 	AssertEqual(t, result.JobInfo.Location, data.JobInfo.Location)
 }
+
+// Test for: panic when serializer field with any type is nil
+func TestSerializerWithAnyType(t *testing.T) {
+	type ProductWithAny struct {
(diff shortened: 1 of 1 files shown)


jinzhu:
Can you add some tests?

doraemonkeys:
> Can you add some tests?

Hi, thanks for the review! I've added comprehensive test cases for this fix.

doraemonkeys:
@jinzhu , this is a PR with no additional side effects, could you merge it as soon as possible?

co60ca:
@jinzhu We appreciate your work on this project. We are experiencing this also. We would appreciate your attention on merging this?

By the way this appears to be a regression, since it was working in `gorm.io/gorm v1.23.8`, also we tested doraemonkeys patch and it appears to passing our tests.

PR #7652, opened 2025-11-13
Fix #6557: Avoid redundant column migrations when using type aliases

### LINKED ISSUE #6557: Postgres: AutoMigrate re-migrates everytime for some types
## GORM Playground Link

https://github.com/go-gorm/playground/pull/646

## Description

Every time we run AutoMigrage we get re-migration for certain specification of types:
```
# Types
type:character varying
type:timestamp with time zone
type:varchar[]
```

A workaround is to specify the types in a different way (see reproduction) but the required types seems sphoradic.
Not sure why, but the issue did not force migration before we upgraded GORM and the Postgres driver.

See the attached reproduction for details.

DasJott:
- [ X ] Do only one thing
- [ X ] Non breaking API changes
- [ X ] Tested

Improves the mapping of types to aliases.  

The mapping keys do not (and CAN not) include any additional information behind the actual type name.  
Therefore to check for aliases in a certain mapping we need to strip that info off, like:  
`varchar(255)[] => varchar`.
To really be sure for the mapping to work reliably the types are mapped best in both ways, like:  
`varchar => character varying`, `character varying => varchar`.  
For at least postgres this was not the case for exactly that case.
For not having to touch all the other driver packages, this code simply tries to map the other way around when the first attempt failed.
 

### User Case Description

We had a data struct containing a field `pq.StringArray` with `gorm:"type:varchar[]"`. That type was returned by postgres as `character varying[]`. The current code could not map that correctly and therefore on every service restart an unnecessary migration was started (which took very long on a table containing millions of entries).  

This is fixed with this PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,3 +7,8 @@ require (
 	github.com/jinzhu/now v1.1.5
 	golang.org/x/text v0.20.0
 )
+
+require (
+	github.com/mattn/go-sqlite3 v1.14.22 // indirect
+	gorm.io/driver/sqlite v1.6.0 // indirect
+)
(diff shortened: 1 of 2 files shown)


jinzhu:
Hi @DasJott 

Can you add some tests？

jinzhu:
Hi @DasJott 

Can you add some tests？

PerryProjects:
Hey @jinzhu, 
Hey @DasJott,

i added some tests for the given changes. Hopefully the tests were sufficient.

PR #7655, opened 2025-11-20
Add support for dynamic GORM valuer registration and retrieval

iseki0:
- [x] Do only one thing
- [x] Tested

Since some bugly database is popular, everyone have to workaround to it.

### User Case Description

I have a library https://github.com/iseki0/goda, I don't think it's a better choice to use the `GormValuer` interface directly, it will make a strong dependency to gorm, and it's unwise since not everyone use gorm. So please give me a chance to register the handler dynamiclly, if someone need, they can register the gorm-enhance-module by the effect import statments, just time `time/tzdata`.

#### About the workaround

The MySQL has a very strange behavior:
```sql
prepare aa from 'select cast(? as time), ?';
set @a '09:00:00';
execute aa using @a, @a;
```
```
+-----------------+----------+
| cast(? as time) | ?        |
+-----------------+----------+
| 00:00:00        | 09:00:00 |
+-----------------+----------+
```
The problem is on MySQL server, not driver. The Go MySQL driver do paratermized query like the prepare command, I refer to the https://github.com/go-sql-driver/mysql.

The solution is:
```sql
prepare aa from 'select cast(cast(? as char) as time)';
```

Do you have any suggestions? Maybe we should fix it in the MySQL dialect. 毕竟，又不是只有我会倒这个霉，所有传字符串的都会完蛋。

### DIFF
--- a/statement.go
+++ b/statement.go
@@ -243,6 +243,10 @@ func (stmt *Statement) AddVar(writer clause.Writer, vars ...interface{}) {
 			writer.WriteString(subdb.Statement.SQL.String())
 			stmt.Vars = subdb.Statement.Vars
 		default:
+			if valuer := GetDynamicGormValuer(reflect.TypeOf(v)); valuer != nil {
+				stmt.AddVar(writer, valuer(stmt.Context, stmt.DB, v))
+				return
+			}
 			switch rv := reflect.ValueOf(v); rv.Kind() {
 			case reflect.Slice, reflect.Array:
 				if rv.Len() == 0 {
--- a/valuer.go
+++ b/valuer.go
@@ -1,0 +1,19 @@
+package gorm
+
+import (
+	"context"
+	"reflect"
+
+	"gorm.io/gorm/clause"
+)
+
+var dynamicRegisteredGormValuer = make(map[reflect.Type]func(context.Context, *DB, any) clause.Expr)
+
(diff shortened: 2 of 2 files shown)


iseki0:
@jinzhu Just a proposal, implementation is a simple demo... Please just ignore the AI reviews😁

jinzhu:
I don't get what the PR for, maybe you can just use the database/sql's Valuer interface?

iseki0:
> I don't get what the PR for, maybe you can just use the database/sql's Valuer interface?

@jinzhu 
The root cause is, on MySQL:
```mysql
PREPARE aa FROM 'select cast(? as time), ?';
SET @a '09:00:00';
EXECUTE aa USING @a, @a;
```
The result is:
```
+-----------------+----------+
| cast(? as time) | ?        |
+-----------------+----------+
| 00:00:00        | 09:00:00 |
+-----------------+----------+
```
MySQL is bugly, I found a solution:
```mysql
PREPARE aa FROM 'select cast(cast(? as char) as time)'
```
I don't know why MySQL is that exactly. Just workaround it,
I consider use your GormValuer interface, but it will make the library hard-dependent to gorm. It's not a good choice.

If the gorm allowing register Valuer dynamically, we can make it like `import _ "time/tzdata"`.

PR #7687, opened 2026-01-12
Fix/preload composite foreign keys

### LINKED ISSUE #7686: preload doesn't work in gorm 1.31.1 works in 1.31.0
## GORM Playground Link

I don't understand how to use the playground, but I added a self contained program that reproduces it.

https://github.com/go-gorm/playground/pull/840

## Description

When building against 1.31.0, Preload for this testcase works, but when building against 1.31.1 it does not

## Works correctly with gorm v1.31.0

Note User and Location are both non-nil as expected

```
bigbaby:gorm-preload-bug tadhunt$ grep gorm.io go.mod
	gorm.io/gorm v1.31.0
bigbaby:gorm-preload-bug tadhunt$ make
go mod tidy
go vet
staticcheck
go build -o gorm-bug
bigbaby:gorm-preload-bug tadhunt$ ./gorm-bug

=== Loading shift with Preload ===

Shift ID: shift-001
Shift UserID: 1001
Shift LocationID: 2001
User: Alice Smith (ID: 1001)
Location: Office A (ID: 2001)
```

## Fails with gorm v1.31.1

```
bigbaby:gorm-preload-bug tadhunt$ grep gorm.io go.mod
	gorm.io/gorm v1.31.1
bigbaby:gorm-preload-bug tadhunt$ make
go mod tidy
go vet
staticcheck
go build -o gorm-bug
bigbaby:gorm-preload-bug tadhunt$ ./gorm-bug

=== Loading shift with Preload ===

Shift ID: shift-001
Shift UserID: 1001
Shift LocationID: 2001
ERROR: User is nil - Preload failed!
```

kankankanp:
- [x] Do only one thing
  - [x] Tested

  Fix `LookUpField` to prevent returning wrong embedded fields when using namer-based column lookup.

  When searching for a field like "UserID" in a schema with embedded structs, the namer-based lookup (added in #7619) would incorrectly return embedded fields with matching DBNames. For example, searching for "UserID" would return an embedded "ID" field with DBName "user_id" instead of returning nil.

  This fix adds a check to ensure the namer-based lookup only returns a field if its `Name` exactly matches the search name.

  **Changes:**
  - `schema/schema.go`: Add name matching check in `LookUpField`
  - `schema/schema_test.go`: Add regression test `TestLookUpFieldWithEmbeddedStruct`

  ### User Case Description

  Using composite foreign keys with embedded structs causes Preload to fail in v1.31.1.

  ```go
  type SlingUser struct {
      ID    int64      `gorm:"primaryKey"`
      SrcID string     `gorm:"primaryKey"`
      User  EmbeddedUser `gorm:"embedded;embeddedPrefix:user_"`
  }

  type SlingShift struct {
      ID     string `gorm:"primaryKey"`
      SrcID  string `gorm:"primaryKey"`
      UserID int64
      User   *SlingUser `gorm:"foreignKey:SrcID,UserID;references:SrcID,ID"`
  }

  // This fails in v1.31.1, works in v1.31.0
  db.Preload("User").First(&shift)
  // ERROR: User is nil - Preload failed!
```

  Fixes #7686

### DIFF
--- a/schema/schema_test.go
+++ b/schema/schema_test.go
@@ -350,3 +350,44 @@ func TestCompositePrimaryKeyWithAutoIncrement(t *testing.T) {
 		t.Fatalf("PrioritizedPrimaryField of non autoincrement composite key should be nil")
 	}
 }
+
+// TestLookUpFieldWithEmbeddedStruct tests that LookUpField does not return
+// embedded fields when searching for a field name that doesn't exist in the schema.
+// This is a regression test for issue #7686.
+func TestLookUpFieldWithEmbeddedStruct(t *testing.T) {
+	// EmbeddedUser represents an embedded struct with ID field
+	type EmbeddedUser struct {
+		ID    int64  `gorm:"column:user_id"`
+		Email string `gorm:"column:user_email"`
+	}
+
+	// ModelWithEmbedded has an embedded struct and its own fields
+	type ModelWithEmbedded struct {
+		ID     int64        `gorm:"primaryKey"`
+		SrcID  string       `gorm:"primaryKey"`
+		UserID int64        // This is a separate field, not the embedded ID
+		User   EmbeddedUser `gorm:"embedded;embeddedPrefix:user_"`
+	}
+
+	s, err := schema.Parse(&ModelWithEmbedded{}, &sync.Map{}, schema.NamingStrategy{})
+	if err != nil {
+		t.Fatalf("failed to parse schema: %v", err)
+	}
+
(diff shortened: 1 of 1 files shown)


kankankanp:
@jinzhu 
I've submitted a PR. Could you please take a look when you have a moment?

arraytad:
I can confirm that building against this PR, in addition to fixing the testcase from #7686 also fixes my issue in the production codebase that I extracted the testcase from.  Thank you for such a fast review + fix! 💚

Issue #6391, opened 2023-06-07
`gorm.Statement.QuoteTo(writer clause.Writer, field interface{})` doesn't do what I would expect for a `clause.Table` with an `Alias` field


tempoz:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/614

## Description

We see [here](https://github.com/go-gorm/gorm/blob/7dd702d379bca5e28a2b65278b5a373f326c72ca/clause/expression_test.go#L128-L134) that the expected output (according to the tests) when building a `clause.Table{Table: "table", Alias: "alias"}` using a `Statement` as the `Builder` is `table alias` or `` `table` `alias` ``, as produced by `gorm.Statement.QuoteTo` [here](https://github.com/go-gorm/gorm/blob/7dd702d379bca5e28a2b65278b5a373f326c72ca/statement.go#L91-L105), but that isn't actually what the SQL should look like as far as I can tell.

Rather, it should produce `table AS alias` or `` `table` AS `alias` ``, which would be usable in the case of e.g. `clause.Insert{Table: clause.Table{Name: "table", Alias: "alias"}}`, which would then produce the working SQL `INSERT INTO table AS alias` or `` INSERT INTO `table` AS `alias` `` as opposed to the non-functional `INSERT INTO table alias` or `` INSERT INTO `table` `alias` ``, which is what it currently does. In fact, we can see that this is how aliases work for columns only [several lines above](https://github.com/go-gorm/gorm/blob/7dd702d379bca5e28a2b65278b5a373f326c72ca/clause/expression_test.go#L120-L126) in the tests.

### What I expect to see

When I run

```go
package main

import (
  "database/sql"
  "fmt"
  "time"

  "gorm.io/driver/sqlite"
  "gorm.io/gorm"
  "gorm.io/gorm/clause"
)

type User struct {
  ID           uint
  Name         string
  Email        *string
  Age          uint8
  Birthday     *time.Time
  MemberNumber sql.NullString
  ActivatedAt  sql.NullTime
  CreatedAt    time.Time
  UpdatedAt    time.Time
}

func main() {
  db, err := gorm.Open(sqlite.Open("test.db"), &gorm.Config{})
  if err != nil {
    panic("failed to connect database")
  }
  insertWithTableAliasDemo(db)
}

func insertWithTableAliasDemo(db *gorm.DB) {
  // Migrate the schema
  db.AutoMigrate(&User{})

[...]

PR #7706, opened 2026-02-11
Use `" AS "` when aliasing tables

tempoz:
- [X] Do only one thing
- [X] Tested

Addresses #6391

Uses `" AS "` instead of `" "` between a table and its alias.

### User Case Description

As described in the related issue, #6391:

> My specific use case is that I want to be able to use the table alias in `clause.OnConflict.DoUpdates` because the table names I'm working with are mixed case, which means I need to manually quote them to reference them directly in the expressions for those updates when I'm using postgres (and I have to use a table name or alias to disambiguate the column from the row that failed to insert, which is available as `excluded`), which I do not want to do.

### DIFF
--- a/clause/expression_test.go
+++ b/clause/expression_test.go
@@ -137,11 +137,11 @@ func TestNamedExpr(t *testing.T) {
 	}, {
 		SQL:    "?",
 		Vars:   []interface{}{clause.Table{Name: "table", Alias: "alias"}},
-		Result: "`table` `alias`",
+		Result: "`table` AS `alias`",
 	}, {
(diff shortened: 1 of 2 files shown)


tempoz:
The failures in the checks seem ignorable:

- the linter is complaining about not checking the error returned by `writer.WriteString`, but that matches both the behavior that was there before when we were calling `writer.WriteByte` as well as all other calls to `writer.WriteString` in the rest of the function
- the mysql test failures seem wholly unrelated to this change.

tempoz:
Is there anything blocking this getting merged?

PR #7711, opened 2026-02-23
fix: Use DO NOTHING when associated model lacks primary keys

### LINKED ISSUE #6622: Incorrect ON CONFLICT when using composite primary key
## GORM Playground Link

https://github.com/go-gorm/playground/pull/652

## Explain your user case and expected results
I'm using a composite primary key for a table that references a list of entities in another table. 

The error I get is
```
2023/10/02 15:24:09 testing postgres...

2023/10/02 15:24:10 gorm-playground/main_test.go:13 ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)
[0.952ms] [rows:0] INSERT INTO "tags" ("pet_first_name","pet_last_name","name") VALUES ('Hasso','McDog','foo') ON CONFLICT DO UPDATE SET "pet_first_name"="excluded"."pet_first_name","pet_last_name"="excluded"."pet_last_name"

2023/10/02 15:24:10 gorm-playground/main_test.go:13 ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)
[3.761ms] [rows:1] INSERT INTO "pets" ("first_name","last_name") VALUES ('Hasso','McDog')
--- FAIL: TestGORM (0.00s)
    main_test.go:14: ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)
FAIL
FAIL	gorm.io/playground	0.238s
```

veeceey:
Fixes #6622

When saving has-many associations where the child model doesn't have explicit `primaryKey` tags (or has no `ID` field), `onConflictOption` would still try to generate `ON CONFLICT DO UPDATE SET ...` but without any conflict target columns. PostgreSQL rightfully rejects this:

```
ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)
```

The generated SQL looked like:
```sql
INSERT INTO "tags" (...) VALUES (...) ON CONFLICT DO UPDATE SET "pet_first_name"=...
```

when it should either have columns after `ON CONFLICT` or use `DO NOTHING`.

This PR adds a check: if the associated schema has no primary key columns, we fall back to `ON CONFLICT DO NOTHING` rather than generating invalid SQL.

All existing tests pass.

### DIFF
--- a/callbacks/associations.go
+++ b/callbacks/associations.go
@@ -364,6 +364,15 @@ func onConflictOption(stmt *gorm.Statement, s *schema.Schema, defaultUpdatingCol
 			onConflict.Columns = append(onConflict.Columns, clause.Column{Name: dbName})
 		}
 
+		// If the associated model has no primary key columns, we cannot build a valid
+		// ON CONFLICT (columns) DO UPDATE clause. Fall back to DO NOTHING to avoid
+		// generating invalid SQL (e.g. ON CONFLICT DO UPDATE SET ... without column
+		// specification, which PostgreSQL rejects).
+		if len(onConflict.Columns) == 0 {
+			onConflict.DoNothing = true
+			return
+		}
+
 		onConflict.UpdateAll = stmt.DB.FullSaveAssociations
 		if !onConflict.UpdateAll {
 			onConflict.DoUpdates = clause.AssignmentColumns(defaultUpdatingColumns)
--- a/callbacks/associations_test.go
+++ b/callbacks/associations_test.go
@@ -1,0 +1,84 @@
+package callbacks
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"gorm.io/gorm"
+	"gorm.io/gorm/schema"
+)
+
+func TestOnConflictOption_NoPrimaryKeys_DoNothing(t *testing.T) {
+	// Model without any primary key fields
+	type NoPKModel struct {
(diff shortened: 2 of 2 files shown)


veeceey:
### Test results

```
$ go test ./...
ok  	gorm.io/gorm         0.542s
ok  	gorm.io/gorm/callbacks  1.757s
ok  	gorm.io/gorm/clause     1.012s
ok  	gorm.io/gorm/logger     1.323s
ok  	gorm.io/gorm/schema     1.858s
ok  	gorm.io/gorm/utils      2.212s
```

All existing tests pass. The fix targets the specific case where an associated model has no declared primary key fields — `onConflictOption` now falls back to `DO NOTHING` instead of generating invalid SQL.

veeceey:
Fair point about the regression test. I'll add a unit test that sets up a schema with no primary key fields and verifies that onConflictOption falls back to DoNothing instead of generating invalid SQL. Will push that shortly.

veeceey:
added the regression test — covers both the no-primary-key path (verifies DoNothing fallback) and the normal path with primary keys. all existing tests still pass locally.

veeceey:
hey, just checking in on this. the CI failures look infrastructure-related (cache restore issues, not code failures). happy to rebase or make any changes if needed!

PR #7716, opened 2026-02-28
fix(utils): resolve incorrect line number reporting in FileWithLineNum

huningzizhitaibai:
The introduction of CallerFrame in v1.31.1 increased the call stack depth, but the hardcoded skip value in runtime.Callers was not adjusted. This caused FileWithLineNum to report its own caller's line number instead of the original business code's line number.

Also added a regression test to verify the fix.

- [x] Do only one thing
- [x] Tested

Fixed a regression in FileWithLineNum introduced in v1.31.1. The new CallerFrame wrapper increased call stack depth, but runtime.Callers skip value was not updated. The skip value has been adjusted to correctly pinpoint the original caller.

### User Case Description
When using v1.31.1 in the local project, FileWithLineNum() reports the line number of the internal FileWithLineNum() call itself instead of the business code line that triggered the GORM operation.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -54,9 +54,17 @@ func CallerFrame() runtime.Frame {
 
 // FileWithLineNum return the file name and line number of the current file
 func FileWithLineNum() string {
-	frame := CallerFrame()
-	if frame.PC != 0 {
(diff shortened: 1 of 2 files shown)


huningzizhitaibai:
Sorry, I push my commit on master branch. I will change it soon.

PR #7717, opened 2026-02-28
fix(utils): restore FileWithLineNum caller depth and add regression test

### LINKED ISSUE #7715: Bug : utils.FileWithLineNum returns inaccurate location after slog CallerFrame change
## GORM Playground Link

https://github.com/go-gorm/playground/pull/845

## Description

As noted in https://github.com/go-gorm/gorm/pull/7610#issuecomment-3877337170, after https://github.com/go-gorm/gorm/pull/7610 was merged, a new regression was introduced.

Specifically, comparing v1.31.1 and v1.31.0 shows that the behavior of `utils.FileWithLineNum()` is inconsistent.

Here is a minimal reproduction:

```go
package main

import (
	"context"
	"fmt"
	"log/slog"
	"os"
	"time"

	"gorm.io/gorm/logger"
	"gorm.io/gorm/utils"
)

func main() {
	demoExternal()
	demoSlog()
}

func where() string {
	return utils.FileWithLineNum() // line 20
}

func demoExternal() {
	actual := where() // line 24
	fmt.Println("external FileWithLineNum:", actual)
}

func demoSlog() {
	handler := slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{AddSource: true})
	l := logger.NewSlogLogger(slog.New(handler), logger.Config{LogLevel: logger.Info})
	l.Trace(context.Background(), time.Now(), func() (string, int64) {
		return "select 1", 0
	}, nil)
}
```

`go.mod` with v1.31.0:

```
module hello

go 1.25.4

require gorm.io/gorm v1.31.0
```

Output:

```
external FileWithLineNum: /Users/nx/GolandProjects/hello/main.go:24
time=2026-02-28T00:43:55.136+08:00 level=INFO source=/Users/nx/go/pkg/mod/gorm.io/gorm@v1.31.0/logger/slog.go:87 msg="SQL executed" trace.duration=0.024ms trace.sql="select 1" trace.rows=0
```

But if we switch to v1.31.1:

```
module hello

go 1.25.4

require gorm.io/gorm v1.31.1
```

Output becomes:

```
external FileWithLineNum: /Users/nx/GolandProjects/hello/main.go:20 # this changed
time=2026-02-28T00:46:05.436+08:00 level=INFO source=/Users/nx/GolandProjects/hello/main.go:31 msg="SQL executed" trace.duration=0.017ms trace.sql="select 1" trace.rows=0
```

Please assign this issue to me; I am preparing a PR.  
This likely requires more than just changing a single number, and I will review related areas for other potential issues.

smf-h:
## Summary
- restore FileWithLineNum() caller resolution behavior that regressed in v1.31.1
- extract stack scan into internal callerFrame(skip int) helper
- keep CallerFrame() API unchanged and use explicit skip handling
- add regression test for wrapped call chain (TestFileWithLineNumWrappedCallRegression)

## Root cause
After #7610, stack scanning was extracted into CallerFrame(). The call path for FileWithLineNum() gained an extra wrapper frame, while fixed caller-depth handling caused wrapper-level frame resolution in v1.31.1.

## Validation
- slog source behavior checked in local matrix probe (Info/Warn/Error/Trace): unchanged before vs after patch.

Fixes #7715

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -36,9 +36,14 @@ func sourceDir(file string) string {
 //   - Exclude test files (*_test.go)
 //   - go-gorm/gen's Generated files (*.gen.go)
 func CallerFrame() runtime.Frame {
+	// Preserve the original CallerFrame stack depth after introducing callerFrame().
+	return callerFrame(4)
+}
+
+func callerFrame(skip int) runtime.Frame {
 	pcs := [13]uintptr{}
-	// the third caller usually from gorm internal
-	len := runtime.Callers(3, pcs[:])
+	// skip is caller-path sensitive and should be selected by each public helper.
+	len := runtime.Callers(skip, pcs[:])
 	frames := runtime.CallersFrames(pcs[:len])
 	for i := 0; i < len; i++ {
 		// second return value is "more", not "ok"
@@ -54,7 +59,8 @@ func CallerFrame() runtime.Frame {
 
 // FileWithLineNum return the file name and line number of the current file
 func FileWithLineNum() string {
-	frame := CallerFrame()
+	// Keep FileWithLineNum one frame outer than CallerFrame to preserve pre-v1.31.1 semantics.
+	frame := callerFrame(4)
 	if frame.PC != 0 {
 		return string(strconv.AppendInt(append([]byte(frame.File), ':'), int64(frame.Line), 10))
 	}
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -4,7 +4,10 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"errors"
+	"fmt"
 	"math"
+	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


qqxhb:
Thanks for the fix!

LGTM on the approach: keeping CallerFrame() semantics stable while making FileWithLineNum() explicitly select its own skip depth via an internal helper makes the call-stack sensitivity much clearer. The wrapped-call regression test is also a good addition and matches the pre-v1.31.1 behavior.

I ran go test ./... locally and it passed.

CI looks mostly green; the only failing job is tests/mysql (mysql:9, 1.24) which seems flaky across several recent PRs. Could you please re-run the failed job / workflow? If it goes green, this should be good to merge.

smf-h:
Thanks for the rerun.

I checked the latest failed job:
https://github.com/go-gorm/gorm/actions/runs/22523502123/job/65501950447

This failure is different from the previous flaky test failure. It fails during dependency resolution, before test execution:

go: github.com/microsoft/go-mssqldb@v1.9.8 requires go >= 1.25.7 (running go 1.24.13)

In CI, tests/tests_all.sh runs `go get -u -t ./...` (tests/tests_all.sh:12), which upgrades dependencies at runtime.
So this looks like dependency drift in CI, not a regression from this PR’s utils changes.

Could a maintainer advise whether to pin deps / avoid `-u` in CI, or run this job with a compatible Go version?

smf-h:
I don’t seem to have permission to rerun workflows on this PR from my side, so I’m unable to trigger the failed job directly.Could you rerun it,please?

PR #7718, opened 2026-03-01
Fix PostgreSQL INCLUDE option parsing for multi-column indexes

### LINKED ISSUE #7707: POSTGRES: Covering index fails when more than one field is provided in INCLUDE statement
## GORM Playground Link
https://github.com/go-gorm/playground/pull/843

## Description

On POSTGRES:
I want to create a covering index on "created_at" field, and add "email" to that index with some include fields like "status" and "friendly" (I've made up the names here).

Gorm fails to create the table in the postgres database. It seems that it fails on creating the right SQL command.

If I only include one field like "status" then Gorm creates the table and the index correctly.

**NOTE**: Why does the go.mod file in the playground codebase try to pull "gorm.io/cmd/gorm" which does not exist?
If i delete this line from the go.mod file then it all runs fine. This issue is showing on GitHub Actions too. There is also another issue with the version of github actions which is preventing my PR tests from running. Is anyone actively working on the playground codebase? I have pushed some changes that make the postgres tests run correctly, you can check the error there for the issue I am facing with covering index on postgres.

pawannn:
- [x] Do only one thing
- [x] Tested

Fixed a bug where PostgreSQL covering indexes with multiple INCLUDE columns were silently truncated due to GORM's tag parser splitting on every comma. Added a paren-aware SplitTagCommas helper and re-extract the OPTION value to correctly handle commas inside parentheses.

### User Case Description

Fixes #7707 

## Summary
When using PostgreSQL covering indexes with multiple `INCLUDE` columns, GORM silently truncates the option value and generates invalid SQL.

**Consider the following test case**
 ```go
 func TestIndexOptionWithMultipleIncludeColumns(t *testing.T) {
	type Item struct {
		gorm.Model
		Status    string
		Friendly  string
		CreatedAt time.Time `gorm:"index:idx_created,option:INCLUDE (status, friendly)"`
	}

	s, err := schema.Parse(&Item{}, &sync.Map{}, schema.NamingStrategy{})
	if err != nil {
		t.Fatalf("failed to parse schema: %v", err)
	}

	indexes := s.ParseIndexes()

	var idx schema.Index

	for _, index := range indexes {
		if index.Name == "idx_created" {
			idx = *index
		}
	}

	want := "INCLUDE (status, friendly)"
	if idx.Option != want {
		t.Errorf("expected Option = %q, got %q", want, idx.Option)
	}
}
 ```

**Output**:
Screenshot 2026-03-01 at 12 55 07 PM

## Root Cause
In `parseFieldIndexes`, the tag is split on every comma to extract settings:
```go
tagSetting = strings.Join(strings.Split(tag, ",")[1:], ",")
settings   = ParseTagSetting(tagSetting, ",")
```

For a tag like:
```go
`gorm:"index:idx_created,option:INCLUDE (status, friendly)"`
```

The `strings.Split` splits on **every** comma including the one inside `INCLUDE (status, friendly)` — so `settings["OPTION"]` becomes `INCLUDE (status` instead of `INCLUDE (status, friendly)`, producing broken SQL on AutoMigrate.

This works fine with a single column (`INCLUDE (status)`) because there is no comma, which is exactly why the bug is hard to spot.

## Fix

**1. Added `SplitTagCommas()` helper** in `schema/utils.go`:**

Splits by commas while ignoring commas inside parentheses:
```go
func SplitTagCommas(s string) []string {
    // ...
    // "idx_created,option:INCLUDE (status, friendly)"
    // → ["idx_created", "option:INCLUDE (status, friendly)"]
}
```

**2. Use `SplitTagCommas` in `parseFieldIndexes`** and re-extract `OPTION` value directly from the paren-aware split:

```go
// After
tagParts   = SplitTagCommas(tag)
tagSetting = strings.Join(tagParts[1:], ",")
settings   = ParseTagSetting(tagSetting, ",")

// Re-extract OPTION directly from tagParts to avoid truncation
for _, part := range tagParts[1:] {
    part = strings.TrimSpace(part)
    if strings.HasPrefix(strings.ToUpper(part), "OPTION:") {
        settings["OPTION"] = part[len("OPTION:"):]
        break
    }
}
```

> **Note:** `ParseTagSetting` is intentionally left unchanged. It is used broadly across GORM (field tags, relationships,
> constraints) and changing it would risk breaking other tag parsing. The OPTION value is re-extracted surgically after the fact.

### After Fix
Screenshot 2026-03-01 at 2 02 58 PM

### DIFF
--- a/schema/index.go
+++ b/schema/index.go
@@ -106,7 +106,7 @@ func parseFieldIndexes(field *Field) (indexes []Index, err error) {
 					name       string
 					tag        = strings.Join(v[1:], ":")
 					idx        = strings.IndexByte(tag, ',')
-					tagParts   = SplitTagCommas(tag)
+					tagParts   = splitTagCommas(tag)
 					tagSetting = strings.Join(tagParts[1:], ",")
 					settings   = ParseTagSetting(tagSetting, ",")
 					length, _  = strconv.Atoi(settings["LENGTH"])
--- a/schema/index.go
+++ b/schema/index.go
@@ -106,10 +106,20 @@ func parseFieldIndexes(field *Field) (indexes []Index, err error) {
 					name       string
 					tag        = strings.Join(v[1:], ":")
 					idx        = strings.IndexByte(tag, ',')
-					tagSetting = strings.Join(strings.Split(tag, ",")[1:], ",")
+					tagParts   = SplitTagCommas(tag)
+					tagSetting = strings.Join(tagParts[1:], ",")
 					settings   = ParseTagSetting(tagSetting, ",")
 					length, _  = strconv.Atoi(settings["LENGTH"])
 				)
+
+				for _, part := range tagParts[1:] {
+					part = strings.TrimSpace(part)
+					upper := strings.ToUpper(part)
+					if strings.HasPrefix(upper, "OPTION:") {
+						settings["OPTION"] = part[len("OPTION:"):]
+						break
+					}
+				}
 
 				if idx == -1 {
 					idx = len(tag)
--- a/schema/index_test.go
+++ b/schema/index_test.go
@@ -3,7 +3,9 @@ package schema_test
 import (
 	"sync"
 	"testing"
+	"time"
 
+	"gorm.io/gorm"
 	"gorm.io/gorm/schema"
 	"gorm.io/gorm/utils/tests"
 )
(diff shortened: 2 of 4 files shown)


qqxhb:
LGTM overall — thanks for the fix and the regression test.

One optional cleanup suggestion: since parseFieldIndexes already has a paren-aware split (tagParts), you could parse the settings directly from tagParts[1:] (each already representing a full k:v segment) instead of joining back to a string and running ParseTagSetting(..., ","). That would naturally preserve values containing commas (like OPTION) without needing a second pass to re-extract OPTION.

pawannn:
Hey @qqxhb, 
I Tried this approach but it reintroduces complexity around the existing `\\`, escape convention. I've kept the two-pass approach for avoiding regressing that behavior.

pawannn:
These CI failures appear to be pre-existing environment issues and are unrelated to the changes in this PR.

Issue #7403, opened 2025-03-28
A race condition which may lead to a dead lock inside ScanRows and causes DB locks which are never closed.


swojtasiak:
I don't include the playground link because it's a race condition which is really hard to reproduce. In our case it happens once a few days, while we issue thousands of queries per seconds to our databases.

The issue shouldn't be difficult to understand taking into account the fact that the SQL API provided by Go is used in a wrong way in the scanning.

The `rows.Err()` method called here: 

https://github.com/go-gorm/gorm/blob/a9d27293de2267a36fa6c9f8892977d3159cf8ea/scan.go#L353

Can cause a dead lock. This method should be called only and only if we are sure that the underlying `sql.Rows` instance is closed. Unfortunately it's not always the case here, as it can be called just after the scan operation is done:

https://github.com/swojtasiak/gorm/blob/a9d27293de2267a36fa6c9f8892977d3159cf8ea/scan.go#L346

As the documentation says, every call to `rows.Err()` should be protected by a call to `rows.Next()` and only if it returns false, we can be sure that it's safe to call the `rows.Err()` method.

It's not safe, because every call to `rows.Scan()` may leave the rows instance in a read locked state, by acquiring a read lock on an internal RWMutex. The lock is held until `rows.Close()` or `rows.Next()` is called. 

When `rows.Err()` is called it tries to acquire the read lock again before accessing the error instance. In most cases it's not an issue, because the lock is being acquired only for a moment and it's allowed to acquire it more than once for read operations. It just increments its internal number of calls. 

The problem arises when a call to `rows.Scan()` leaves the rows instance in a locked state and before `rows.Err()` is called, another goroutine tries to close the rows instance due to a context cancellation, like here:

```
goroutine 5258713 [sync.RWMutex.Lock, 2190 minutes]:
sync.runtime_SemacquireRWMutex(0x10?, 0x20?, 0x4011aa9e01?)
	/usr/local/go/src/runtime/sema.go:87 +0x28
sync.(*RWMutex).Lock(0x4011aa9fa8?)
	/usr/local/go/src/sync/rwmutex.go:151 +0xf8
database/sql.(*Rows).close(0x4010285950, {0x268ca00, 0x3a30180})
	/usr/local/go/src/database/sql/sql.go:3399 +0x88
database/sql.(*Rows).awaitDone(0x4010285950, {0x26a8508, 0x4016e1a000}, {0x26a8540, 0x401327f590}, {0x26a8540, 0x4013553d60})
[...]

PR #7720, opened 2026-03-03
fix: remove rows.Err() from Scan to prevent deadlock (#7403)

DukeDeSouth:
Scan() calls rows.Err() after scanning, which can deadlock when a context
cancellation goroutine (awaitDone) tries to WLock the internal closemu
RWMutex concurrently. Since rows.Err() needs an RLock, it blocks behind
the pending WLock — permanent deadlock, leaked DB connection.

This moves the rows.Err() responsibility to the 4 callback callers that
actually own the rows lifecycle. Each now calls `!rows.Next()` before
`rows.Err()`, ensuring all locks are released first. ScanRows (public API)
is unchanged — callers manage their own lifecycle per standard Go patterns.

Same logical fix as #7404, rebased cleanly on current HEAD.

- [x] Do only one thing
- [x] Tested

Made with [Cursor](https://cursor.com)

### DIFF
--- a/callbacks/create.go
+++ b/callbacks/create.go
@@ -93,6 +93,9 @@ func Create(config *Config) func(db *gorm.DB) {
 			)
 			if db.AddError(err) == nil {
 				defer func() {
+					if !rows.Next() {
+						db.AddError(rows.Err())
+					}
 					db.AddError(rows.Close())
 				}()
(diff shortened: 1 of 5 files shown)


jinzhu:
why it cause deadlock? can you add some tests?

PR #7726, opened 2026-03-09
ci: switch tests Go matrix to stable/oldstable and update setup-go

smf-h:
This updates the tests workflow Go matrix from `['1.24', '1.25']` to `['1.25', 'stable']`.

Reason:
`tests/tests_all.sh` runs `go get -u -t ./...` inside `tests/`, so CI can pick up newer transitive dependencies than the versions currently recorded in the repo.

At the moment, this path is no longer compatible with Go 1.24. I reproduced it locally with the current script:

- `github.com/microsoft/go-mssqldb@v1.9.8 requires go >= 1.25.7`
- `golang.org/x/sync@v0.20.0 requires go >= 1.25.0`

So the failing Go 1.24 jobs are blocked by dependency/toolchain requirements during test environment setup, before the actual repository changes are meaningfully exercised.

This PR keeps the change minimal and only updates the CI Go matrix.
It does not change test logic or application code.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,13 +16,13 @@ jobs:
   sqlite:
     strategy:
       matrix:
-        go: ['1.25', 'stable']
+        go: ['stable', 'oldstable']
         platform: [ubuntu-latest] # can not run in windows OS
     runs-on: ${{ matrix.platform }}
 
     steps:
     - name: Set up Go 1.x
-      uses: actions/setup-go@v4
+      uses: actions/setup-go@v5
       with:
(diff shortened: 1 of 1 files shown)


jinzhu:
Hi @smf-h,

Could you update this to use stable and oldstable as described in the documentation?

It may also require upgrading the current setup-go version.

https://github.com/actions/setup-go?tab=readme-ov-file#supported-version-syntax

jinzhu:
can you fix the confliction？

smf-h:
@jinzhu  I have fixed the confliction.

Issue #7709, opened 2026-02-21
Generics API: Updates(ctx, map[string]interface{}) fails with type mismatch (cannot use map[string]interface{} as User)


immortal521:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/1

## Description

On the GORM documentation page (https://gorm.io/docs/update.html#Updates-multiple-columns
), it says the Generics API can update multiple columns using:
```go
rows, err := gorm.G[User](db).Where("id = ?", 111).Updates(ctx, map[string]interface{}{"name": "hello", "age": 18, "active": false})
```
However, this results in an error:
```
cannot use map[string]interface{}{} (value of type map[string]interface{}) as User value in argument to gorm.G[User](db).Where("id = ?", 111).Updates [IncompatibleAssign]
```

cxlblm:
Yes, this is indeed an issue. I have encountered the same problem as well.

cxlblm:
this link https://github.com/go-gorm/playground/pull/847

jinzhu:
Sorry for the incorrect document — I’ve just updated it. You should use a map like this
```go
result, err := gorm.G[map[string]interface{}](db).
    Table("users").
    Where("id = ?", 111).
    Updates(ctx, map[string]interface{}{
        "name":   "hello",
        "age":    18,
        "active": false,
    })
// UPDATE users SET name='hello', age=18, active=false, updated_at='2013-11-17 21:34:10' WHERE id=111;
```

PR #7734, opened 2026-03-16
fix: generics updates arg use map

philhuan:
- [x] Do only one thing
- [x] Tested

This pull request fixes issue #7709.

It relaxes the parameter type of the Updates method in the generic API gorm.G[T] to any. This change aligns the generic method's behavior with the polymorphic nature of gorm.DB.Updates, enabling it to correctly handle updates from both map[string]interface{} and struct types. Previously, the strict type constraint limited the method's usability in dynamic field update scenarios.

The implementation involves modifying the Updates method signature in the CreateInterface[T] and ChainInterface[T] interfaces. The core logic in chainG[T].Updates is adjusted to first set the model with Model(r) (where r is of type *T) and then apply the updates. This ensures the correct table schema is targeted before executing the update operation with the provided values.

```
// In CreateInterface[T] and ChainInterface[T]
Updates(ctx context.Context, values any) (rowsAffected int, err error)

// In chainG[T]
func (c chainG[T]) Updates(ctx context.Context, values any) (rowsAffected int, err error) {
    var r *T
    res := c.g.apply(ctx).Model(r).Updates(values)
    return int(res.RowsAffected), res.Error
}

```

Verification was completed by adding new tests in tests/generics_test.go, which cover both map and struct update paths. This change is non-breaking and maintains the behavior of existing calls while enhancing functionality for dynamic updates.

#### Example 1: Updating with a map

update specific fields by passing a map[string]interface{}. 

```
// User wants to update only name and age
updates := map[string]interface{}{"name": "hello", "age": 18}
gorm.G[User](DB).Where("id = ?", user.ID).Updates(context.Background(), updates)
```

#### Example 2: Updating with a struct
The existing behavior of updating from a struct instance remains fully supported. 

```
// User updates the name and saves the object
user.Name = "world"
gorm.G[User](DB).Where("id = ?", user.ID).Updates(context.Background(), user)

```

### DIFF
--- a/generics.go
+++ b/generics.go
@@ -59,7 +59,7 @@ type CreateInterface[T any] interface {
 
 	Delete(ctx context.Context) (rowsAffected int, err error)
 	Update(ctx context.Context, name string, value any) (rowsAffected int, err error)
-	Updates(ctx context.Context, t T) (rowsAffected int, err error)
+	Updates(ctx context.Context, t any) (rowsAffected int, err error)
 	Count(ctx context.Context, column string) (result int64, err error)
 
 	Table(name string, args ...interface{}) CreateInterface[T]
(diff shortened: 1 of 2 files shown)


jinzhu:
Sorry for the earlier mistake — I’ve just updated the document. You should use a map like this. We don’t plan to introduce any interfaces for the generics API. Thank you for your understanding.

```go
result, err := gorm.G[map[string]interface{}](db).
    Table("users").
    Where("id = ?", 111).
    Updates(ctx, map[string]interface{}{
        "name":   "hello",
        "age":    18,
        "active": false,
    })
// UPDATE users SET name='hello', age=18, active=false, updated_at='2013-11-17 21:34:10' WHERE id=111;
```

PR #7760, opened 2026-04-25
fix(generics): align Select signature with DB.Select

MichaelSatoshi1983:
- [ - ] Do only one thing
- [ - ] Non breaking API changes
- [ - ] Tested

Changed the Generics API Select signature from string to any so it matches DB.Select. This allows passing []string directly and fixes incorrect SQL generation when using dynamic field lists.

### User Case Description

This fixes issue #7743 When the selected columns are built dynamically as []string, the previous Generics API required using Select("?", fields), which could generate incorrect SQL in PostgreSQL by treating column names as string literals. After this change, callers can use gorm.G[User](db).Select(fields) directly, consistent with the underlying API.


MichaelSatoshi1983:
@jinzhu Please review,thanks

PR #7770, opened 2026-05-12
fix(migrator): normalize quoted default values in non-string type comparison

### LINKED ISSUE #7553: AutoMigration always attempts to set the default value, even when it hasn’t changed.
## GORM Playground Link

https://github.com/go-gorm/playground/pull/823

## Description

We have a table definition goes like the following:
```
type Payload struct {
	ID string
}
type Table struct {
	ID      string  `gorm:"primaryKey;type:uuid"`
	Payload Payload `gorm:"serializer:json;type:jsonb;default:'{}'"`
}
```

We noticed that every time we run the auto migration, GORM always executes the SQL `ALTER TABLE <table> ALTER COLUMN payload SET DEFAULT '{}'` to set the default value, even though it hasn’t changed.

After some investigation, I found that GORM interprets the default value as `'{}'` when parsing the struct, but retrieves it as `{}` from the online database. This mismatch causes GORM to attempt to update the default value every time, even when it hasn’t changed.

lihan3238:
Fixes #7553

## Summary

Fix default value comparison in MigrateColumn for non-String GORMDataTypes (e.g. custom types like jsonb) that don't strip surrounding quotes before comparison.

When a field has both a custom type tag (type:jsonb) and a serializer tag (serializer:json), the GORMDataType becomes the custom type name (jsonb) instead of String. This causes the default value comparison to fall into the default case, which compares values directly without normalizing quotes.

For example, with default:'{}' on a jsonb field:
- field.DefaultValue = '{}' (literal single quotes from struct tag)
- Database returns dv = '{}' (without quotes)
- Direct comparison detects mismatch, causing unnecessary ALTER TABLE SET DEFAULT on every AutoMigrate

The fix applies the same quote-stripping normalization used by the String case to the default case.

## Test plan

- go build ./... passes
- go test ./... -short passes (existing TestSlogLogger failure is pre-existing, unrelated)
- Verify with reproduction case from #7553

### DIFF
--- a/migrator/migrator.go
+++ b/migrator/migrator.go
@@ -568,7 +568,7 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 					alterColumn = true
 				}
 			default:
-				if dv != field.DefaultValue {
+				if dv != field.DefaultValue && dv != strings.Trim(field.DefaultValue, "'\"") {
 					alterColumn = true
 				}
 			}


migrator/migrator.go
```diff
@@ -568,7 +568,7 @@ func (m Migrator) MigrateColumn(value interface{}, field *schema.Field, columnTy
 					alterColumn = true
 				}
 			default:
-				if dv != field.DefaultValue {
+				if dv != field.DefaultValue && dv != strings.Trim(field.DefaultValue, "'\"") {
 					alterColumn = true
```
dshoreman, line 572: If a `CREATE` is run manually with quoted defaults, the parser returns `dv` with quotes too. Would it break anything if both sides were trimmed and the strings case was handled by the default again?

The suggestion bugged out and missed the first two lines but something like this:

```diff
- case schema.String:
- if dv != field.DefaultValue && dv != strings.Trim(field.DefaultValue, "'\"") {
- 		alterColumn = true
- }
  default:
- 		if dv != field.DefaultValue && dv != strings.Trim(field.DefaultValue, "'\"") {
+  		if strings.Trim(dv, "'\"") != strings.Trim(field.DefaultValue, "'\"") {
				alterColumn = true
```

PR #7798, opened 2026-05-30
Fix potential rows leak on panic by deferring rows.Close()

### LINKED ISSUE #7790: Bug: rows.Close() not deferred in delete/update returning paths — connection leak on panic
## Describe the feature

Fix resource leaks where `*sql.Rows` are not deferred in the delete and update returning paths, risking connection pool exhaustion if `gorm.Scan` panics.

## Motivation

`callbacks/query.go` and `callbacks/create.go` correctly use `defer rows.Close()`. However, `callbacks/delete.go:173-179` and `callbacks/update.go:89-99` call `rows.Close()` inline without defer. If `gorm.Scan` panics (possible given heavy reflect usage), `rows.Close()` is never called, leaking the underlying database connection. Under panic conditions, this can exhaust the connection pool.

### `callbacks/delete.go:173-179`
```go
if rows, err := db.Statement.ConnPool.QueryContext(...); db.AddError(err) == nil {
    gorm.Scan(rows, db, mode)
    // ...
    db.AddError(rows.Close())  // NOT deferred
}
```

### `callbacks/update.go:89-99`
```go
if ok, mode := hasReturning(db, supportReturning); ok {
    if rows, err := db.Statement.ConnPool.QueryContext(...); db.AddError(err) == nil {
        // no defer rows.Close() here!
        gorm.Scan(rows, db, mode)
        db.AddError(rows.Close())  // NOT deferred
    }
}
```

### Expected pattern (from `callbacks/query.go:24-26`)
```go
if rows, err := db.Statement.ConnPool.QueryContext(...); db.AddError(err) == nil {
    defer func() { db.AddError(rows.Close()) }()
    gorm.Scan(rows, db, mode)
}
```

## Related Issues

- Performance architecture review of GORM codebase

omer-cengel:
Fixes https://github.com/go-gorm/gorm/issues/7790

- [X] Do only one thing
- [X] Tested

## Description

This PR updates `rows.Close()` handling to use `defer` in code paths where `*sql.Rows` may remain unclosed if a panic occurs.

* Defer `rows.Close()` in `callbacks/delete.go`
* Defer `rows.Close()` in `callbacks/update.go`
* Defer `rows.Close()` in `finisher_api.go`
* Use `defer rows.Close()` in `tests/query_test.go`

This aligns the behavior with the existing pattern already used in `callbacks/query.go` and `callbacks/create.go`.

### User Case Description

Previously, `rows.Close()` was called inline after `gorm.Scan(...)` / `ScanRows(...)`. In the event of a panic while scanning rows, `rows.Close()` could be skipped, potentially leaving database connections open.

Using `defer` ensures cleanup happens even during panic unwinding while keeping behavior unchanged during normal execution.

### Testing

* Run existing test suite
* No functional behavior changed

### DIFF
--- a/callbacks/delete.go
+++ b/callbacks/delete.go
@@ -172,7 +172,9 @@ func Delete(config *Config) func(db *gorm.DB) {
 
 			if rows, err := db.Statement.ConnPool.QueryContext(db.Statement.Context, db.Statement.SQL.String(), db.Statement.Vars...); db.AddError(err) == nil {
 				defer func() {
-					db.AddError(rows.Close())
+					if err := rows.Close(); err != nil {
+						_ = db.AddError(err)
+					}
 				}()
 
 				gorm.Scan(rows, db, mode)
--- a/callbacks/delete.go
+++ b/callbacks/delete.go
@@ -171,12 +171,15 @@ func Delete(config *Config) func(db *gorm.DB) {
 			}
 
 			if rows, err := db.Statement.ConnPool.QueryContext(db.Statement.Context, db.Statement.SQL.String(), db.Statement.Vars...); db.AddError(err) == nil {
+				defer func() {
+					db.AddError(rows.Close())
+				}()
+
 				gorm.Scan(rows, db, mode)
 
 				if db.Statement.Result != nil {
 					db.Statement.Result.RowsAffected = db.RowsAffected
 				}
-				db.AddError(rows.Close())
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


pageton (approved):
## Review

**LGTM** — clean, minimal fix. The `defer rows.Close()` change is correct and aligns with existing patterns in `callbacks/query.go` and `callbacks/create.go`.

### Verified

- All relevant tests pass (`TestDeleteReturning`, `TestUpdateReturning`, `TestSelectWithVariables`, `TestScanRows`)
- No remaining unfixed instances — `callbacks/create.go`, `callbacks/query.go`, and `migrator/migrator.go` already use `defer`

### Two minor suggestions (non-blocking)

1. **Style inconsistency**: The PR uses `if err := rows.Close(); err != nil { _ = db.AddError(err) }` while existing code in `query.go:25` and `create.go:96` uses `db.AddError(rows.Close())`. Both are correct, but unifying the style across all 4 call sites would reduce noise.

2. **Regression test**: A test that verifies `rows.Close()` is called even when `gorm.Scan` panics would be valuable to prevent future regressions. Hard to test without a mock `ConnPool`, so this is a nice-to-have.

Mergeable as-is.

Issue #7786, opened 2026-05-21
Performance: eliminate hot-path overhead when logger is Silent (Trace closure, time.Now, context leak)


pageton:
## Describe the feature

Eliminate allocation and syscall overhead in the hot callback execution path when logging is disabled or silent.

### 1. `Logger.Trace` called on every query even when Silent — `callbacks.go:139-147`

```go
if stmt.SQL.Len() > 0 {
    db.Logger.Trace(stmt.Context, curTime, func() (string, int64) {
        sql, vars := stmt.SQL.String(), stmt.Vars
        if filter, ok := db.Logger.(ParamsFilter); ok {
            sql, vars = filter.ParamsFilter(stmt.Context, stmt.SQL.String(), stmt.Vars...)
        }
        return db.Dialector.Explain(sql, vars...), db.RowsAffected
    }, db.Error)
}
```

The `Trace` method is called on **every single query**, even when the log level is `Silent`. While `logger.Trace` (`logger/logger.go:161-164`) returns early for `Silent`, the closure is always allocated and `stmt.SQL.String()` + `stmt.Vars` are always captured. `db.Dialector.Explain()` does expensive string interpolation (`logger/sql.go:51-180`) for every query even when discarded.

**Fix:** Add a `TraceEnabled()` or expose `LogLevel()` on the logger interface so the caller can short-circuit before allocating the closure. Or check level in `Execute` before calling `Trace`.

### 2. `time.Now()` syscall unconditional in `Execute` — `callbacks.go:82`

```go
curTime := time.Now()
```

`time.Now()` involves a `clock_gettime` syscall. Called on every `Execute`, even when logging is Silent and the time is never used.

**Fix:** Defer `time.Now()` to only when actually needed (inside the Trace closure or after checking logger level).

### 3. Leaked `context.WithTimeout` cancel func — `callbacks.go:96-99`

```go
if db.DefaultContextTimeout > 0 {
    if _, ok := stmt.Context.Deadline(); !ok {
        stmt.Context, _ = context.WithTimeout(stmt.Context, db.DefaultContextTimeout)
    }
}
```

The `CancelFunc` is discarded (`_`). This leaks the context's internal timer goroutine until it expires. Called on every query when `DefaultContextTimeout` is set.

**Fix:** Store the cancel func and call it after the callback pipeline completes (e.g., in a deferred cleanup at the end of `Execute`).

## Estimated Impact

| Fix | Allocs/Syscalls Saved | Per-query |
[...]

PR #7809, opened 2026-06-08
fix: defer context cancel func to prevent goroutine leak

123456wda:
## Problem

`context.WithTimeout` returns a `CancelFunc` that must be called to release the associated timer goroutine. In `callbacks.go` `Execute()` (line 97), the `CancelFunc` was discarded with `_`, causing the timer goroutine to leak until the timeout expires on every query when `DefaultContextTimeout` is configured.

## Changes

**callbacks.go** `Execute()` (line 98):
```go
// Before
stmt.Context, _ = context.WithTimeout(stmt.Context, db.DefaultContextTimeout)

// After
var cancel context.CancelFunc
stmt.Context, cancel = context.WithTimeout(stmt.Context, db.DefaultContextTimeout)
defer cancel()
```

This is safe because the callback runs synchronously within `Execute()` — the `cancel` fires only after the operation completes.

## Note on `finisher_api.go` `Begin()`

An earlier revision of this PR also added `defer cancel()` inside `Begin()`. That was incorrect and has been reverted (thanks @viacheslav-danilin for [catching it](https://github.com/go-gorm/gorm/pull/7809#issuecomment-4674263951)): the returned `tx` outlives `Begin()`, so deferring at function scope would cancel the context immediately and break every subsequent operation on the transaction.

Properly fixing the `Begin()` path requires tying `cancel` to the transaction's lifetime (calling it from `Commit()` / `Rollback()`), which is a larger change and feels out of scope here. Happy to do it as a follow-up if maintainers want.

## Testing

All existing tests pass (`go test ./...`).

Related: #7786

### DIFF
--- a/callbacks.go
+++ b/callbacks.go
@@ -95,7 +95,9 @@ func (p *processor) Execute(db *DB) *DB {
 
 	if db.DefaultContextTimeout > 0 {
 		if _, ok := stmt.Context.Deadline(); !ok {
-			stmt.Context, _ = context.WithTimeout(stmt.Context, db.DefaultContextTimeout)
+			var cancel context.CancelFunc
+			stmt.Context, cancel = context.WithTimeout(stmt.Context, db.DefaultContextTimeout)
+			defer cancel()
 		}
 	}
 


viacheslav-danilin:
Should we really call defer cancel() here? This may cancel the transaction context immediately after Begin() returns

123456wda:
> Should we really call defer cancel() here? This may cancel the transaction context immediately after Begin() returns

@viacheslav-danilin You're right, thanks for the catch. Deferring cancel() inside Begin() would cancel the context
 as soon as Begin() returns and break every subsequent operation on the returned tx.
  
 I've reverted that hunk and kept only the callbacks.go fix, where the cancel is safe because the callback runs
 synchronously within Execute(). Properly tying the cancel to a transaction's lifetime would need plumbing through
 Commit()/Rollback(), which feels out of scope for this PR — happy to do it as a follow-up if maintainers want it.

System-Glitch:
I think this is still an issue that could cause context leaks.

Long ago I wrote a plugin to handle this but having the `DefaultContextTimeout` baked in Gorm directly is so much simpler and better than hacking callbacks like I did.

I am reluctant to use it though because of this issue. Any plan for a fix? Could we simply add a `cancelFunc` field in the statement so we can call it at the most appropriate time?

System-Glitch:
> @viacheslav-danilin You're right, thanks for the catch. Deferring cancel() inside Begin() would cancel the context as soon as Begin() returns and break every subsequent operation on the returned tx.
> 
> I've reverted that hunk and kept only the callbacks.go fix, where the cancel is safe because the callback runs synchronously within Execute(). Properly tying the cancel to a transaction's lifetime would need plumbing through Commit()/Rollback(), which feels out of scope for this PR — happy to do it as a follow-up if maintainers want it.

I think the only missing piece here is to clone the statement before assigning the new timeout context. This way the original statement isn't altered and can be re-used. Just like `db.WithContext()` works actually, it clones with a new statement.

Each statement/query have its own timeout context with this system. So cancelling the `BEGIN` context after it returns would be perfectly fine.

Edit: one small problem with this approach on the `Scan` method: the context would be cancelled before `rows.Next()` because the callback processor returns before `rows.Next` and `tx.ScanRows`. For this case I would advise moving the row scanning to its own callback just after `gorm:row`. I think it would be pretty easy and would solve the issue entirely!

PR #7837, opened 2026-08-07
perf: reduce scan allocations

### LINKED ISSUE #7789: Performance: reduce per-query allocations with sync.Pool for Statement and scan buffers
## Describe the feature

Use `sync.Pool` for `Statement` objects and scan value slices to reduce per-query heap allocations by an estimated ~60-70%.

### 1. `Statement` allocated fresh on every query — `gorm.go:450-457`

`getInstance()` is called on every GORM operation. It creates:
```go
tx.Statement = &Statement{
    DB:       tx,
    ConnPool: db.Statement.ConnPool,
    Context:  db.Statement.Context,
    Clauses:  map[string]clause.Clause{},
    Vars:     make([]interface{}, 0, 8),
    // ...
}
```

This is **3-5 heap allocations per query** (struct, map, slice backing array). At 10k QPS that's 30k-50k allocations/sec.

**Fix:** Add `sync.Pool` with a `Reset()` method that clears maps/slices without deallocating. Return to pool after `Execute` completes.

### 2. Scan values slice allocated per scan — `scan.go:129`

```go
values = make([]interface{}, len(columns))
```

For a 10-column table, 160 bytes allocated per scan. Thousands of allocations/sec under load.

**Fix:** Pool these slices keyed by column count, or reuse a per-Statement scratch buffer.

### 3. `map[string]interface{}` should be `map[string]struct{}` — `scan.go:69`

```go
joinedNestedSchemaMap := make(map[string]interface{})
```

Only ever stores `nil` (line 98). Using `interface{}` forces boxing for every nil value. Also allocated per row for join queries instead of being reused across rows within the same scan.

**Fix:** Use `map[string]struct{}`, and reuse the map across rows by clearing instead of reallocating.

### 4. `Statement.clone()` creates maps without capacity — `statement.go:537,542`

```go
Clauses:  map[string]clause.Clause{},
Preloads: map[string][]interface{}{},
```

Then immediately copies all entries. Empty map triggers multiple rehashes during copy.

**Fix:** `make(map[string]clause.Clause, len(stmt.Clauses))`.

## Estimated Impact

| Fix | Allocs Reduced | Mechanism |
|-----|---------------|-----------|
| Pool `Statement` objects | ~60% of per-query allocs | `sync.Pool` + `Reset()` |
| Pool scan value slices | ~10% | `sync.Pool` keyed by column count |
| `map[string]struct{}` + reuse | ~5% for join queries | Eliminate nil boxing + realloc |
| Preallocate clone maps | ~3% | Avoid rehash |

## Related Issues

- Performance architecture review of GORM codebase

adexaja:
## Summary
- reuse scan value buffers with sync.Pool
- avoid interface-valued join tracking map entries
- pre-size cloned statement maps

Closes #7789

## Validation
- `go test ./... -count=1` (root module)


adexaja:
Wait i think i should include the benchmark testing result. Still error on that. working on it.

Issue #7831, opened 2026-07-30
Cached prepared statement closed by eviction between lookup and execution


sazary:
## GORM Playground Link

https://github.com/go-gorm/playground/pull/856

## Description

With `gorm.Config{PrepareStmt: true}` and a bounded cache (`PrepareStmtMaxSize`, or just `PrepareStmtTTL` expiry), live requests intermittently fail with `sql: statement is closed`.

[`prepare`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/prepare_stmt.go#L66-L86) has released every cache lock by the time it returns, so the execution on the next line runs unprotected — [`prepare_stmt.go#L109-L118`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/prepare_stmt.go#L109-L118):

```go
func (db *PreparedStmtDB) ExecContext(ctx context.Context, query string, args ...interface{}) (result sql.Result, err error) {
	stmt, err := db.prepare(ctx, db.ConnPool, false, query)
	if err == nil {
		result, err = stmt.ExecContext(ctx, args...)
		if errors.Is(err, driver.ErrBadConn) {
			db.Stmts.Delete(query)
		}
	}
	return result, err
}
```

In that window another caller can evict the entry — by size in [`LRU.Add`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/internal/lru/lru.go#L138-L142) or by the TTL cleanup goroutine in [`deleteExpired`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/internal/lru/lru.go#L324-L339).
Both reach [`removeElement`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/internal/lru/lru.go#L313-L320):

```go
func (c *LRU[K, V]) removeElement(e *Entry[K, V]) {
	c.evictList.Remove(e)
	delete(c.items, e.Key)
	c.removeFromBucket(e)
	if c.onEvict != nil {
		c.onEvict(e.Key, e.Value)
	}
}
```

[...]

PR #7839, opened 2026-08-08
fix(lru): evict expired entries when they are read

### LINKED ISSUE #7832: Prepared statement leak: expired cache entry overwritten without Close
## GORM Playground Link

https://github.com/go-gorm/playground/pull/855

## Description

A long-lived service using `gorm.Config{PrepareStmt: true}` (and therefore a `PrepareStmtTTL`, default 24h) slowly accumulates `*sql.Stmt` values that are never closed. The count is not bounded by `PrepareStmtMaxSize`, and the statements are released only when the whole `*sql.DB` is closed — so on PostgreSQL/MySQL the server-side prepared statements stay allocated too. It leaks one statement per TTL cycle per hot query.

[`LRU.Get`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/internal/lru/lru.go#L147-L160) treats an expired entry as a miss but leaves it in `c.items`:

```go
	if ent, ok = c.items[key]; ok {
		// Expired item check
		if time.Now().After(ent.ExpiresAt) {
			return value, false
		}
```

Because it is a miss, [`PreparedStmtDB.prepare`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/prepare_stmt.go#L66-L86) falls through to [`Stmts.New`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/internal/stmt_store/stmt_store.go#L158-L168), which caches the new statement via `Set` → [`LRU.Add`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/internal/lru/lru.go#L118-L131). `Add` finds the key still present and takes the existing-item branch, which overwrites the value without calling `onEvict`:

```go
	// Check for existing item
	if ent, ok := c.items[key]; ok {
		c.evictList.MoveToFront(ent)
		c.removeFromBucket(ent) // remove the entry from its current bucket as expiresAt is renewed
		ent.Value = value
		ent.ExpiresAt = now.Add(c.ttl)
		c.addToBucket(ent)
		return false
	}
```

`onEvict` is the only place cached statements are ever closed — [`internal/stmt_store/stmt_store.go#L108-L112`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/internal/stmt_store/stmt_store.go#L108-L112):

```go
	onEvicted := func(k string, v *Stmt) {
		if v != nil {
			go v.Close()
		}
	}
```

After the overwrite nothing in the cache references the previous `*sql.Stmt`, so no later cleanup pass can close it. The window is the gap between an entry expiring and the background [`deleteExpired`](https://github.com/go-gorm/gorm/blob/1d6ce99528060be18a42be09aca8d39efcb47f28/internal/lru/lru.go#L324-L339) goroutine collecting its bucket — exactly when a hot query gets re-run.

### Reproduction

The playground test prepares `SELECT 1`, then `SELECT 2` half a cleanup tick later so both land in the same expiration bucket. Since `deleteExpired` waits for the newest entry in a bucket before collecting it, that second statement delays collection and opens the window; re-running `SELECT 1` inside it takes the overwrite branch.

After the bucket has been collected, a counting `database/sql` driver has seen **3 prepares and 1 close** (`SELECT 2`'s). The first `SELECT 1` statement is still open, with nothing left that could close it. (`database/sql` exposes no open-statement count, which is why the test needs its own driver.)

### Versions

- **GORM:** master @ [`1d6ce99`](https://github.com/go-gorm/gorm/commit/1d6ce99528060be18a42be09aca8d39efcb47f28) (= v1.31.2)
- **Go:** 1.25.12 linux/amd64 (playground CI, ubuntu-latest); also reproduced on go1.26.5 darwin/arm64
- **Driver:** `gorm.io/driver/sqlite` v1.6.0 / `mattn/go-sqlite3` v1.14.49 — the affected code is dialect-independent, so it reproduces on every driver

### Related

#7831 is a use-after-close race in the same cache. Both are symptoms of statement lifetime being tied to cache-slot lifetime rather than to whether anyone is still using the statement, so a fix for either should account for the other.

alliasgher:
Fixes #7832

### Problem

`LRU.Get` treats an expired entry as a cache miss but leaves it in `c.items`:

```go
if ent, ok = c.items[key]; ok {
    if time.Now().After(ent.ExpiresAt) {
        return value, false   // miss, but the entry stays in the map
    }
```

The caller sees a miss and adds a replacement for the same key. `Add` finds the key still present, takes its already-present branch, and overwrites `ent.Value` — returning without ever calling `c.onEvict`.

For the prepared statement store that callback is the only thing that closes a statement (`internal/stmt_store/stmt_store.go`, `onEvicted` → `go v.Close()`). Once the value is overwritten, nothing in the cache references the previous `*sql.Stmt`, so no later eviction, TTL sweep or `Purge` can reach it. It stays open until the whole `*sql.DB` is closed, and on PostgreSQL/MySQL the server-side prepared statement stays allocated with it.

The entry count does not change on overwrite, so this is not bounded by `PrepareStmtMaxSize`. It leaks roughly one statement per TTL cycle per hot query.

### Fix

Remove the entry in `Get` when it is found expired, rather than leaving it for the cleanup goroutine. The subsequent `Add` is then a fresh insert instead of a silent overwrite, and the old statement is closed at the moment it is detected as expired.

This reuses the existing eviction path — `removeElement` is what the TTL sweeper (`deleteExpired`) already calls, and `Get` already holds the write lock, which is what `removeElement` documents that it requires. So it does not introduce a new close call site or a new close-while-in-use window; it moves the close that the sweeper would have performed anyway to the point where expiry is first observed.

`Peek` has the same leave-it-in-the-map behaviour but only takes a read lock, so it is deliberately left unchanged.

### Scope

Two related things I left out on purpose, happy to fold either in if you'd prefer:

- I did not add an `onEvict` call to `Add`'s already-present branch. With this fix that branch is no longer reachable via expiry, and calling `onEvict` there would close a value that a concurrent reader may still hold — which is the subject of #7831.
- `PreparedStmtDB.Close()` iterates `Stmts.Keys()`, and `LRU.Keys()` filters out expired entries, so an expired-but-still-present statement is also skipped by an explicit `Close`/`Reset`. Same root cause, but a separate code path.

### Tests

`internal/lru` had no test file; this adds one. `TestGetExpiredEvictsValue` backdates an entry's `ExpiresAt` under the lock and asserts the eviction callback fires — deterministic, no sleeps. It fails on master:

```
--- FAIL: TestGetExpiredEvictsValue (0.00s)
    lru_test.go:50: onEvict got [] after reading an expired entry, want [old]
```

`TestGetLiveEntryDoesNotEvict` guards the other direction.

Verified locally: `go test -race ./...` in the root module and `GORM_DIALECT=sqlite go test -race -count=1 ./...` in `tests/` both pass, and `GOOS=linux GOARCH=386 go build ./...` is clean.


alliasgher:
The postgres jobs here are red for a reason unrelated to this change: `TestMigrateIndexes` fails with `syntax error at or near "CURRENT_SCHEMA" (SQLSTATE 42601)` from `gorm.io/driver/postgres/migrator.go:192`.

Every PR run since 2026-08-01 fails the same way regardless of content (last green run on any branch was 2026-07-30) — `tests_all.sh` does `go get -u -t ./...`, so CI picks up the newest driver rather than the pinned one. Nothing in this PR touches the migrator.

The sqlite job, which is where `internal/lru` actually runs, is green.

PR #7845, opened 2026-08-27
fix(logger): apply the connection logger's ParamsFilter in DB.Scan

### LINKED ISSUE #7844: ParameterizedQueries is not honoured by DB.Scan
## GORM Playground Link

https://github.com/go-gorm/playground/pull/858

## Description

With a logger configured `ParameterizedQueries: true`, statements are logged with placeholders as expected — except from `DB.Scan`, which logs the interpolated parameter values.

Same logger, same `WHERE name = ?` in all four cases:

| query                                                            | logged SQL                                                       |
| ---------------------------------------------------------------- | ---------------------------------------------------------------- |
| `Where("name = ?", secret).Find(&users)`                         | ``SELECT * FROM `users` WHERE name = ?``                         |
| `Model(&User{}).Where("name = ?", secret).Pluck("name", &names)` | ``SELECT `name` FROM `users` WHERE name = ?``                    |
| `Raw("SELECT name FROM users WHERE name = ?", secret).Row()`     | `SELECT name FROM users WHERE name = ?`                          |
| `Model(&User{}).Where("name = ?", secret).Scan(&out)`            | ``SELECT * FROM `users` WHERE name = "SHOULD_NOT_BE_LOGGED"`` ⚠️ |

It isn't specific to raw SQL — the `Raw` query is fine, while the plain model query with `Scan` is not. Tested against v1.31.2; the playground PR runs against master.

### Cause

`DB.Scan` replaces the connection's logger with a recorder for the inner execution:

```go
// finisher_api.go
currentLogger, newLogger := config.Logger, logger.Recorder.New()
config.Logger = newLogger
```

`logger.Recorder` is seeded at package level from `logger.Default`, and `New()` copies that embedded logger:

```go
// logger/logger.go
Recorder = traceRecorder{Interface: Default, BeginAt: time.Now()}

func (l *traceRecorder) New() *traceRecorder {
	return &traceRecorder{Interface: l.Interface, BeginAt: time.Now()}
}
```

So the connection's logger, and its `Config`, aren't consulted while the query runs. `callbacks.go` asks the current logger for `ParamsFilter`, which is now the recorder, and `traceRecorder.ParamsFilter` can only delegate to the package-level `RecorderParamsFilter` — which defaults to passing params through. By the time the real logger is called, it receives an already-rendered SQL string.

### Workaround, and why I'm raising it

Setting `RecorderParamsFilter` (added in #7336) does cover this path, and it's what we use. Being a package-level variable, it applies process-wide: two connections opened with different `ParameterizedQueries` settings still behave the same on `Scan`, and it isn't safe to change once connections are in use.

### Possible direction

Seeding the recorder from `config.Logger` and having `traceRecorder.ParamsFilter` prefer the wrapped logger's filter, with `RecorderParamsFilter` kept as the fallback, would honour the connection's setting without affecting anyone relying on the global. I tried this locally against v1.31.2 — the playground test passes, and a logger that doesn't implement `ParamsFilter` still goes through the global.

One detail if you'd like it done that way: the check needs a small interface declared inside the `logger` package rather than `gorm.ParamsFilter`, since referencing that would be an import cycle. Dropping `traceRecorder`'s own `ParamsFilter` and relying on embedding won't work either, because the embedded field's type is `logger.Interface`, which doesn't declare the method.

OrkhanAlikhanov:
- [x] Do only one thing
- [x] Tested

Fixes #7844 — `logger.Config.ParameterizedQueries` had no effect on `DB.Scan`, which logged the statement with parameter values interpolated.

`DB.Scan` swaps the connection's logger for a recorder so it can emit one log line after scanning, with the final row count and any scan error. That recorder came from `logger.Recorder`, a package-level value seeded from `logger.Default`, so the connection's logger — and therefore its `Config` — was never consulted while the statement ran. `callbacks.go` asked the recorder for `ParamsFilter`, the recorder could only delegate to the package-level `RecorderParamsFilter`, and that defaults to passing params through.

Two changes:

- Build the recorder from `config.Logger`, which is already on that line, via a new `logger.NewRecorder`.
- Have `traceRecorder.ParamsFilter` prefer the recorded logger's filter, keeping `RecorderParamsFilter` as the fallback.

The check is a runtime assertion against an interface declared inside `logger`: `gorm.ParamsFilter` can't be referenced there without an import cycle, and embedding alone doesn't work because `traceRecorder`'s own `ParamsFilter` shadows the promoted method and the embedded field's static type doesn't declare it.

**Not breaking:** nothing is removed or changed in signature, and a logger that doesn't implement `ParamsFilter` still goes through `RecorderParamsFilter` exactly as before. The one behavioural difference is the intended one — a logger implementing `ParamsFilter`, or configured with `ParameterizedQueries: true`, now has that honoured on `Scan` as well.

**Tested:** unit tests added in `logger/logger_test.go` covering both branches — a recorder built for a logger with `ParameterizedQueries` in each state, and a logger without `ParamsFilter` still reaching the global. End-to-end reproduction in go-gorm/playground#858, which fails on `master` and passes with this change.

### User Case Description

We route gorm through a zerolog adapter that implements `gorm.ParamsFilter` so that API tokens, deploy keys and OAuth secrets never appear in SQL logs. That works for `Exec`, `Find`, `First` and `Pluck`, but not `Scan` — a statement like `db.Table("api_tokens").Where("token = ?", secret).Scan(&out)` wrote the token straight into the log, with no indication the filter had been bypassed. Setting `RecorderParamsFilter` does cover it, but as a package-level variable it can't vary per connection and isn't safe to change once connections are in use.


goransh-walia:
Clean and minimal — appreciate that it stays scoped to the actual fix.

PR #7853, opened 2026-09-07
fix(migrator): return early when ParseWithSpecialTableName fails in ReorderModels

### LINKED ISSUE #6578: Nil pointer dereference in migrator.ReorderModels
## GORM Playground Link

https://github.com/go-gorm/playground/pull/647

## Description

I've been trying to work around SQLite's limitation of not being able to modify columns by creating a new table with `AutoMigrate` and copying the contents of the original table over before deleting the original table and renaming the new one to the name of the origin one.

Like so:
```go
db.Table("foo_temp").AutoMigrate(&Foo{})
```

Unfortunately this triggers a panic for my type `Foo`. The playground link contains a minimal example that reproduces the issue.
Expected behaviour would be a more meaningful error in case what I'm trying to do is illegal.

Stacktrace:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x104f4a2f4]

goroutine 5 [running]:
testing.tRunner.func1.2({0x105670380, 0x105b9f9b0})
	/opt/homebrew/Cellar/go/1.21.0/libexec/src/testing/testing.go:1545 +0x274
testing.tRunner.func1()
	/opt/homebrew/Cellar/go/1.21.0/libexec/src/testing/testing.go:1548 +0x448
panic({0x105670380?, 0x105b9f9b0?})
	/opt/homebrew/Cellar/go/1.21.0/libexec/src/runtime/panic.go:920 +0x26c
gorm.io/gorm/migrator.Migrator.ReorderModels({{0x8?, 0xc0004f9e60?, {0x105741c68?, 0xc0004f9e90?}}}, {0xc0005e6700, 0x1, 0xc000194c18?}, 0x1)
	/Users/cschinnerl/go/src/github.com/ChrisSchinnerl/playground/gorm/migrator/migrator.go:939 +0x644
gorm.io/gorm/migrator.Migrator.AutoMigrate({{0x80?, 0xc0004f9e60?, {0x105741c68?, 0xc0004f9e90?}}}, {0xc0005e6700, 0x1, 0x1})
	/Users/cschinnerl/go/src/github.com/ChrisSchinnerl/playground/gorm/migrator/migrator.go:113 +0x74
gorm.io/gorm.(*DB).AutoMigrate(0xc000194e78?, {0xc0005e6700, 0x1, 0x1})
	/Users/cschinnerl/go/src/github.com/ChrisSchinnerl/playground/gorm/migrator.go:24 +0x54
gorm.io/playground.TestGORM(0x0?)
	/Users/cschinnerl/go/src/github.com/ChrisSchinnerl/playground/main_test.go:22 +0xdc
testing.tRunner(0xc00048eb60, 0x105733e38)
	/opt/homebrew/Cellar/go/1.21.0/libexec/src/testing/testing.go:1595 +0x198
created by testing.(*T).Run in goroutine 1
	/opt/homebrew/Cellar/go/1.21.0/libexec/src/testing/testing.go:1648 +0x5dc
FAIL	gorm.io/playground	0.235s
```

zmsmao:
Fixes #6578

## What

In `migrator.ReorderModels`, when `ParseWithSpecialTableName` fails, the error was only logged and execution continued:

```go
if err := dep.ParseWithSpecialTableName(value, m.DB.Statement.Table); err != nil {
    m.DB.Logger.Error(context.Background(), "failed to parse value %#v, got error %v", value, err)
}
// ...dep.Statement.Schema is nil here on failure
if _, ok := parsedSchemas[dep.Statement.Schema]; ok {
    return
}
for _, r := range dep.Schema.Relationships.Relations { // nil pointer dereference
```

At that point `dep.Statement.Schema` is `nil`, so the loop over `dep.Schema.Relationships.Relations` panics.

This change adds a `return` so the value is skipped. The existing error log is kept.

## Why

A panic during migration is much harder to diagnose than a logged error. Returning early turns a hard crash into a visible, actionable message, which is what the issue asks for ("a more meaningful error instead of a panic").

## How it was verified

Added `TestReorderModelsWithUnsupportedType` in `migrator/reorder_models_test.go`:

- **Before the fix**: the test panics — the log shows `failed to parse value 123, got error unsupported data type`, immediately followed by a nil pointer dereference.
- **After the fix**: `ReorderModels` returns without panicking and the error is still logged.

No real database is required: `ParseWithSpecialTableName` returns `ErrUnsupportedDataType` for non-struct values, so constructing `migrator.Migrator` on top of `tests.DummyDialector` is enough to reproduce.

`go test ./migrator/` passes.


zmsmao:
A note on the CI failures (`lint` and `tests / postgres (postgres:14, oldstable, ubuntu-latest)`), in case it saves you time:

These look pre-existing on `master`. From `gh run list` for this repo:

```
lint (golangci-lint workflow) — last 5 runs:
  failure  fix/migrator-nil-schema-guard   34098085501
  failure  add-ping-context-timeout        33758583921
  failure  fix/wrap-multiline-boolean-...  33467146637
  success  codex/go127-db-generic          33412444435
  failure  buildcond_values                33388558453

tests workflow — last 5 runs: all failure
```

Of the lint failures on this PR, none point at the line this PR changes (migrator/migrator.go:942). The reported sites are:

- `logger/sql.go:11` — `could not import unicode (-: export data version 4 is greater than maximum supported version 2)`. That's a Go toolchain / golangci-lint version mismatch, not source.
- `schema/field_test.go:115-116` — `user.ID` / `user.CreatedAt undefined`. Unrelated package.
- `migrator/migrator.go:940, 950, 977, 980` — `dep.Schema` / `dep.ParseWithSpecialTableName undefined (type Dependency has no field or method …)`. These lines exist on master unchanged.

So the only diff this PR introduces is the `return` at line 942, which cannot affect typecheck resolution of `dep` or the standard library import.

For `postgres:14 oldstable`: the other postgres variants (13, 15) are reported as "cancelled after 2-3m", which on this workflow usually means they were queued behind the oldstable matrix slot. So I don't have a clean signal from a non-14 postgres to compare.

Happy to re-run the failed jobs or dig into a specific failure if you'd like one isolated.

PR #7863, opened 2026-09-14
ci: pin gorm.io/driver/postgres to v1.6.0 in tests module to fix broken postgres CI

tr1v3r:
## Problem

All postgres CI jobs fail on every recent PR (e.g. #7853–#7861) with:

```
--- FAIL: TestMigrateIndexes
    migrate_test.go:485: Failed to drop index for user's name, got err
    ERROR: syntax error at or near "CURRENT_SCHEMA" (SQLSTATE 42601)
    DROP INDEX CURRENT_SCHEMA()."idx_index_structs_name"
```

## Root Cause

This is a known regression in `gorm.io/driver/postgres` v1.6.1+ (introduced by go-gorm/postgres#340, tracked in go-gorm/postgres#350): `DropIndex`/`RenameIndex` qualify the index with `CURRENT_SCHEMA()` when no explicit schema is set, which is invalid PostgreSQL syntax. Last working driver version is v1.6.0.

`tests/go.mod` already requires `gorm.io/driver/postgres v1.6.0`, but `tests/tests_all.sh` runs `go get -u -t ./...` on every CI run, which upgrades the driver to the latest (currently broken) v1.6.2 — so the checked-in pin is silently defeated.

## Fix

Add a `replace` directive pinning `gorm.io/driver/postgres` to v1.6.0 in `tests/go.mod`. Unlike a `require` pin, a `replace` is skipped by `go get -u`, so the pin survives the test script's upgrade step.

Verified locally:

- `go list -m gorm.io/driver/postgres` → `v1.6.0 => v1.6.0`
- `go get gorm.io/driver/postgres@latest` leaves the module at v1.6.0 (replaced modules are not upgraded)
- `go mod tidy` produces no go.sum churn (v1.6.0 hashes already checked in)

Remove the directive once a fixed driver is released (go-gorm/postgres#350).


tr1v3r:
Closing in favor of fixing the root cause in the driver repo (go-gorm/postgres#350) — thanks for the pointer.

PR #7866, opened 2026-09-17
fix: clear stale time pointers when scanning NULL

### LINKED ISSUE #7855: First retains stale *time.Time when reloading a selected SQL NULL into an existing struct
## GORM Playground Link

https://github.com/go-gorm/playground/pull/861

## Description

Reloading a record leaves an old `*time.Time` value hanging around after the column becomes `NULL`.

1. Load a row with a non-null timestamp using `First(&record, 1)`.
2. Set that column to `NULL` in the database.
3. Call `First(&record, 1)` again.

I'd expect the pointer to become `nil`. Instead, it still holds the old timestamp. Loading the same row into a fresh struct gives `nil`, as expected. No query errors.

The repro is just an ID and a `*time.Time` field. No hooks or custom types. Everything runs in one transaction.

Reproduced with GORM v1.31.1 and v1.31.2, PostgreSQL driver v1.6.0. The linked test fails on PostgreSQL; I haven't checked other databases.

ryanchou1994:
- [x] Do only one thing
- [x] Tested

When a reused query destination has an existing non-nil `*time.Time`, scanning SQL `NULL` now clears that field instead of leaving the stale value. The setter checks the existing field before writing the zero value, so an all-NULL embedded pointer struct remains nil.

This adds regression coverage for reloading the same destination and for preserving fields that were not selected. The change passed the full SQLite race suite and focused race tests on PostgreSQL 15 and MySQL 8.

### User Case Description

A record can be loaded into a struct, updated so its nullable timestamp becomes `NULL`, and then loaded into the same struct again. Previously, the timestamp still appeared non-null because the NULL scan skipped the assignment. The reloaded struct now matches the database row.

Fixes #7855


schema/field.go
```diff
@@ -862,8 +862,12 @@ func (field *Field) setupValuerAndSetter(modelType reflect.Type) {
 			field.Set = func(ctx context.Context, value reflect.Value, v interface{}) error {
 				switch data := v.(type) {
 				case **time.Time:
-					if data != nil && *data != nil {
-						field.ReflectValueOf(ctx, value).Set(reflect.ValueOf(*data))
+					if data != nil {
+						if *data != nil {
```
efedosee, line 866: Are this and next check for zero value needed? Another approach is to call `Set` function with `*data` as it's done at line 840.

ryanchou1994: Thanks, using `Set` for the assignment makes sense. The NULL case still needs the zero-value guard: the `*time.Time` setter calls `ReflectValueOf`, which allocates a nil embedded parent even when the value being assigned is nil. `TestEmbeddedPointerTypeStruct` expects that parent (`Author`) to remain nil when its columns are all NULL.

In a focused Linux check, delegating directly to `Set(ctx, value, *data)` under the existing `data != nil` guard makes that test fail with a non-nil `Author`; the current implementation passes.

Would keeping the zero-value guard and delegating the assignment to `Set` address your suggestion?

efedosee: I see why zero value check is needed, thanks for explanation.
About Set call - it's optional and more about code style. I would better unwrap nested if statement. But it's completely up to you.

ryanchou1994: Thanks for clarifying. I'll keep the current implementation as it is, including the zero-value guard. Appreciate the review!

